PR #27, opened 2014-11-29
8.0-MP94809

cowardvassal:


nhomar:
This fix is out of date, re open if we need again this.

PR #33, opened 2014-12-02
[MERGE] the purchase order lines created via a purchase requisition will now have the same product quantity, product uom that the one in the purchase requisition line. Also the unit price is set to 0 because the analyst will introduce it manually.

zaoral:
[MERGE] the purchase order lines created via a purchase requisition will now have the same product quantity, product uom that the one in the purchase requisition line. Also the unit price is set to 0 because the analyst will introduce it manually.


zaoral:
yellow runbot because cicsa.
This functionality is a priority
please merge.

PR #43, opened 2014-12-09
[FIX][invoice_datetime] fixed error,  duplicate  and sign invoice with pac

ernesto-gm:
[FIX][invoice_datetime] fixed error,  duplicate  and sign invoice with pac

### DIFF
--- a/invoice_datetime/invoice.py
+++ b/invoice_datetime/invoice.py
@@ -144,7 +144,10 @@ class account_invoice(osv.Model):
             context = {}
         if default is None:
             default = {}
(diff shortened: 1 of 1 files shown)


nhomar:
@ernesto-gm can you tell me, where is the issues which eplain this change please, it is not related to this!

Regards.

jorgenaranjoVauxoo:
:+1: 

This PR was create because, when you duplicate the invoice,  not setting false in date_invoice_tz field. Then when you try sign the electronic invoice, show a message that already exist a invoice with sign previous.

PR #44, opened 2014-12-12
Module 'purchase_requisition_line_analytic' should not really depend on 'account_analytic_plans'

JordiBForgeFlow:
Should only depend on 'account_analytic'. The dependence on 'account_analytic_plans' installs this module, which can be a problem for companies that really don't use it.

If purchase requisitions should really make use of analytic plans, then I suggest a new module on top of the existing called 'purchase_requisition_line_analytic_plan'.

Regards,
Jordi.

### DIFF
--- a/purchase_requisition_for_everybody/security/requisition_security.xml
+++ b/purchase_requisition_for_everybody/security/requisition_security.xml
@@ -12,14 +12,25 @@
         <field name="implied_ids"
         eval="[(6,0, [ref('purchase_requisition.group_purchase_requisition_user')])]"/>
     </record> 
+    <record id="requisition_modify_ev_followers_rule" model="ir.rule">
+        <field name="name">Follow Purchase Requisition</field>
+        <field name="model_id" ref="purchase_requisition.model_purchase_requisition"/>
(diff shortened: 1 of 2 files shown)


nhomar:
@hbto Can you send me a +1 please. I think @jbeficent has a point here.

@jbeficent can you complement this mp with the second module that you are commenting in order to make only one merge with the split, please.

JordiBForgeFlow:
@nhomar Module 'purchase_requisition_line_analytic_plan' that I was not mentioning really does not exist. I was just pointing out that it would need to be created, if needed, instead of adding the dependency of 'account_analytic_plans' in  'purchase_requisition_line_analytic'.

JordiBForgeFlow:
Hi, I just added a new proposed change to this PR. @nhomar @hbto can you review & merge if you agree?

JordiBForgeFlow:
@moylop260 Can you review this one also?

moylop260:
@jbeficent 
You have two changes.
First one: purchase_requisition_for_everybody
Second one: account_analytic_plans

Two changes LGTM.
But is better if you next fix is in 2 PR.

moylop260:
@jbeficent 
Merged
Thanks for you contribution.

PR #55, opened 2014-12-18
[FIX] Commenting account.invoice inherited views as they collide with purchase_analytic_plans

hbto:
Commenting account.invoice inherited views as they collide with purchase_analytic_plans


hbto:
Runbot become yellow.

PR #56, opened 2014-12-18
[FIX] cureate by create

yaniaular:
Error in method name is fixed...


yaniaular:
@nhomar Fix found. To merge ASAP

PR #79, opened 2015-01-23
8.0 jose producto extended variants

josemoralesp:
Overwrite the method that recompute the cost from mrp bom to allow it update from the variants


josemoralesp:
:+1:  @dsabrinarg

PR #89, opened 2015-01-28
[ADD][product_unique_default_code] Unitest added to prove new unique constraint for product's internal reference

sergioernestotostadosanchez:
A test is added to prove the new constraint added with this module:
- Test 1: To prove if the module can copy a product with default code included
- Test 2: To prove if the constraint for unique internal references is correctly active

### DIFF
--- a/product_unique_default_code/tests/test_for_unique_ref.py
+++ b/product_unique_default_code/tests/test_for_unique_ref.py
@@ -24,9 +24,10 @@
 #
 ###############################################################################
 
+from psycopg2 import IntegrityError
+
 from openerp.tests.common import TransactionCase
 from openerp.exceptions import except_orm
-from openerp.tools import mute_logger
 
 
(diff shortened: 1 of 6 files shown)


product_unique_default_code/tests/test_for_unique_ref.py
```diff
@@ -0,0 +1,90 @@
...
+        product_data = {
+            'name': 'Test Cellphone ACME',
+            'default_code': '101001000',
+            'categ_id': self.env.ref('product.product_category_all').id,
+            'standard_price': 100.90,
+            'list_price': 123.50
+        }
+        product = self.env['product.product'].create(product_data)
+        self.assertTrue(
+            product.copy().default_code == '%s (copy)' % product.default_code,
```
moylop260, line 60: @sergioernestotostadosanchez 
Excelente código, la nueva api trabajando al 100%
`product.copy().default_code`

PR #93, opened 2015-01-30
[FIX] Fix in controller of the report

josemoralesp:
When we make a super we need to send the original values of the funcion, to avoid mistake for minimal changes in their values


josemoralesp:
:+1: @dsabrinarg

PR #95, opened 2015-01-30
[FIX][sale_order_customized] problem is solved with empty description lines, to change the quantity.

ernesto-gm:
[FIX][sale_order_customized] problem is solved with empty description lines, to change the quantity.


josemoralesp:
:+1:  @dsabrinarg

PR #98, opened 2015-02-03
Render Report

josemoralesp:
Added new variables to work with the differents way to generate the url to create the reports


josemoralesp:
:+1:  @dsabrinarg

PR #99, opened 2015-02-04
Added validation to cant validate an statement or voucher with invoice refund

luistorresm:
Added validation to cant validate an statement or voucher with invoice refund


nhomar:
@luistorresm Is this PR valid yet?

Can you rebase your commits because it is not mergeable.

PR #107, opened 2015-02-06
[REF][account_refund_early_payment] use the accounts to reconcile to reconcile line with tax

ernesto-gm:
[REF][account_refund_early_payment] use the accounts to reconcile to reconcile line with tax


JulioSerna:
:+1:

nhomar:
Esto no está bien.. oi no lo entendí

PR #138, opened 2015-02-22
[REF][stock_move_entries] forward port to 8.0

ernesto-gm:
[REF][stock_move_entries] forward port to 8.0


stock_move_entries/model/stock_move_entries.py
```diff
@@ -114,7 +114,7 @@ def _create_account_move_line(self, cr, uid, move, src_account_id,
```
hbto, line 115: Get rid of one of these two inheritance as they are doing the same think.

Best Regards.

PR #143, opened 2015-02-25
[REF][stock_move_entries] methods duplicates are removed, the class stock_picking

ernesto-gm:
[REF][stock_move_entries] methods duplicates are removed, the class stock_picking


hbto:
+1

PR #148, opened 2015-02-27
[IMP] added validation in process to manual reconciliation in module account_voucher_tax and account_refund_reconciliation #521

JulioSerna:


hbto:
+1

PR #153, opened 2015-03-02
[IMP][account_move_filters]Changed filter Unreconciled AP/AR by the existing filter Unreconciled

luistorresm:
Reused filter Unreconciled, only change the domain and help


hbto:
+1

PR #161, opened 2015-03-09
[REF][account_move_report] migrate journal entries report to version 8.0

ernesto-gm:
[REF][account_move_report] migrate journal entries report to version 8.0


jorgenaranjoVauxoo:
:+1:

PR #162, opened 2015-03-11
[ADD][lodi_supplier_invoice_qty_attachments] added nuember attachments i...

hugho-ad:
...n tree view on account_invoice


nhomar:
@hugho-ad 

Modules in addons-vaxuoo should have generic names the lod_\* part of this module is not correct, can you remove it please usinf a git mv command?.

The same with the String name on descriptor.

One time you depends of account, you do not need to depend onf "base" can you remove such dependency?.

Descriptor should be explaining the "why" and not the "how" technically speaking, can you talk about "Invoice" and not "account_invoice" and please explain why we developed such feature.

Why did you overwrite the  "def fields_view_get" instead simply inherit the tree view for invoices?

hugho-ad:
@nhomar 
I had problems trying to hide the "qty_attachments" field in customer bills, we attempted to attributes

```
attrs = "{'invisible': [('type','!=','in_invoice')]}"
```

Something like that. But it did not work.

@JulioSerna 
 recommended me to do with the method "def fields_view_get", even I try putting direct attr invisuble in  method,  but got no result, It only worked with remove.

nhomar:
@hugho-ad 

Please create another branch and make it with inheritance of the view to see what especifically you tried.

I need to read the code I think it is failing you because another different issue.

And also please check if any other module overwrite the field_view_get on account.invoice model, I think there is none, but I can be wrong.

Regards.

hugho-ad:
@nhomar 
*With inheritance did'nt work
*overwrite the fields_view_get method did work pretty good.
any way , I'll made another branch to show you.

PR #170, opened 2015-03-13
[REF][account_move_report] information is complete journal entries in the report.

ernesto-gm:
[REF][account_move_report] information is complete journal entries in the report.


josemoralesp:
:+1:  @dsabrinarg

PR #172, opened 2015-03-16
[IMP] product_unique_default_code: change the way the default code is se...

foutoucour:
...t unique

Previously, a sql constraint was set on default code. This constraint makes
tests of purchase module to fail.
Because of that, the constraint has been removed in favor of a onchange
trigger. That makes the process a bit weaker as the check is only done
against user actions.

### DIFF
--- a/product_unique_default_code/__openerp__.py
+++ b/product_unique_default_code/__openerp__.py
@@ -26,7 +26,7 @@
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
 {
     "name": "Product Default Code Unique", 
-    "version": "1.1", 
+    "version": "2.0",
     "author": "Vauxoo", 
     "category": "", 
     "description": """
@@ -40,7 +40,7 @@
     "demo": [
         'demo/test_unique_ref_demo.xml',
     ], 
-    "data": [], 
+    "data": [],
     "test": [], 
     "js": [], 
     "css": [], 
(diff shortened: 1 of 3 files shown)


nhomar:
Hello @foutoucour .

The onchage trigger will not ensure the consistency of database when for example you use webservices.

What you need to do to test this module is install purchase allways before This module (it can be automized adding a sequence in the module descriptor [not tested yet]), if you propose you change in that way we can merg it, but we can not move from a solid sql contraint to an onchange it is technically and functionally incorrect (IMHO).

Thanks for contribute.

Other point is that in the las version 6 of odoo we solved that with some fixing in demo data.

I hope it helps.

nhomar:
BTW we use purchase in all our enviroments and it is green!

Regards.

bwrsandman:
@nhomar I will take over from here.

What is the issue with this PR, and do you think this module could be moved to OCA?

nhomar:
Hello @bwrsandman.

I explained the issue above, with some PR that we propose to Odoo on demo data (already merged I think) you do not have the need to propose a change on Python the uniqueness can be managed as it is in sqlwith no problems on CI.

Our specific need of this module is because we need create hundred of thousands of products with sql and it should not work to us have only a python constraint.

Then, if we are agreed that the plan is: 
- default code unique by sql.
- if in the future some test/demo on the core brake such uniqueness we must path such data and not the module.

Then I give you the witness from here and we can move to OCA.

If we can not be agreed on such behavior then we will need to develop 2 modules a module which ensure uniqueness by Python (your version) and this one.

Thanks for your great effort and understanding.

PR #182, opened 2015-03-24
[REF] Fix pylint

moylop260:


moylop260:
@nhomar,
:+1:
runbot with same mumber of errors vs runbot stabe branches

PR #183, opened 2015-03-24
[ADD] app that shows the products purchased by another customers 

oscarolar:
It will show the products that were boutght in the same sale order of the current product, here is the [HU](https://www.vauxoo.com/?ts=1426631879796#id=338&view_type=form&model=user.story&action=726) and here is the [RUNBOT](http://runbot.vauxoo.com/runbot/team/yoytec-12#8_0-yoytec-what-other-customers-purchased) build


website_customer_also_purchased/model/product.py
```diff
@@ -0,0 +1,75 @@
...
+        cr.execute("""
+            select id
+            from product_product
+            where product_tmpl_id = {0}
+            """.format(ids[0]))
+        pid = cr.fetchall()
+        for p in pid:
+            pids_t.append(p[0])
+        pids = '(%s)' % ', '.join(map(repr, tuple(pids_t)))
+        cr.execute("\
```
nhomar, line 52: This can be done with an UNION you do not need run two separate SQL to achieve this deal.

http://www.postgresql.org/docs/8.3/static/queries-union.html

website_customer_also_purchased/model/product.py
```diff
@@ -0,0 +1,75 @@
...
+##############################################################################
+
+from openerp.osv import osv
+from openerp.osv import fields
+
+
+class product_template(osv.osv):
+    _inherit = 'product.template'
+
+    def _get_purchased(self, cr, uid, ids, field_names, arg=None,
```
nhomar, line 33: If you use @api.one

This is V8 ready.

website_customer_also_purchased/model/product.py
```diff
@@ -0,0 +1,75 @@
...
+        res = cr.fetchall()
+        for r in res:
+            result[ids[0]].append(r[0])
+        return result
+
+    _columns = {
+        'customer_purchased': fields.function(_get_purchased, method=True,
+                                              type='one2many',
+                                              relation='product.template',
+                                              string="Customers Purchased")
```
nhomar, line 74: USe V8.0 format please.

website_customer_also_purchased/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+    "version": "1.0",
+    "author": "Vauxoo",
+    "category": "Website",
+    "description": """
+Website Customer Also Purchased
+===============================
+    This app shows the products that were bought in the same
+    purchase order by a other customers related to the current
+    product
+    """,
```
nhomar, line 13: This module will be very well accepted publically, can you set description in a separate README.md please.

website_customer_also_purchased/model/product.py
oscarolar, line 52: This was done like this to avoid pep8 long lines and to avoid making a sql query that cannot be easy to read.

website_customer_also_purchased/model/product.py
oscarolar, line 33: WIP

website_customer_also_purchased/model/product.py
oscarolar, line 74: WIP

website_customer_also_purchased/__openerp__.py
oscarolar, line 13: WIP

website_customer_also_purchased/model/product.py
oscarolar, line 52: BTW Union is for something else, this here was made to Nest queries depending on results.

nhomar, line 52: 2015-03-24 14:28 GMT-06:00 Olar Alca [Vauxoo] notifications@github.com:

> This was done like this to avoid pep8 long lines and to avoid making a sql
> query that cannot be easy to read.

It is not a good reason, I think 1 best SQL is better.

Even thinks on such terms you can create/reuse sale_order_report which
already has part of this SQL, don't you think?

## 

---

Saludos Cordiales

  [image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
     http://about.me/nhomar?promo=email_sig

PR #188, opened 2015-03-26
[ADD] new module that shows the clasic 5 star rating on products

oscarolar:
Here is the [RUNBOT](http://runbot.vauxoo.com/runbot/team/yoytec-12#8_0-yoytec-website-rate-product)

### DIFF
--- a/website_rate_product/__init__.py
+++ b/website_rate_product/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import controllers
--- a/website_rate_product/__init__.py
+++ b/website_rate_product/__init__.py
@@ -1,2 +1,3 @@
+# -*- encoding: utf-8 -*-
 from . import models
 from . import controllers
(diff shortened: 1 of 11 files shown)


website_rate_product/controllers/__init__.py
```diff
@@ -0,0 +1 @@
+from . import main
```
nhomar, line 1: Falato

# -_\- coding: utf-8 -_-

website_rate_product/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import controllers
```
nhomar, line 3: Falta # -_\- coding: utf-8 -_-

oscarolar:
Are you serious??? [LINK](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/website_variants_extra/models/__init__.py)

oscarolar:
If runbot is green here we are ready to merge.

PR #202, opened 2015-04-02
8.0 stock accrual reconcile hbto

hbto:


hbto:
Test went ok, 

Failed test are related to commission_payment, and those Failures are linked to change in dates

PR #219, opened 2015-04-14
[8.0][product_unique_default_code] Added the manufacure_code field

josemoralesp:
Added new field to set the manufacture code, This field allows to products have the same code if the products belong to the same part


nhomar:
HEllo @josemoralesp 

Why this feature here?

https://github.com/OCA/product-attribute/tree/8.0/product_manufacturer

Thi smodule is the one we decided to use.

Regards.

PR #232, opened 2015-04-22
WIP NOT MERGE [FIX] css converted to Less to proper use of styling with page namespaces

oscarolar:
WIP STILL WIP


website_authorized_service_centers/static/src/css/service_centers.less
```diff
@@ -1,6 +1,4 @@
-*{
-    font-family: 'Open Sans', sans-serif;
-}
+.asc {
```
nhomar: set a more specific name (longer) to avoid mix in other module please.

PR #240, opened 2015-04-24
[REF][account_group_auditory] change pass and adding attr noupdate=1 in data

hugho-ad:
This pr change the password, modify the data
and delete menu_acces account.menu_finance_receivables from the Auditor group
This change is motivated by a issue that happens when you install just OML or odoo-ifrs, the menu Accounting/Customer become hidden
the solutions is move this "menu_acces" from addons-vauxoo/account_group_auditory  to lodigroup/lodi-security


JulioSerna:
:+1:

PR #255, opened 2015-05-07
[FIX][account_invoice_number]Repair view that show internal number to customer invoice

luistorresm:
Fix view, because not showed the internal number


JulioSerna:
:+1:

PR #267, opened 2015-05-15
[8.0][product_historical_price] Modified the module to remove useless fields

josemoralesp:
Removed useless fields in product template because these fields shown duplicate info 
Remove the model product historic cost to use a new model of the odoo to save the historical cost of the products
Created action server to avoid overwrite the main methods (create and write)
Created unit test to confirm the functionality of the module


hbto:
+1

PR #272, opened 2015-05-19
[UPD] Add french translation

ehdem:


hbto:
+1
Hello ehdem, It is good to know that Frech-speaking people is interested in this module,

I have struggle to keep this module alive because in some way it clashes with another
module of ours, aging_due_report.

In v8 we have invested some time to improve aging_due_report using/bringing some 
of the feature that were added to account_aged_partner_balance.

I would like to have your critics in that module too, aging_due_report.
Do not mind to be rude in your critics. Plain and clear critics are welcome.

I have envisioned that some of the features that we can see in account_aged_partner_balance_vw,
could become mainstream feature in a near future in aging_due_report.

Best Regards.

PR #292, opened 2015-06-02
[8.0][stock_trasnfer_location_dest] Added new module to ease the change of the destination location at the transfer moment

josemoralesp:
Added new module that adds two fields in the wizard to do any transfer from the picking. These fields allow you change the destination location in all lines related with the transfer. With this you decide where you want put the products at the moment to do the transfer


luisg123v:

PR #297, opened 2015-06-09
[MERGE] add a new v7 functional field to show the puchase incoming qty of a product like is done with the incoming aty and on hand qty fields in odoo.

zaoral:
Task [vx#3589](https://www.vauxoo.com/#id=3589&view_type=form&model=project.task) HU [320](https://www.vauxoo.com/?debug=#id=320&view_type=form&model=user.story&action=726) Sprint 13


zaoral:
This PR will not be used. The field created ad a multi functional field is not working. The method _product_available() is call and return the correct value but the field is not updated in the instance. 

This approach do not work for the functionality that was require.
For that reason this PR is closed and will be development in another branch.

zaoral:
This PR will not be used. The field created ad a multi functional field is not working. The method _product_available() is call and return the correct value but the field is not updated in the instance.

This approach do not work for the functionality that was require.
For that reason this PR is closed and will be development in another branch.

PR #312, opened 2015-06-15
[REF] .travis.yml: Only run test of pylint and flake8 in modules changed in each pr

moylop260:
Currenty this project is used from other ones as dependencies.
Then addons-vauxoo don't have a environment to test it.
Then we need a way to test basic cases (lint test) to accurance don't have errores in other environments but we have a lot (...really a lot) of modules to fix all in a PR.

This PR create a pylint and flake8 test only in modules changes in your current PR to fix it
incrementally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,28 @@ language: python
 python:
   - "2.7"
 
-env:
-  - LINT_CHECK="1"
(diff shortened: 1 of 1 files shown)


moylop260:
@nhomar & @oscarolar,
Can you give me plus one or your feedback, please?

nhomar:
+1

PR #317, opened 2015-06-16
[REF] .travis.yml: Remove import checks. Because we don't have all environments in addons-vauxoo

moylop260:
If you have:

``` python
from openerp.addons import OTHER_MODULE_FROM_OTHER_PROJECT
```

Before: You see a error.
Now: You don't see a error.


oscarolar:
+1 working well

PR #321, opened 2015-06-17
[FIX] project_issue_management: Fix syntaxis error

moylop260:


nhomar:
@moylop260 Man but we should not delete the help, the fix should lefting the help, No?

nhomar:
@moylop260 forget my comment.

PR #329, opened 2015-06-24
[IMP][account_voucher_tax] when invoice has taxes 0% the rouding amou…

JulioSerna:
…nt should be 0.0 and was getting null value


luistorresm:
:+1:

jorgenaranjoVauxoo:
:+1:

PR #333, opened 2015-07-04
[IMP][send_author_mail] applied pylint continuation PR #331Fixes Vauxoo/instance-vauxoo-com#90

JulioSerna:
Fixes Vauxoo/instance-vauxoo-com#90


nhomar:
@JulioSerna 

Tu avoid it happen again, can you add a unit test for this... (if you think it is not more than 3 hours of work).

JulioSerna:
@nhomar 

Ok, WIP

PR #346, opened 2015-07-09
[FIX] addons-vauxoo on some customer validations

oscarolar:
- website_product_comment_purchased: [REM] No longer necessary to show if customer didnt bought the item on product.
- website_product_availability: [FIX] similar products link only will be shown if there are alternative products
- product_category_attributes: [ADD] validation on server action if attribute is set it wont add it again

### DIFF
--- a/product_category_attributes/data/server_actions.xml
+++ b/product_category_attributes/data/server_actions.xml
@@ -11,10 +11,14 @@
 categ_obj = pool.get('product.category')
 categ_ids = categ_obj.search(cr, uid, [('id','=',object.categ_id.id)])
 datas = []
+line_attrs = []
+for line in object.attribute_line_ids:
+    line_attrs.append(line.attribute_id.id)
 if categ_ids:
     for categ in categ_obj.browse(cr, uid, categ_ids, context):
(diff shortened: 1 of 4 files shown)


product_category_attributes/data/server_actions.xml
```diff
@@ -11,10 +11,14 @@
...
+line_attrs = []
+for line in object.attribute_line_ids:
+    line_attrs.append(line.attribute_id.id)
 if categ_ids:
     for categ in categ_obj.browse(cr, uid, categ_ids, context):
         for attribute in categ.attribute_ids:
-            datas.append((0, 0, {'attribute_id':attribute.id}))
+            if attribute.id not in line_attrs:
+                datas.append((0, 0, {'attribute_id':attribute.id}))
         object.write({'attribute_line_ids': datas})]]>
```
josemoralesp, line 22: @oscarolar This is the line that we need to change to avoid use a write using the key attribute_line_ids, because when you use that the method creates a new variant[1]

[1] https://github.com/Vauxoo/odoo/blob/8.0/addons/product/product.py#L761-L762

PR #351, opened 2015-07-13
[ADD] Account Income of Discount Early Payment product is added to re…

yaniaular:
…fund line created in Refund Invoice wizard


yaniaular:
@hbto the feature you request me about early payment account and the account of the product

nhomar:
@hbto your +1 is needed here.

PR #368, opened 2015-07-25
[VERSION] 

nhomar:
[IMP] increasing all version to *.5 in order to ensure this is the 1st version with full descriptions.
[ADD] .version.py script to manage the versioning for all system


nhomar:
I will improve the logic

PR #387, opened 2015-07-30
[FIX] account_bank_statement_vauxoo: Import problem

nhomar:


nhomar:
Ignoring lint check on this module only fixing the import stuff that brakes other instances tests.

PR #390, opened 2015-08-01
[ADD] website_order_confirm: new controller that confirms order directly instead of creating a quotation HU#548

oscarolar:


oscarolar:
@nhomar just wait for the green signal on CI platforms and merge please.

PR #395, opened 2015-08-04
[FIX] Fixing Dependency

hbto:


hbto:
@Nhomar need your review

nhomar:
Here:

https://github.com/vauxoo-dev/addons-vauxoo/blob/8.0-external-dependencies-hbto/commission_payment/model/commission.py#L18

You are using mx.Datetime, that library is not python standard.

You can check it doing a blank virtualenv and trying to import such statement.

Here you can find the travis failing for such library ( a blank travis with only addons-vauxoo declared).

The correct solution is not depend at all of mx.DateTime and use standard datetime.

hbto:
Sorry @nhomar 

Your confusion is comprehensible

I _used_ to use it mx.Datetime 

but no longer it is been used,

in v6 it was true
https://github.com/Vauxoo/addons-vauxoo/blob/6.0/commission_payment/commission.py#L5

in v7 it was true
https://github.com/Vauxoo/addons-vauxoo/blob/7.0/commission_payment/commission.py#L2

but not in v8
https://github.com/Vauxoo/addons-vauxoo/blob/8.0/commission_payment/model/commission.py#L3

Check for datetime definition in python.org
https://docs.python.org/2/library/datetime.html

Best Regards.

nhomar:
@hbto 

Thanks the new update is better, you were using it (even with the intention to not to), can you only before merge verify the main runbots in order to ensure all is ok.

Thanks for the fizx is cool.

PR #399, opened 2015-08-04
[IMP]Show product availability by location

hugho-ad:
[TASK#3928](https://www.vauxoo.com/web?#id=3928&view_type=form&model=project.task&menu_id=136&action=138)
[HU#543](https://www.vauxoo.com/web?#id=543&view_type=form&model=user.story&menu_id=557&action=726)


website_product_availability/models/product.py
```diff
@@ -79,3 +79,84 @@ def _get_availability(self, cr, uid, ids, name, args, context=None):
     _default = {
         'low_stock': 0,
     }
+
+
+class LocationQuants(osv.Model):
+    _name = 'location.quants'
+
+    _columns = {
+        'product_id': fields.many2one('product.template'),
```
oscarolar: You need to add a string attribute to this fields

``` python
string="Product"
```

because headers on tree appear as 'Unknown'
image

website_product_availability/views/product_view.xml
```diff
@@ -22,5 +22,36 @@
                 </xpath>
             </field>
         </record>
+
```
oscarolar, line 25: Dude, the two new trees you created are one in the upper side and one in the down side, can you put them at the same level, I mean like in the same line:
image

hugho-ad:
@oscarolar Done, can you review modifications in the same instance

oscarolar:
:+1: 
@nhomar can you merge if you are ok with this please?, this is a partial feature that must be used in the frontend.

website_product_availability/models/product.py
```diff
@@ -82,3 +82,84 @@ def _get_availability(self, cr, uid, ids, name, args, context=None):
...
+
+
+class LocationQuants(osv.Model):
+    _name = 'location.quants'
+
+    _columns = {
+        'product_id': fields.many2one('product.template', 'Product'),
+        'location_id': fields.many2one('stock.location', 'Stock Location'),
+        'qty': fields.float('Quantity')
+    }
```
nhomar: usa api. v8 pliz esto es api v7.

website_product_availability/models/product.py
```diff
@@ -82,3 +82,84 @@ def _get_availability(self, cr, uid, ids, name, args, context=None):
...
+
+
+class ProductTemplate(osv.Model):
+    _inherit = 'product.template'
```
nhomar, line 91: models.Model

website_product_availability/models/product.py
```diff
@@ -82,3 +82,84 @@ def _get_availability(self, cr, uid, ids, name, args, context=None):
...
+
+    def _get_locations_quants(
+            self, cr, uid, ids, field_names, arg=None, context=None):
```
nhomar: nueva api

website_product_availability/models/product.py
```diff
@@ -50,35 +50,90 @@ class product_product(osv.Model):
...
+        stock_locations_ids = []
+        for quant in quants:
+            if quant.location_id.id in stock_locations_ids:
+                continue
+            stock_locations_ids.append(quant.location_id.id)
+        location_quant_ids = stock_locations_obj.search(
+            [('product_id', 'in', products),
+             ('location_id', 'in', stock_locations_ids)])
+        if location_quant_ids:
+            stock_locations_obj.unlink(location_quant_ids)
```
nhomar: why are you "unlilinking" things in a _get_\* object, did it not "delete" such ids? or only remove from cache? (it is a doubt)

nhomar:
thanks @hugho-ad can you just add some test cases here, i read som extrange unlink stuff in the method which can introduce some errors.

website_product_availability/models/product.py
hugho-ad: I've been 'unlinking' in order to update the records, like cleaning the 'cache' and get the new locations with the 'qty' info

nhomar: But this line will delete in the database suchj records that's what you want?

nhomar: and your _ids is searching for 100% of quants of this product + location_id, it means in this transaction you will delete passed data outside the namespace of this method also.....

Frequently the get_methods should not modify data only retreive it, and you must try to do in a separate function the updated information (certain conditions apply but it is the generic rule).

hugho-ad: You are right, also in that way, method brings a enormous computational cost, I already made the correction, please review.

PR #401, opened 2015-08-05
[ADD] - checkout validations with error messages and more fields added HU#923

oscarolar:


website_form_validations/README.rst
```diff
@@ -0,0 +1,7 @@
+Website Form Validations
```
nhomar: Use this template please.

https://github.com/OCA/maintainer-tools/tree/master/template/module

website_form_validations/controllers/main.py
```diff
@@ -0,0 +1,218 @@
...
+            'checkout': checkout,
+            'shipping_id': partner.id != shipping_id and shipping_id or 0,
+            'shippings': shippings,
+            'error': {},
+            'has_check_vat': hasattr(registry['res.partner'], 'check_vat')
+        }
+
+        return values
+
+    def checkout_form_save(self, checkout):
```
nhomar: One method of 100 lines with too much IF, can you split on submethods please?

nhomar: Too much business logic in the controller side.

In the controlled side should be only mandatory buisness logic for the feature try to put this logic on object side,.

website_form_validations/controllers/main.py
```diff
@@ -0,0 +1,218 @@
+# -*- coding: utf-8 -*-
+from openerp import SUPERUSER_ID
+from openerp.http import request
+from openerp.addons.website_sale.controllers.main import website_sale
+
+
+class website_sale_inh(website_sale):
+
+    def checkout_values(self, data=None):
```
nhomar: Same than other.... one method of 100 line is unreadable.

website_form_validations/controllers/main.py
```diff
@@ -0,0 +1,218 @@
...
+
+        country_ids = orm_country.search(cr, SUPERUSER_ID, [], context=context)
+        countries = orm_country.browse(cr, SUPERUSER_ID, country_ids, context)
+        states_ids = state_orm.search(cr, SUPERUSER_ID, [], context=context)
+        states = state_orm.browse(cr, SUPERUSER_ID, states_ids, context)
+        partner = orm_user.browse(
+            cr,
+            SUPERUSER_ID,
+            request.uid,
+            context).partner_id
```
nhomar: Wrong, set in the controller a browseable with superuser id is a security fault, can you set a logic in object side for such deal.

website_form_validations/controllers/main.py
```diff
@@ -0,0 +1,218 @@
...
+                    if not user_ids or request.website.user_id.id not in user_ids:  # noqa
+                        checkout.update(
+                            self.checkout_parse(
+                                "billing",
+                                order.partner_id))
+        else:
+            checkout = self.checkout_parse('billing', data)
+            try:
+                shipping_id = int(data["shipping_id"])
+            except ValueError:
```
nhomar: Try? Why a try in the middle of the method?

website_form_validations/controllers/main.py
oscarolar: This is not a security fault, if you carefully read the code you realize that this is made because you can buy something online without being logged in, this is when you just fill the checkout form with your data, this will cause a order with the portal user but with the data asociated to the order placed on line.

website_form_validations/README.rst
oscarolar: Done

website_form_validations/controllers/main.py
oscarolar: I cannot do the "Too much bussiness logic" thing, this is a very wired feature that requires a lot of data processing, this data is retrieved from the website, and because of the nature of the inherited controller this is necessary.

On the other hand I can try to split the code in several lines, but cant promise it will be good.

on the "To much IF" thing you can see this is not complex code and I'm not violating code complexity rules: https://en.wikipedia.org/wiki/Cyclomatic_complexity 
this was also checked by McCabe Complexity checker on my local instance: https://pypi.python.org/pypi/mccabe

website_form_validations/controllers/main.py
oscarolar: what is wrong with it?

website_form_validations/controllers/main.py
josemoralesp: The method can be separated, but the principal reasons of size are the pep rules, in some lines we can see the call of a method occupying [8 or more lines.](https://github.com/Vauxoo/addons-vauxoo/pull/401/files#diff-49d6a7001a0a953514b3bdf96264fc19R197).

website_form_validations/controllers/main.py
josemoralesp: To avoid use the try you can create a validation method and use the method is_digit

e.g.

def validate_shipping(ship):
    if ship[0] in ('-', '+'):
        return ship[1:].isdigit()
    return ship.isdigit()

website_checkout_validations/README.rst
```diff
@@ -0,0 +1,37 @@
...
+Contributors
+------------
+
+* Oscar Alcala <oscar@vauxoo.com>
+* Vauxoo <info@vauxoo.com>
+
+Maintainer
+----------
+
+.. image:: https://www.vauxoo.com/logo.png
```
nhomar, line 26: logo a s3

website_checkout_validations/README.rst
```diff
@@ -0,0 +1,37 @@
...
+    :alt: License: AGPL-3
+
+Website Form Validations
+========================
+
+This app allows you to render error messages on sale
+form after checking out on an online purchase, the main
+purpose is to show the messages instead of just making 
+red the input fields.
+
```
nhomar, line 11: Ponte unn printscreen antes y después aquí pliz.

nhomar:
I will merge. You can improve the comments in a different PR.

PR #402, opened 2015-08-06
[IMP][controller_report_xls]Reverted use to lxml by BeautifulSoup

luistorresm:


luistorresm:
@hbto 

Waiting only for runbot

Regards

hbto:
+1

nhomar:
Why is comming back the bs4? which one is the problem or the deficience with lxml?

nhomar:
BTW @luistorresm and @hbto this should be explained better in the body of the PR to not expend time asking for something that may be is not important.

nhomar:
I saw that the change bring problems, but it bring back 2 things.

1.- Reporst which depende of this module should be testes with unittest to allow validate everything is Ok always.
2.- This module itself should have a test.

@luistorresm I will merge this PR in order to trigger the tasks around tests and validate correctly the exclude of bs4.

Regards.

@hbto are you agreed with this plan?

hbto:
@nhomar so far, module was working like a charm with the inefficiency of the bs4.

Please leave it like that.

Thanks

nhomar:
@hbto 

Yes and no, we can not install bs4 in all repositories dude, sooner or later this dependency will die.

If no there are any reasonable reason to use an incorrect dependency then why force that?.

BUt BTW, yes it will staty like that until proper tests will be done.

And no it is not working like a charm in 500 modules declared this is the only one that is needing bs4 then it has a design problem in my opinion.

Regards.

hbto:
Gimme a reason why can not we install the library?

http://www.crummy.com/software/BeautifulSoup/

the thing is that controller is the only module using it, 
BUT the thing is this controller is being used by other module that translate HTML to xls.

What is the real problem?
I am using a library to do something that is neat and clear, extract data, 

why todo things by hand when libraries can do that job.

what is then the object of using libraries,

better do everything in assembler,  or punch cards.

nhomar:
On Thu, Aug 6, 2015 at 3:23 PM, Hbto [ Vauxoo ] notifications@github.com
wrote:

> better do everything in assembler, or punch cards.

Dude, to save 2 lines of code (and include 2 more with tries) you are
reimplementing all a logic?

bs4 and lxml have only differences in paradigm, with both of them you so
"exactly the same" in your case.... you are not taking a single approach of
the benefits here that bs4 can bring "Which is pretty print some stuff".

Then the technical reason is not good, your dacisiont in based on ignorance
in the use of lxml.

## 

Nhomar Hernandez
CEO Vauxoo.
Site: http://vauxoo.com
Twitter: @nhomar
Blog: http://geronimo.com.ve
LPUser: http://launchpad.net/~nhomar
Odoo Gold Partner
Móvil Venezuela:
+58 4144110269
Móvil México:
+52 1 4773933942

hbto:
@nhomar  in order to prove your point do not yell at my ignorance with more letters
provide code with _less_ line than mine do _not_create_ a new module based on lxml by trying
to prove I am wrong, if not, we are going to be two wrong guys.

Best Regards.

nhomar:
On Thu, Aug 6, 2015 at 4:13 PM, Hbto [ Vauxoo ] notifications@github.com
wrote:

> @nhomar https://github.com/nhomar in order to prove your point do not
> yell at my ignorance with more letters
> provide code with _less_ line than mine do _not_create_ a new module
> based on lxml by trying
> to prove I am wrong, if not, we are going to be two wrong guys.

cool

## 

Nhomar Hernandez
CEO Vauxoo.
Site: http://vauxoo.com
Twitter: @nhomar
Blog: http://geronimo.com.ve
LPUser: http://launchpad.net/~nhomar
Odoo Gold Partner
Móvil Venezuela:
+58 4144110269
Móvil México:
+52 1 4773933942

PR #415, opened 2015-08-17
[IMP][account_invoice_supplier_quantity_attachments]Added method search to field that get qty attachents

luistorresm:


JulioSerna:
:+1:

PR #416, opened 2015-08-20
Description field on task form view to html widget

sysadminmatmoz:
Since note brings a html description in task form after conversion, it seems only right to use a html widget on task description too.


sysadminmatmoz:
The travis problem is relative to website_order_confirm and not to my PR I think.

sysadminmatmoz:
... but I'll close this PR and open a fresh one from a freshly rebased repo, just in case

sysadminmatmoz:
New PR: https://github.com/Vauxoo/addons-vauxoo/pull/417

PR #420, opened 2015-08-26
[IMP][commission_payment] the payment is getting date with context_to…

JulioSerna:
…day() and commission with strftime() and there is different of timezone between function


JulioSerna:
@hbto 

I think with this PR we can solve the error in the test of commission payment when is executed in specific date,

The date of payment when you reconciled invoice demo with function `pay_and_reconcile` it is getting date with [fields.date.context_today()](https://github.com/odoo/odoo/blob/802d1abe7a446e4034f0af74d023acdd4dd45324/addons/account/account_invoice.py#L1133) and the date_start and date_stop of commission you are using time.strftime(),

I just propose to use the same method of time in creation of payment by function  `pay_and_reconcile`

hbto:
+1

PR #427, opened 2015-09-03
hide supplier invoice number column in tree view of Sale Invoices.

cjcruz:
Supplier invoice number in tree view of Sale Invoices has no has no sense.


hbto:
+1

PR #431, opened 2015-09-11
[FIX][user_story] Change the model in the inheritance

josemoralesp:
For prototype inheritance we need to add the compute fields in the main object because at the moment to use the create method the fields are not computed correctly only the fields are computed when we use  the write method


nhomar:
@josemoralesp this PR is valid yet?

PR #434, opened 2015-09-14
[ADD] website_product_filters: order products by popularity (times viewed) HU#925 CA#3

oscarolar:


nhomar:
@oscarolar can you call the field simply "viewed"? do you think is better?

oscarolar:
Yes, tought about it, but seems to me like a name for a boolean field, thats why I added the 'times_' prefix.

nhomar:
On Mon, Sep 14, 2015 at 1:19 PM, Olar Alca [Vauxoo] <
notifications@github.com> wrote:

> Yes, tought about it, but seems to me like a name for a boolean field,
> thats why I added the 'times_' prefix.

It sounds logical also, but if viewed <> 0 its behavior is like a boolean
btw.... then just saving characters and I think (subjective opinion) it is
more ellegant.

But I left you the final ups to you, just tell me and I merge as it is or
simply change it..

Regards.

## 

Nhomar Hernandez
CEO Vauxoo.
Site: http://vauxoo.com
Twitter: @nhomar
Blog: http://geronimo.com.ve
LPUser: http://launchpad.net/~nhomar
Odoo Gold Partner
Móvil Venezuela:
+58 4144110269
Móvil México:
+52 1 4773933942

oscarolar:
Something better came up: `views` it refers more to a number than a boolean. merge if you consider it good that way.

website_product_filters/models/product.py
```diff
@@ -0,0 +1,34 @@
...
+##############################################################################
+from openerp import models, fields
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    views = fields.Integer(
+        'Views',
+        help='This field shows the nomber of times a product has been\
```
nhomar: typo: s/nomber/number/

nhomar: In multiline helps it is better:

``` python
help="Line 1 "
         "Line 2 "
```

If you use the `\` it will regognize the tabs as blank spaces and will render incorrectly the tooltip.

nhomar:
@oscarolar Agreed with views (it was my first thought also).

I just made some little typo tips in the help.

After that :+1:

PR #440, opened 2015-09-18
[REF] website_product_availability: Fix AccessError in website tests

moylop260:


moylop260:
@nhomar
This fix avoid red errors in `website` test.
I don't know if is a good idea, but fix it

PR #442, opened 2015-09-21
[FIX] remove default to website_published field

hugho-ad:


website_product_availability/models/product.py
```diff
@@ -170,11 +170,10 @@ def get_states(self):
 class StockLocation(models.Model):
     _inherit = 'stock.location'
 
-    stock_alias = fields.Char('Alias Stock')
-    website_published = fields.Boolean(
-        'Available in the website',
-        copy=False, default="_get_default_alias")
-
     @api.one
```
oscarolar: No more @api.one, please make the fix for @api.multi

website_product_availability/__openerp__.py
```diff
@@ -5,13 +5,16 @@
     "author": "Vauxoo",
     "category": "Website",
     "website": "http://www.vauxoo.com/",
+    "license": "AGPL-3",
```
nhomar: LGPL

PR #445, opened 2015-09-22
[ADD] website_product_filters: filter by price range added on online store HU#925 CA#13

oscarolar:


oscarolar:
Here a screenshot Image

PR #454, opened 2015-09-24
[IMP] Brand filtering improved by using direct httprequest ar args HU#925 CA#9 CA#11

oscarolar:


oscarolar:
BEFORE:
BEFORE
AFTER:
AFTER

PR #471, opened 2015-10-07
[REF] .travis.yml: Standardize file changing tests and adding matrix.

KarenKawaii:


yaniaular:
@KarenKawaii is it missing   `- TESTS="1"`?

KarenKawaii:
My question is if i hace to add it por not, because in the previous commit the tests un env are tests="0"

yaniaular:
@KarenKawaii ok, we go to merge.

PR #473, opened 2015-10-09
Fix dependency name

alan196:
Module name changed in OCA


oscarolar:
This PR Must be rolled back, since we dont use the [OCA](https://github.com/OCA/e-commerce/tree/9.0/website_sale_product_brand) official module, instead we use [our own](https://github.com/Vauxoo/SerpentCS_Contributions-v8/tree/8.0/website_product_brand) with specific changes to our needs.

PR #481, opened 2015-10-19
[8.0][purchase_rfq_xls] Added new module to update prices of a RFQ using a xls template. HU#972

suniagajose:
The general purpose of this module is to update the prices of products of a RFQ using a xls template.

When an RFQ is created and then this is sent by email, this module attached to the mail an xls template without prices so that the provider fill their prices. Once send it to me back to the supplier module is able to allow load the template and update the prices of products taking into account the following considerations:
1. If the product in the RFQ is equal in quantity and code to the request answered ( Filled excel) then only the cost is updated.
2. If thhe product in the RFQ is equal in code but not in quantity then updated quantity and price.
3. If the Product is listed in my RFQ but the Supplier removed in their answered template, then this product will be removed of my RFQ.


moylop260:
@suniagajose 
Can you review the build status: https://travis-ci.org/Vauxoo/addons-vauxoo/jobs/86251691?

purchase_rfq_xls/README.rst
```diff
@@ -0,0 +1,4 @@
+Update Purchase RFQ using XLS
+=============================
+
+This module update and export purchase rfq using xls files..
```
nhomar: Please use the standard format for README.

purchase_rfq_xls/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+###############################################################################
+{
+    "name": "Update Purchase RFQ using XLS",
+    "version": "1.6",
+    "author": "Vauxoo",
+    "category": "Purchase",
+    "website": "http://www.vauxoo.com/",
+    "license": "AGPL-3",
```
nhomar: LGPL licence

purchase_rfq_xls/report/report.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class PurchaseQuotationReportXLS(osv.AbstractModel):
+
+    # As we are inheriting a report that was previously a particular report we
+    # have to keep it like that, i.e., we will keep _name the same than the
+    # original
+
+    # _name = `report.` + `report_name` (FQN)
+    # report_name="purchase_rfq_xls.report_template"
```
nhomar: Delete commented unused code.

purchase_rfq_xls/static/description/index.html
```diff
@@ -0,0 +1,52 @@
...
+      .backgrounds{background-color:#fff;color:#a41d35}
+  </style>
+</head>
+<section class="oe_container">
+ <div class="oe_row oe_spaced">
+  <h2 class="oe_slogan">
+   Update Purchase RFQ using XLS
+  </h2>
+  <p class="oe_mt32">
+   This module update and export purchase RFQ using xls files
```
nhomar: I think you can write here the same text on the PR it is good.

purchase_rfq_xls/wizard/wizard.py
```diff
@@ -0,0 +1,187 @@
...
+        new_products = []
+        for row in range(eof):
+            if not can_start:
+                if col_start == sheet.cell_value(row, 0):
+                    can_start = True
+                continue
+            # External ID
+            xml_id = sheet.cell_value(row, 0)
+            # Model
+            # default_code = sheet.cell_value(row, 1)
```
nhomar: delete unused code.

Issue #496, opened 2015-10-27
[IMP] warehouse_interim_receipt module: Warehouse Receipt Input wizard filter by location


zaoral:
module: warehouse_interim_receipt
version: 8.0.1.0.0

**Current Behavior**
- Go to `Reporting / Warehouse / Warehouse Receipt Input` and run wizard.
- Show me all the move lines

**Expected Behavior**
- Expect to filter only the moves lines form Supplier location to Input location.
# TODO

I need that the warehouse receipt input wizard manage a source/destination location to filter the result move lines. This to ensure that the retrieved moves are really the ones for input operations.

In theory we want to filter (by default) from Supplier Location (Source) to Input Location (Destination).
For be a generic module the default source/destination location of this wizard could be defined in a the stock settings wizard.
- [x] add location_id field to the whr input wizard  
- [x] add location_dest_id field to the whr input wizard  
- [x] update whr input wizard to take into account location_id and location_dest_id when they are defined.

zaoral:
new approach will be apply by using picking type instead locations fields.

PR #497, opened 2015-10-27
[WIP] manage a source and dest location in the warehouse receipt input wizard.

zaoral:
Try to solve the issue #496
- [ ] Fix: try to manage default form stock.config.settings but find and error.
- [ ] Add missing warehouse configuration to enable WH/Input location
- [ ] Check if is required to add new data demo for the Wh/Input location.


zaoral:
This approach will not be use. after meeting with @nhomar  we agree to use picking type field instead two location fields.

PR #499, opened 2015-10-27
[ADD] add picking type as one of the parameter to generate the stock move search. Now the warehouse receipt input report can be filter by picking type

zaoral:
Fixes https://github.com/Vauxoo/addons-vauxoo/issues/496


zaoral:
@nhomar please review this merge.

PR #501, opened 2015-10-27
8.0 account move line production hbto

hbto:
Migration of module account_relation_move

This module has been outstripped as 
part of its functionality has already being deployed by 
stock_move_entries,

And due to its name does not recall any meaningful functionality 
a new module with its main functionality has been created

account_move_line_production

old module account_relation_move has been kept only to provide
backward compatibility.

Best Regards.


JulioSerna:
:+1:

PR #510, opened 2015-10-30
[FIX] less space used and better look and feel, removed searchbox, placed all items to fullfill space on website

oscarolar:


oscarolar:
### BEFORE:

BEFORE

### AFTER:

AFTER

PR #512, opened 2015-11-01
Revert "[ADD] add new module purchase_changeless_move_lines to restri…

moylop260:
Reverts Vauxoo/addons-vauxoo#296

PR #513, opened 2015-11-02
Revert "Revert "[ADD] add new module purchase_changeless_move_lines t…

moylop260:
Reverts Vauxoo/addons-vauxoo#512


moylop260:
This PR is the revert of revert...
Because I found other error: https://github.com/Vauxoo/yoytec/issues/266
Then maybe the last reverted PR don't fix the real error.

moylop260:
The problem is in other module: `purchase_order_line_sequence`
I will create a new issue for that.

Then merging this PR

PR #520, opened 2015-11-05
[FIX] Phantom tour 

oscarolar:


oscarolar:
Runbot build https://github.com/Vauxoo/yoytec/pull/287

PR #521, opened 2015-11-05
[FIX]: change field names and sele.config.settings is selection in Odoo9

alan196:


invoice_discount/model/config_settings.py
```diff
@@ -34,10 +34,10 @@ class SaleConfigSettings(osv.TransientModel):
     _inherit = 'sale.config.settings'
 
     def get_default_sale_config_settings(self, cr, uid, fields, context=None):
-        return {'group_discount_per_so_line': True}
+        return {'group_discount_per_so_line': 1}
```
moylop260, line 37: @alan196 
What happens with `1` instead of `True`?

alan196:
@moylop260 The field group_discount_per_so_line is not longer a boolean in Odoo 9, now it is a selection field with two options '0' and '1', so if you try to install this module in Odoo 9 you get error with True.

PR #533, opened 2015-11-17
[REF] Removed disabled class from add to cart button, instead hidden to show only the button when product is available

oscarolar:
Fixes https://github.com/Vauxoo/yoytec/issues/336


oscarolar:
Building here https://github.com/Vauxoo/yoytec/pull/379

Issue #535, opened 2015-11-18
[8.0] user_story: "AssertionError: [] is not True : The message was not created"


moylop260:
In build 8.0 of [instance](https://magnum.travis-ci.com/Vauxoo/instance/jobs/25249245#L1049-L1050) of runbot show next traceback

``` bash
2015-11-18 15:46:21,479 117 ERROR openerp_test openerp.addons.user_story.tests.test_user_story: FAIL
2015-11-18 15:46:21,480 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: test_approve_button (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:22,672 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: test_copy_method (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:23,667 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: test_create_method (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:24,509 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: test_unlink_method (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:25,524 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: test_write_method (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:26,517 117 INFO openerp_test openerp.addons.user_story.tests.test_user_story: ======================================================================
2015-11-18 15:46:26,518 117 ERROR openerp_test openerp.addons.user_story.tests.test_user_story: FAIL: test_acceptability_criterial_buttons (openerp.addons.user_story.tests.test_user_story.TestUserStory)
2015-11-18 15:46:26,518 117 ERROR openerp_test openerp.addons.user_story.tests.test_user_story: Traceback (most recent call last):
2015-11-18 15:46:26,518 117 ERROR openerp_test openerp.addons.user_story.tests.test_user_story: `   File "/root/odoo-8.0/openerp/tools/misc.py", line 1139, in deco
2015-11-18 15:46:26,518 117 ERROR openerp_test openerp.addons.user_story.tests.test_user_story: `     return func(*args, **kwargs)
[...]

PR #536, opened 2015-11-18
[WIP] find cause of error reported in Vauxoo/addons-vauxoo#535 

zaoral:
Fix Vauxoo/addons-vauxoo#535

At runbot the message is created after clicking the approve of the AC.
It can be seen because after click the approve button a message is added to the user story message log.
### Finding the error
- [x] make a runbot test tryting to approve a acceptability criteria and check if the message appers in the user story message log.

> The message appears. The message is been created.
- [x] update unit test to check if the criteria is really accepted. Maybe the message has not been created because the criteria really is not been approved.

> The criteria was really approved
- [x] download a local copy and run the instance to check from the interface if the message is created.

> The messege is created and added to the user story message log.
- [x] the unit test use a text in spanish to make the comparation. Maybe this is the problem. use the original text and use the translation _ function to manage the compare (use pdb)

> **The error was found**

NOTE: Another error was throw. is the same error. so will be fixed and push in this same solution.

### DIFF
--- a/user_story/__openerp__.py
+++ b/user_story/__openerp__.py
@@ -43,8 +43,8 @@
         "demo/demo.xml"
     ],
     "data": [
-        #"data/data_us_report.xml",
-        #"report/user_story_report_view.xml",
+        # "data/data_us_report.xml",
+        # "report/user_story_report_view.xml",
         "security/userstory_security.xml",
         "security/ir.model.access.csv",
         "view/userstory_view.xml",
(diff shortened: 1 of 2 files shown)


zaoral:
@moylop260 this PR was a red travis for a LINT_CHECK. The user story module is using api v7. To fix the odoo lint error the module need to be change to api v8 and could generate another issues. I think that for this PR in particular we should ignore this red and create a migration task for the module in a future.

PR #544, opened 2015-11-19
[IMP][aging_due_report] change the copy by create because is causing …

JulioSerna:
…conflict with new api with the  super of copy in project module


hbto:
+1

PR #548, opened 2015-11-24
[FIX] website_product_filters: Breadcrumb and attribute tags of filters now are shown in the same line

oscarolar:
BULID https://github.com/Vauxoo/yoytec/pull/407
Fixes https://github.com/Vauxoo/yoytec/issues/387

image


friedi1:
Oscar could you please put the lables of the filters directly behind the breadcrumb not with a big distance like shown on example. In the example it would mean directly behind "All Products" there would be the 16GB Label, followed by the rest.

oscarolar:
@friedi1 yeah, sure.

oscarolar:
@friedi1 https://github.com/Vauxoo/addons-vauxoo/pull/549 there you can review.

PR #550, opened 2015-11-25
[WIP] - [ADD] badges with qty of products per category

oscarolar:
BUILD https://github.com/Vauxoo/yoytec/pull/409
- [x] "Show more/show less" should only be shown when more then 5 filters available in category (see image 1)
- [x] Filters should be collapsed to 5 visible with "Show more/show less" button if more then 5 filters available. Right now it just shows just one filter of the ones available
- [x] Move active Filters to Breadcrumb line. Right now they are below

This PR is related to issue vauxoo/yoytec#344

### DIFF
--- a/website_product_filters/models/product.py
+++ b/website_product_filters/models/product.py
@@ -83,9 +83,11 @@ class ProductCategory(models.Model):
 
     product_ids = fields.Many2many('product.template', compute="_get_products")
     total_tree_products = fields.Integer("Total Subcategory Prods",
-                                         compute="_get_product_count")
+                                         compute="_get_product_count",
+                                         store=True,)
 
     @api.multi
+    @api.depends('product_ids')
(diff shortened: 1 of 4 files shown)


nhomar:
el campo product_ids en la categoría pública se muestra SOLAMENTE lo que tienes existencia no?, por que es el objetivo, no solamente que "existen" los productos....

Es lo que entiendo en el template que modificaste sucederá:

Todos los productos de la categoría.

Cuando creo es.

Todos los productos publicos de la categoría que tengan existencias.

Saludos.

oscarolar:
No its not, you can show products that are not available but published, so the customer can see when are they going to be available again.
so the final criteria must be: products with [('website_published', '=', True)]

oscarolar:
Can you add the label WIP to this PR Please @nhomar

PR #555, opened 2015-11-29
[CC] Getting rid of non-necessary overriding of method compute_price on

hbto:
model product.template as the overriding one does the same as original one


hbto:
Please @josemoralesp 
Validate this assertion.

Best Regards.

josemoralesp:
:+1: @hbto

Issue #556, opened 2015-12-02
Module mrp_default_location stops inconsistent database


nestoravalos:
I install the module mrp_default_location, buy this stops inconsistent database.

This the error in the other modules:

Traceback (most recent call last):
  File "/opt/odoo/openerp/http.py", line 537, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/openerp/http.py", line 574, in dispatch
    result = self._call_function(*_self.params)
  File "/opt/odoo/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/opt/odoo/openerp/service/model.py", line 118, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/opt/odoo/openerp/http.py", line 307, in checked_call
    return self.endpoint(_a, *_kw)
  File "/opt/odoo/openerp/http.py", line 803, in __call__
    return self.method(_args, *_kw)
  File "/opt/odoo/openerp/http.py", line 403, in response_wrap
    response = f(_args, *_kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 944, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/opt/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/opt/odoo/openerp/api.py", line 345, in old_api
    result = method(recs, _args, *_kwargs)
  File "/opt/addons-vauxoo/sale_team_warehouse/models/sales_team.py", line 34, in _default_get
    res_users_brw.default_section_id.default_warehouse.id
AttributeError: 'crm.case.section' object has no attribute 'default_warehouse'

Regards

PR #558, opened 2015-12-02
[DEL] deprecated modules in favor to use mrp_location_route module

JulioSerna:
https://github.com/Vauxoo/addons-vauxoo/issues/556


nhomar:
@JulioSerna 

Just waiting for customers instances tu ensure all ok. For me +1.

PR #559, opened 2015-12-02
[IMP][product_exetended_variants] changed replace of month in datetim…

JulioSerna:
…e because cause error in month 12


josemoralesp:
:+1:

PR #573, opened 2015-12-10
[FIX] Fix the My user stories execute menu

zaoral:
Fix Vauxoo/instance#202


zaoral:
@nhomar the travis is red because this module use api v7
This PR resolve is a menu bug, Please we can pass this red and merge?

Best Regards.

PR #575, opened 2015-12-12
[FIX] website_product_filters: rating default set to zero, it was rating with 3 without comments

oscarolar:
Build https://github.com/Vauxoo/yoytec/pull/534


oscarolar:
This gives a green on the build https://github.com/Vauxoo/yoytec/pull/534 runbot is green but travis times out:
MERGE

PR #579, opened 2015-12-15
[DEL][sale_cash_register]deprecate module because is moved to another…

JulioSerna:
… repository


paraita:
Hi ! where was the module moved ? I'd like to try it please !

PR #594, opened 2015-12-28
[IMP] user_story: Adding task in the user story will add it at the top of the list.

zaoral:
This to improve the time spent to add more task to a user story.
Do not need to scroll down in the user story page.


zaoral:
@nhomar Please review this PR.

PR #595, opened 2015-12-29
8.0 stock card merge hbto

hbto:
Negative Stocks are recomputed properly


hbto:
Merge as Runbot is green and previous run of travis were green though latest one is red because of timeout

PR #598, opened 2015-12-30
[FIX]invoice_datetime port to new API

alan196:


invoice_datetime/models/account_config_settings.py
```diff
@@ -0,0 +1,84 @@
...
+        return user.company_id.id
+
+    def get_default_select_date(self, cr, uid, fields_name, context=None):
+        key_by_company_id = "acc_invoice.date_invoice_type_" + str(
+            self._default_company(cr, uid))
+        type_show_date = self.pool.get("ir.config_parameter").get_param(
+            cr, uid, key_by_company_id, default='date', context=context)
+        return {'select_date': type_show_date}
+
+    # @api.multi
```
alan196, line 50: I have a question here, when I try to migrate this function to new API it doesn't work... and I think it's the same result.

nhomar, line 50: remove the @api.multi I think it is not necesary on v9. but not sure

nhomar:
This módule will not be mantained anymore, for v9.0 we should try to use th date as it is.

nhomar:
@alan196 If you considere this must continue open, please cleanup and reopen.

PR #601, opened 2015-12-30
8.0 stock card merge hbto

hbto:
Translation and UnitTest for negative quants


hbto:
Runbot Green For Lodi

PR #602, opened 2015-12-30
8.0 stock landed costs average merge hbto

hbto:
Allows to use Landed Cost Module on Products with Costing Method Average & Standard


hbto:
Failure in Runbot is unrelated to modules being tested.

That is related to change in Date, as the end of the official date approaches (UTC time) sensible fields to date changes tend to fail.

"/root/build/Vauxoo/lodigroup/bias_import_sale_data/tests/test_import_sale_from_csv.py", line 40, in test_01_create_perfect_file
2015-12-31 00:01:25,166 109 ERROR openerp_test openerp.addons.bias_import_sale_data.tests.test_import_sale_from_csv: ` 'The date was no correctly set')

PR #604, opened 2015-12-31
8.0 stock landed costs segmentation merge hbto

hbto:
stock_landed_costs_segmentation module is added with demo data & UnitTests


hbto:
Runbot went green

PR #605, opened 2015-12-31
[REF] account_move_line_production: add demo content, add unit test based on demo data

hbto:


hbto:
Runbot has gone green

PR #610, opened 2016-01-05
8.0 mrp workcenter account move merge hbto

hbto:


hbto:
Runbot Green for this PR

PR #624, opened 2016-01-08
[FIX] avoid copy account_move_id when mrp.production is duplicated

osvalr:
## Changes made:
- [X] Disable copy for `account_move_id` field
- [X] Add unit test to validate it


osvalr:
@hbto review asayc

PR #629, opened 2016-01-12
[IMP] stock_easy_internal_transfer: Enable suggest move lines

suniagajose:
## Features

Added a boolean field to enable suggest move lines when select a source location. This allow load automatically all products available in location selected, so avoid reducing manual loading times. 

Note: Is recommeded when normally you need move all products available in a location


nhomar:
@josemoralesp can you give your +1¿?

@suniagajose 7 days without feedback for a PR on this case in unaceptable, please move forward on your side in order to be sure it can be mergeable.

Regards.

suniagajose:
@nhomar i'm sure it is mergeable, just waiting for @josemoralesp  +1 .... meanwhile i have moved forward with fiscal printer

josemoralesp:
:+1: Sorry for the delay

PR #630, opened 2016-01-12
8.0 stock card segmentation fixing sgmnt

hbto:
Fixing Segmentation Costs on Stock Card


hbto:
Runbot has gone green

PR #631, opened 2016-01-12
[REF] website_product_filters: display only one level of subcategories on categories column HU#1006 CA#27

oscarolar:
This PR is part of this whole issue:  https://github.com/Vauxoo/yoytec/issues/489
Build https://github.com/Vauxoo/yoytec/pull/631
- [x] After mergeing this merge this too: https://github.com/Vauxoo/SerpentCS_Contributions-v8/pull/20


website_product_filters/models/product.py
```diff
@@ -80,11 +80,10 @@ class ProductCategory(models.Model):
         "product_template_id", readonly=True)
     total_tree_products = fields.Integer("Total Subcategory Prods",
                                          compute="_get_product_count",
-                                         store=True,)
+                                         store=False,)
```
oscarolar, line 83: @moylop260 dude, this is ok for you?

moylop260, line 83: @oscarolar 
I don't have a full context of this change, but the `store` param affect directly the instance migrated, you should ask to @josemoralesp

oscarolar, line 83: @josemoralesp I made this change, when I coded this it was intended to be `store=False` do you have any problem with this?

oscarolar:
This is what has been achieved so far in order to solve https://github.com/Vauxoo/yoytec/issues/489:
iamge 

@nhomar any comments are welcome

oscarolar:
So far with no external dependencies installed (yoytec in this case) this is how the App works now: 
NOE

website_product_filters/models/product.py
nhomar, line 83: @oscarolar I prefere store = True with a good trigger, but I do not know if it is technically possible, compute this field any time view is loaded is too time consuming.

Can you propose a solution for that?

website_product_filters/controllers/main.py
```diff
@@ -45,27 +46,55 @@ def shop(self, page=0, category=None, search='', ppg=False, **post):
...
+                                                  reverse=True),
+                'hottest':
+                    filtered_products.sorted(key=lambda r: r.create_date),
+                'rating': filtered_products.sorted(key=lambda r: r.rating),
+                'popularity': filtered_products.sorted(key=lambda r: r.views)}
+            if post.get('product_sorter', '0') != '0':
+                sortby = post['product_sorter']
+                res.qcontext['sortby'] = sortby
+                ordered_products = keys.get(sortby)
+            elif request.httprequest.cookies.get('default_sort', 'False') != 'False':  # noqa
```
moylop260, line 89: `# noqa?`

oscarolar, line 89: Just because I didn't wanted to split the line for a few characters longer

moylop260:
:+1:

moylop260:
We will merge it
After, @hugho-ad will creates change to IMP performance.

PR #636, opened 2016-01-15
8.0 redirect quant cost computation to sgmnts osval

hbto:
Computation on Standard Products takes into account segmentation on products of Costing Method AVG, even recomputes its segments, and write standard_price field properly.

### DIFF
--- a/product_extended_segmentation/__openerp__.py
+++ b/product_extended_segmentation/__openerp__.py
@@ -28,7 +28,7 @@
     "depends": [
         'product',
         'product_extended',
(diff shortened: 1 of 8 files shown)


hbto:
This PR has already been tested by product owner and is green on product owner runbot

hbto:
@nhomar 

can you check?

hbto:
This PR has already been greened on Product Owner CI, it is going to be merged
though been merged does not mean that it is deployed on Product Owner Instance

PR #648, opened 2016-01-18
[FIX]-product_do_merge: Don't consider weight when comparing stock moves

ghost:


nhomar:
Hello,

Thanks for contribuye, can you extender The use case where this pr solve? I do not understand exactly why replace the logic instead correct it.

Thanks again

nhomar:
@josemoralesp can you give your green light here?

josemoralesp:
Hello @lg-clearcorp 

Can you please tell me what are you trying to do here?

I don't understand the reason for your change.

Best Regards.

luisg123v:
Contributor doesn't exist anymore, so closing.

PR #649, opened 2016-01-18
[FIX] account_move_line_extended: Replace old reconcile field by the new one reconcile_ref in 8.0

KarenKawaii:
- [x] Module loaded successfully after replace the old field `reconcile` by the new one `reconcile_ref` in branch 8.0:

account_move_line_extended

### DIFF
--- a/account_move_line_extended/view/account_move_line_view.xml
+++ b/account_move_line_extended/view/account_move_line_view.xml
@@ -34,7 +34,7 @@
                     <field name="amount_currency"/>
                     <field name="currency_id"/>
                     <field name="date_maturity"/>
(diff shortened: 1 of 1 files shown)


nhomar:
@KarenKawaii I am merging I checked and the only customer which is using this module is the Oil Company (which is under the v7.0 and will be migrated next year), but the name is too generic, I should call it "account_move_line_multicurrency"  @hbto set this in yout todo for that date (just save this email) I am merging just in order to have it installable on v8.0 (but it will not worl on v9.0).

KarenKawaii:
@nhomar  got it! I have referenced your email like a note in the spreadsheet

PR #653, opened 2016-01-19
[FIX] account_reconcile_search: To allow locate element in parent view.

KarenKawaii:
- [x] Module loaded successfully after replacing `<xpath expr="//group[@string='Group By...']` by the new one `<xpath expr="//group[@string='Group By']` in branch 8.0

account_reconcile_search
- [x] Add the field `reconcile_ref` in search options.

For references of this change visit: [view_account_move_line_filter v7.0](https://github.com/odoo/odoo/blob/7.0/addons/account/account_view.xml#L1169) and [view_account_move_line_filter v8.0](https://github.com/odoo/odoo/blob/8.0/addons/account/account_view.xml#L1258)

### DIFF
--- a/account_reconcile_search/view/account_reconcile_search.xml
+++ b/account_reconcile_search/view/account_reconcile_search.xml
@@ -7,6 +7,9 @@
             <field name="model">account.move.line</field>
             <field name="inherit_id" ref="account.view_account_move_line_filter"/>
             <field name="arch" type="xml">
+                <xpath expr="//field[@name='partner_id']" position="after">
+                    <field name="reconcile_ref"/>
+                </xpath>
                 <xpath expr="//group[@string='Group By']/filter[@string='Period']" position="after">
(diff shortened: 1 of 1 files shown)


nhomar:
@KarenKawaii 

This module is usefull but the nam is not in line with the behavior.

As you saw the behavior is add the "group by." but th module is names "... reconcile_search"

Can you please add the field "reconcile_ref" also as a filter to search in the search view in order to be consistent with the module name?

Thanks.

nhomar:
@KarenKawaii just to be sure, I am waiting you add an extra xpath adding the field reconcile_ref in the search options-.

Is that clear? it is a 3 lines change in order to put this mergeable.

KarenKawaii:
@nhomar ...so what happen to this filter:
[filter](https://github.com/vauxoo-dev/addons-vauxoo/blob/8.0-vx%234604-account_reconcile_search-dev-karenkawaii/account_reconcile_search/view/account_reconcile_search.xml#L11)

It has a filter Reconcile too.

nhomar:
See Image

PR #660, opened 2016-01-20
[FIX] stock_picking_security_force: Change view to make it compatible for v. 8.0

KarenKawaii:
- [x] Module loaded successfully after changing view to make it compatible for v. 8.0 in `stock_view.xml`:
  - [x] Remove records relating to xml ids that no longer exist in v. 8.0: `view_picking_in_form` and `view_picking_out_form`.
  - [x] Use the buttons in v. 8.0. instead of those used in v. 7.0

stock_picking_security_force

### DIFF
--- a/stock_picking_security_force/stock_picking_security_force_view.xml
+++ b/stock_picking_security_force/stock_picking_security_force_view.xml
@@ -1,42 +1,20 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <openerp>
     <data>
-        
-        <record model="ir.ui.view" id="view_stock_picking_inherit_button_force_inc_form">
-            <field name="name">view.stock.picking.inherit.button.force.inc.form</field>
-            <field name="model">stock.picking</field>
-            <field name="inherit_id" ref="stock.view_picking_in_form"/>
-            <field name="arch" type="xml">
-                <xpath expr="//button[@name='force_assign']" position="replace">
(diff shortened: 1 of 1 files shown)


stock_picking_security_force/stock_picking_security_force_view.xml
```diff
@@ -1,42 +1,20 @@
...
             <field name="inherit_id" ref="stock.view_picking_form"/>
             <field name="arch" type="xml">
-                <xpath expr="//button[@name='draft_validate']" position="replace">
-                    <button name="draft_validate" states="draft" string="Process Now" type="object" icon="gtk-jump-to" groups="stock.force_availability"/>
+                <xpath expr="//button[@name='action_confirm']" position="replace">
+                    <button name="action_confirm" states="draft" string="Process Now" type="object" groups="stock.force_availability"/>
                 </xpath>
                 <xpath expr="//button[@name='force_assign']" position="replace">
-                    <button name="force_assign" states="confirmed" string="Force Availability" type="object" icon="gtk-jump-to" groups="stock.force_availability"/>
+                    <button name="force_assign" states="confirmed,waiting,partially_available" string="Force Availability" type="object" groups="stock.force_availability"/>
```
nhomar, line 14: This change, is putting the button in different states whcih are not the original ones, Is there any reason because it happens?

In other hand, this module is wrongly done, the group can be added with a simple attribute laben and not replacing the original xml entry.

This feature is available on @sabrina's project @sabrina can you tell us which modules are doing this features there? because I see this module a litle redundant.

Regards.

luisg123v:
This is abandoned, so closing

PR #666, opened 2016-01-21
[FIX] For average Costing Method Products book the actual landed cost

hbto:
[FIX] For average Costing Method Products book the actual landed cost. 
Leave for later booking COGS in a new and improved method


hbto:
Due to the strong believe of my colleague @nhomar  on the occult and all that $#!+ 
I must stress that I don't believe that crap, But the audience has demanded it
I have accepted to respect those believe not mine

PR #672, opened 2016-01-22
Commented invoice_validate because it was affecting the performance.

ruiztulio:
The pgbadger output are available here:
Before commenting and 11 workers: http://192.99.47.168/apex80_pgbadger.html
Before commenting and 0 workers: http://192.99.47.168/apex80_pgbadger_1worker.html
After commenting and 0 workers: http://192.99.47.168/apex80_pgbadger_commented.html

This is the branch that is currently in Apex 8.0 instance.

This test was made with @josemoralesp @JulioSerna @nhomar and me.


nhomar:
Guys.

This PR is done expecting:

1.- Unit test to validate the behavior of this method commented with 1 invoice + 2 product lines correctly configured. (AS METHOD IS NOW).
2.- Once (1) is done then refactor the method which will comply with the approach expected.

## Technical Notes during the review:
1. This method is triggering also the commision payment computation for [_date_last_payment](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/commission_payment/model/account_invoice.py#L7), then it is increasing a lot the SQL consumption.
2. May be this PR will include a refactor there also with some selective computation and writing of th AML in order to avoid recompute the field "date_last_payment" with a reconciliation which involve only stock accrual move lines.
3. I think the excess of recursive checking on the mthod can be fixed with a good SQL please @josemoralesp take care of that.
4. Use new API may be brings benefits.
5. @hbto Please document the use case for the test in order to avoid any regression, and try to make the use case BEFORE the change of the behavior, I do not want any regression due to this refactor.
   @JulioSerna pay atention to this case maybe you can help also.

Regards and thanks everybody.

luisg123v:
@ruiztulio can we close this PR? Is this branch still used in apex?

luisg123v:
I couldn't find any reference of this branch in Apex, so it seems it's not used anymore. Besides, it has conflicts from a long time ago, so I'll close this PR.

PR #674, opened 2016-01-23
[FIX] stock_invoice_state_editable: Fixes to make it compatible with v. 8.0

KarenKawaii:
- [x] Module successfully installed after doing the following fixes:
  - [x] The `invoice_state` field belongs to `stock_account` module in v. 8.0. So, that dependency was added to manifest.
  - [x] The views `view_picking_in_form` and `view_picking_out_form` no longer exist in v.8.0. The `view_picking_form` still remains in this version. Check as a reference the next migration and conversation: [migration example](https://github.com/OCA/account-invoicing/pull/62/files), [conversation](http://stackoverflow.com/questions/26504057/odoo-external-id-not-found-in-the-system-stock-view-picking-out-form).
  - [x] Change xml id reference in view to `view_picking_inherit_form2` because this view add the `invoice_state` field to the parent view `view_picking_form`.

stock_invoice_state_editable
- [x] Use `stock_account` as dependency instead `stock`.


stock_invoice_state_editable/__openerp__.py
```diff
@@ -31,7 +31,8 @@
     "website": "http://www.vauxoo.com/",
     "license": "AGPL-3",
     "depends": [
-        "stock"
+        "stock",
```
nhomar: if you have stock_account you do not need stock.

KarenKawaii:
@nhomar done... you are right, my bad. Sorry for the mega-delay.

PR #677, opened 2016-01-26
[8.0][mrp_partial_production] Added new module to partial production

josemoralesp:
In some case is needed start a production with partial reservation because you
can deliver your order in different parts. This module allow process a
manufacturing order with partial reservation, producing multiple times until
the order is completed.

The module add a new compute field to show the quantity available to produce
considering the number of the reservation in the products to consume. This
field is considered in all process and you cannot produce more than this field
indicates.

This module don't allow the creation of negative quant, this mean you cannot
force the production.


mrp_partial_production/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import models
+from . import wizard
+# from . import tests
```
nhomar, line 30: tests is commented, that's ok?

josemoralesp, line 30: Yes, the test files into the folder tests are executed automatically[1]. It is unnecessary  import the folder.

https://drive.google.com/a/vauxoo.com/file/d/0B2kzKLGF6ZvLalF0Q21CNjJ5bGc/view?usp=drivesdk

josemoralesp:
@nhomar Can you merge this please.

@JulioSerna  and Juan Aguilar tested the functionality and they approve the module.

PR #679, opened 2016-01-26
[FIX] _dispatch method in overwrite in website_product_filters

hugho-ad:
FIX https://github.com/Vauxoo/yoytec/issues/626


oscarolar:
:+1:

PR #680, opened 2016-01-27
[FIX] sale_test_data_imp: Fix import error and avoid warnings like LocalService() deprecated.

KarenKawaii:
- [x] Module loaded successfully after replacing `import tools` by `from openerp import tools` for example.
- [x] Follow the next recomendation `For workflows, instead of using
  LocalService('workflow'), openerp.workflow should be used` to avoid warning `LocalService() is deprecated since march 2013`.

sale_test_data_imp

### DIFF
--- a/sale_test_data_imp/test/sale_order_product_can_be_sold.yml
+++ b/sale_test_data_imp/test/sale_order_product_can_be_sold.yml
@@ -4,8 +4,8 @@
     Create sale order, picking in and invoice by product
 -
   !python {model: sale.order}: |
-    import tools
-    import netsvc
+    from openerp import tools
+    from openerp import netsvc
     import os
     import tempfile
     import csv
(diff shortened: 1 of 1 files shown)


sale_test_data_imp/test/sale_order_product_can_be_sold.yml
```diff
@@ -109,7 +109,7 @@
             #~ Validate invoice for this sale order 
             if invoice_id:
                 try:
-                    wf_service = netsvc.LocalService("workflow")
+                    wf_service = netsvc.openerp.workflow
```
zaoral, line 112: @KarenKawaii Also you could use this method `invoice_recordset.signal_workflow('signal_name')`. Example [here](https://github.com/odoo/odoo/blob/52484b0132ee3e0063dd67cc3c2eb434f99f5db8/addons/account/wizard/account_invoice_state.py#L22). If you choose use the mention method to call a workflow signals for a record then you will not need to make `import openerp.workflow` and `import openerp.netsvc` in your python files.

I hope this information be useful for you.

Best Regards.

PR #699, opened 2016-01-31
[IMP] Add forcing new cron logging products computed products in order to analyze results.

nhomar:


nhomar:
[TODO]
- [ ] UNITTEST FOR CRON METHOD.
- [x] FIX FLAKE.

PR #706, opened 2016-02-02
[IMP][sale_order_date_due]Added module to can set the date due in sale

luistorresm:
order, and this value is fill by default


luistorresm:
@moylop260 

En una oportunidad puedes revisar esta PR?

PR dummy https://github.com/Vauxoo/yoytec/pull/622.

Saludos

PR #707, opened 2016-02-02
[WIP][8.0] Enhance coverage and unit testing for stock modules

osvalr:
## Changes:

The following image reflects expected demo values for stock_card
image
- [x] Added demo data with customer and supplier product returns
- [x] Added unit test to product return context in both directions
- [X] Add unit test cases when there is not Gain & Loss accounts
- [X] Add unit test cases for user validation messages
- [x] Add unit test cases when aditional cost in valuations lines goes negative
- [X] Split stock move in time to let stock_card show transactions as it should


osvalr:
@hbto as you can see in [here](https://travis-ci.com/Vauxoo/lodigroup/jobs/32210813) all this is ready to merge, the current error showed in Line 2482 is not part of this

PR #715, opened 2016-02-04
[ADD] Add reporting issue guidelines

zaoral:
There is not example template for users trying to report bugs of how do it or what information put in the report.

Currently they click over the `New Issue` button to found a yellow barnner with a link to the guidelines to contribute
image

However this link lead to a [document](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/CONTRIBUTING.md) that do not contain any information for the user of how  reporting issues.

We need to add the `Reporting Issue` information similar to the one that Odoo give in his [wiki](https://github.com/odoo/odoo/wiki/Contributing#reporting-issues) pages.

At first step we can update addons-vauxoo contribution [file](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/CONTRIBUTING.md) adding a copy & paste of the Reporting Issue section form odoo wiki.

At a second phase,  I think this information is needed not only in this repository but in all our repos, public and private. This way our clients could have a very clear example of how report issues for their instances. We can do the same from step 1 for all our repositories or even better we could use some centralization way to add the same information and be able to see to all our repositories, something like Reporting Issues Vauxoo guideline.

Best Regards.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,39 @@
+Reporting Issues
+----------------
+
+1. Make sure you've actually read the error message if there is one, it may really help
+2. No need to create an issue if you're [making a PR](#making-pull-requests) to fix it. Describe the issue in the PR, it's the same as an issue, but with higher priority!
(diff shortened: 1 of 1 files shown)


zaoral:
@nhomar @moylop260 please consider the second step of this issue.
Best Regards.

PR #726, opened 2016-02-12
[FIX][sale_order_due_date]Fixed call to models.Model by TransientModel

luistorresm:


luistorresm:
@JulioSerna 

Puedes darme +1?

Saludos

PR #731, opened 2016-02-16
[FIX] stock_purchase_requisitor: Remove references to stock.picking.i…

KarenKawaii:
…n and stock.picking.out cause that models are depreciated by stock.picking. Also change xpath expr to locate new field in parent view.
- [x] Module loaded successfully after using the proper model and changing the xpath expression to locate the new field:

stock_purchase_requisitor


luisg123v:

PR #732, opened 2016-02-16
[FIX] website_product_comment_purchased: Fix performance issue when create a mail_message

hugho-ad:
Fix https://github.com/Vauxoo/yoytec/issues/716
Fix https://github.com/Vauxoo/yoytec/issues/671

### DIFF
--- a/website_product_comment_purchased/models/mail.py
+++ b/website_product_comment_purchased/models/mail.py
@@ -22,37 +22,40 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp import models, fields
+from openerp import models, fields, api
 
 
 class MailMessage(models.Model):
     _inherit = 'mail.message'
 
-    def _comment_bought(self, cr, uid, ids, name, arg, context=None):
-        res = {}.fromkeys(ids, False)
-        sale_report_obj = self.pool.get('sale.report')
-        mail_cache = self.browse(cr, uid, ids, context)[0]
-        where = """
+    @api.multi
+    def _comment_bought(self):
+        sale_report_obj = self.env['sale.report']
+        sale_select = sale_report_obj._select()
+        sale_from = sale_report_obj._from()
+        sale_group_by = sale_report_obj._group_by()
+        for mail in self:
+            where = """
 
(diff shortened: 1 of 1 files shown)


hbto:
@hugho-ad 

do not forget to use a for cycle 

For the ids and not call the ids[0]

hugho-ad:
@hbto got it

website_product_comment_purchased/models/mail.py
```diff
@@ -22,37 +22,38 @@
...
+        res = {}.fromkeys(ids, False)
         sale_report_obj = self.pool.get('sale.report')
-        product_obj = self.pool.get('product.product')
-        cr.execute("""
-            {0}
-            FROM ( {1} )
-            {2}
-            """.format(sale_report_obj._select(), sale_report_obj._from(),
-                       sale_report_obj._group_by()))
         mail_cache = self.browse(cr, uid, ids, context)[0]
```
moylop260: You can avoid get a browse of all ids if you will get just the first item with:
`mail_cache = self.browse(cr, uid, ids[0], context)`

website_product_comment_purchased/models/mail.py
```diff
@@ -22,37 +22,38 @@
...
+            FROM ( {_from} )
+            {where}
+            {group}
+            """.format(
+            select=sale_report_obj._select(),
+            _from=sale_report_obj._from(),
+            where=where,
+            group=sale_report_obj._group_by())
+        cr.execute(execute)
+        if cr.fetchall():
```
moylop260: If you need a boolean then you could avoid get all rows.
You can use `LIMIT 1` in sql and a `cr.fetchone()` in py.

moylop260:
@hugho-ad 
I added small comments.

Could you share us the result of performance before of this PR and after of this pr, please?

moylop260:
:+1:

moylop260:
@nhomar 
Are you agree with this change?

moylop260:
fix
@hugho-ad 
This pr fix 167 too?
Joke!

moylop260:
I apply this fix to last build of legacy and work very good!
If exists a new comment we can create a new pr.

PR #733, opened 2016-02-17
[REM] stock_picking_cancel: Remove module until be needed.

KarenKawaii:
- [x] Remove module cause the following:
  - References to depreciated models like `stock.picking.in` and `stock.picking.out`.
  - Fixing previous error, after replacing the depreciated models by the new one stock.picking, when it trys to load the workflow, the references no longer exists in v.8.0 for `stock.picking` model. You can check it in the next links: [stock_workflow.xml](https://github.com/odoo/odoo/blob/7.0/addons/stock/stock_workflow.xml) only available for v. <=7.0
  - This module doesn't belong to a client's project.
  - The last significant commit was 3 years ago, duplicating the function `action_cancel_draft` in `stock.picking.in` and `stock.picking.out`:
    addons vauxoo stock_picking_cancel stock py at 8 0 vauxoo addons vauxoo

If this module is removed, the `mrp_production_cancel` has to being removed too.


nhomar:
@KarenKawaii rebase please

KarenKawaii:
@nhomar this branch is already rebased.

PR #734, opened 2016-02-17
[FIX] website_product_comment_purchased: migrate feature to old api, to avoid no enter to the function

hugho-ad:
https://github.com/Vauxoo/yoytec/pull/738

This change is because the field comment_bought is not being calculated when using the new api, using the old api, the field is being calculated, see the next video to show you an easy example.

https://youtu.be/-RcrjbLturE

Also, if we write the record ,decorating the field previsuly with api.depends, the field is not calculated.

### DIFF
--- a/website_product_comment_purchased/models/mail.py
+++ b/website_product_comment_purchased/models/mail.py
@@ -42,7 +42,7 @@ class MailMessage(osv.Model):
                         t.id={res_id} AND s.partner_id={partner_id}
 
                 """.format(res_id=mail.res_id,
-                           partner_id=mail.author_id.id)
+                           partner_id=mail.author_id.id or 'null')
             execute = """
                 {select}
                 FROM ( {_from} )
--- a/website_product_comment_purchased/models/mail.py
+++ b/website_product_comment_purchased/models/mail.py
@@ -22,19 +22,20 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp import models, fields, api
+from openerp.osv import osv, fields
 
 
-class MailMessage(models.Model):
+class MailMessage(osv.Model):
     _inherit = 'mail.message'
 
-    @api.multi
-    def _comment_bought(self):
-        sale_report_obj = self.env['sale.report']
+    def _comment_bought(self, cr, uid, ids, name, arg, context=None):
+        res = {}
+        sale_report_obj = self.pool.get('sale.report')
         sale_select = sale_report_obj._select()
         sale_from = sale_report_obj._from()
         sale_group_by = sale_report_obj._group_by()
-        for mail in self:
+        for mail_id in ids:
+            mail = self.browse(cr, uid, mail_id, context)
             where = """
 
             WHERE l.product_id IS NOT NULL AND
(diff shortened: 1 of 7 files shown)


hugho-ad:
@moylop260 here is the pr

website_product_comment_purchased/models/mail.py
```diff
@@ -22,41 +22,41 @@
...
-        sale_report_obj = self.env['sale.report']
-        sale_select = sale_report_obj._select()
-        sale_from = sale_report_obj._from()
-        sale_group_by = sale_report_obj._group_by()
-        for mail in self:
-            where = """
+    def _comment_bought(self, cr, uid, ids, name, arg, context=None):
+        res = {}.fromkeys(ids, False)
+        sale_report_obj = self.pool.get('sale.report')
+        mail_cache = self.browse(cr, uid, ids, context)[0]
```
moylop260: Use a for to set value in all ids and get the boolean by each id.
Similar to new api script

website_product_comment_purchased/models/mail.py
```diff
@@ -22,41 +22,41 @@
...
-    @api.multi
-    def _comment_bought(self):
-        sale_report_obj = self.env['sale.report']
-        sale_select = sale_report_obj._select()
-        sale_from = sale_report_obj._from()
-        sale_group_by = sale_report_obj._group_by()
-        for mail in self:
-            where = """
+    def _comment_bought(self, cr, uid, ids, name, arg, context=None):
+        res = {}.fromkeys(ids, False)
```
moylop260: Avoid this sentence with the for to asign false or true

website_product_comment_purchased/models/mail.py
```diff
@@ -22,41 +22,41 @@
...
+            """.format(res_id=mail_cache.res_id,
+                       partner_id=mail_cache.author_id.id)
+        execute = """
+            {select}
+            FROM ( {_from} )
+            {where}
+            {group}
+            LIMIT 1
+            """.format(
+            select=sale_report_obj._select(),
```
moylop260: Set this value in top of the method to set a for without recalculation

nhomar:
@hugho-ad I honestly did not understand here the feature¿?

What you need it is the @depends no migrate to old Api.

nhomar:
nad please @hugho-ad add a unit test that's the way you will avoid regressions, I do not understand this change.

website_product_comment_purchased/models/mail.py
hugho-ad: I gues set that values at the top of method must to avoid recalculation in a for loop, but see the next:
image

moylop260: I dont see a parameter received in that methods.
Then you dont need recalc here.
If even return the same value in each iter then you can use it out of the for

hugho-ad: got it

hugho-ad:
@nhomar 

This change is because @moylop260  and me, detected that the field comment_bought is not being calculated when using the new api, using the old api, the field is being calculated, see the next video to show you an easy example.

https://youtu.be/-RcrjbLturE

Also, if we write the record ,decorating the field previsuly with api.depends, the field is not calculated.

hugho-ad:
@nhomar @moylop260 
Test added to validate the feauture of "Customer bought the item"

Regards

hugho-ad:
@nhomar can you merge this please

PR #738, opened 2016-02-18
[WIP][8.0] Enhance coverage and use cases for stock modules

osvalr:
This is intended to get enhanced unit testing for stock modules including new cases of use and coverage as well
### Use cases
- [x] Supplier return movement following (to be discussed)
- [x] Revert Landed Cost document and its effects over in Stock Card
  - [x] Check values before applying landed cost vs. after reverting landed cost
  - [x] Check journal entries after reverting landed cost 
- [x] When product price is set manually before apply Landed Cost
  - [x] A Loss journal entry is registered because price set is higher than the current one.
  - [x] A Gain journal entry is registered because price set is lesser than the current one.
- [X] Emulate trx flow where inventory valuation doesn't match with the product stock. [ more info ](https://www.odoo.com/documentation/user/9.0/accounting/others/inventory/avg_price_valuation.html#explanation-counter-example) 
### Coverage
- [x] Cron methods for **`product_extended_segmentation`** where material cost product value it gets updated where it have not value in any segmentation
### At the end
- [x] Rebase

### DIFF
--- a/product_extended_segmentation/model/template.py
+++ b/product_extended_segmentation/model/template.py
@@ -295,7 +295,7 @@ class ProductTemplate(models.Model):
 
                 # Call compute_price on these subproducts
                 prod_set = set([x.product_id.id for x in bom.bom_line_ids])
-                res = self.compute_price(
+                self.compute_price(
                     cr, uid, product_ids=list(prod_set), template_ids=[],
                     recursive=recursive, test=test,
                     real_time_accounting=real_time, context=context)
--- a/product_extended_segmentation/model/template.py
+++ b/product_extended_segmentation/model/template.py
@@ -299,9 +299,6 @@ class ProductTemplate(models.Model):
                     cr, uid, product_ids=list(prod_set), template_ids=[],
                     recursive=recursive, test=test,
                     real_time_accounting=real_time, context=context)
-                # /!\ NOTE: This is not logical
-                if test:
-                    testdict.update(res)
 
             # Use calc price to calculate and put the price on the product
             # of the BoM if necessary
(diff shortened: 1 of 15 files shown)


osvalr:
As it [stock_landed_costs.py](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/stock_landed_costs_average/model/stock_landed_costs.py#L481) shows in `stock_landed_costs_average` module, there is a validation like the following

``` python
if prod_qty[prod_id] and fst_avg != ini_avg and lst_avg != ini_avg:
```

My question is: 
- `ini_avg` is the current product price, in other words, the resulting average from last transaction, this is the same when comparing with `fst_avg`, when the condition mentioned is reached so: `fst_avg != ini_avg` is false. What use case supports this?

osvalr:
Files not listed belongs to other modules OR they are at 100% test covered

``` bash
Name                                                                                                     Stmts   Miss  Cover   Missing
--------------------------------------------------------------------------------------------------------------------------------------
/root/addons-vauxoo/product_extended_segmentation/model/template.py                           164     10    94%   117-119, 126, 135, 155, 211, 236-242
/root/addons-vauxoo/stock_card/model/stock_card.py                                            223      9    96%   30, 46, 83-87, 93, 132-137
/root/addons-vauxoo/stock_card_segmentation/model/stock_card.py                               185     18    90%   78-87, 95, 114-131
/root/addons-vauxoo/stock_landed_costs_average/model/stock_landed_costs.py                    322     14    96%   42, 47, 51, 53, 55, 250, 279, 393, 509, 547-548, 563, 637, 650
/root/addons-vauxoo/stock_landed_costs_segmentation/model/stock_landed_costs.py                35      2    94%   50, 66
/root/addons-vauxoo/stock_quant_cost_segmentation/model/stock.py                              156      5    97%   254, 266, 271, 279, 303
--------------------------------------------------------------------------------------------------------------------------------------
```

... more info at [coveralls](https://coveralls.io/jobs/12324442)

hbto:
> My question is: 
> - `ini_avg` is the current product price, in other words, the resulting average from last transaction, this is the same when comparing with `fst_avg`, when the condition mentioned is reached so: `fst_avg != ini_avg` is false. What use case supports this?

This is because sometimes User could change price of the product without any purchase involve.

PR #748, opened 2016-02-22
[IMP] Segmentation added on Workcenters

hbto:


hbto:
Everything went fine with these Changes

PR #752, opened 2016-02-26
[FIX]set a specific product selection, to avoid select a different product

hugho-ad:
FIX https://github.com/Vauxoo/yoytec/issues/784

Dummy https://github.com/Vauxoo/yoytec/pull/786


hugho-ad:
@nhomar can you merge this

nhomar:
wired in spanish?

That's wrong, here you should use the generic slector not the translated one (this PR is in Addons-Vaxuoo and broken element is in customer instance)

hugho-ad:
you right, I'll close this pr, the fix must be in customer instace

PR #755, opened 2016-02-29
[FIX] stock_card_segmentation: product price based on quants

osvalr:
## Changes made
- [X] Set product price based on quant movements when computing a supplier product return transaction
- [x] Rebase


osvalr:
@hbto waiting to this gets green, dummy PR is green already

PR #757, opened 2016-02-29
[IMP] Added on_change trigger on Landed Cost Document 

hbto:
to avoid Operators forgetting to push Update Cost from invoices


hbto:
Finally this is green

PR #758, opened 2016-03-01
8.0 product extended segmentation workcenter hbto

hbto:
As workcenter has got new feature based on segmentation 
product_extended_segmentation has to be sensitive of that field on workcenter

### DIFF
--- a/product_extended_segmentation/model/template.py
+++ b/product_extended_segmentation/model/template.py
@@ -183,10 +183,10 @@ class ProductTemplate(models.Model):
                     cr, uid, bom.product_uom.id, routing_price,
                     bom.product_id.uom_id.id)
                 price += routing_price
-                try:
-                    fn = wc.segmentation_cost or 'production_cost'
-                except AttributeError:
-                    fn = 'production_cost'
+                # /!\ NOTE: If not segmentation set on WC fallback to
+                # production_cost segmentation
+                fn = wc.segmentation_cost or 'production_cost'
+                fn = 'production_cost'
                 sgmnt_dict[fn] += routing_price
 
         # Convert on product UoM quantities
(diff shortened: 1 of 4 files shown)


product_extended_segmentation/model/template.py
```diff
@@ -183,7 +183,11 @@ def _factor(factor, product_efficiency, product_rounding):
                     cr, uid, bom.product_uom.id, routing_price,
                     bom.product_id.uom_id.id)
                 price += routing_price
-                sgmnt_dict['production_cost'] += routing_price
+                try:
+                    fn = wc.segmentation_cost or 'production_cost'
+                except AttributeError:
+                    fn = 'production_cost'
+                sgmnt_dict[fn] += routing_price
```
nhomar, line 190: I do not understand this approach in terms of sensitivity! :-8

hbto, line 190: This module is not yet dependent on mrp_workcenter_segmentation
in the meantime while testing this peace of code avoid any depend on that module
when becomes dependent on that module this try-except will get rid.
or when another approach arises

product_extended_segmentation/model/template.py
```diff
@@ -183,7 +183,11 @@ def _factor(factor, product_efficiency, product_rounding):
                     cr, uid, bom.product_uom.id, routing_price,
                     bom.product_id.uom_id.id)
                 price += routing_price
-                sgmnt_dict['production_cost'] += routing_price
+                # /!\ NOTE: If not segmentation set on WC fallback to
+                # production_cost segmentation
+                fn = wc.segmentation_cost or 'production_cost'
+                fn = 'production_cost'
+                sgmnt_dict[fn] += routing_price
```
hbto, line 190: @nhomar 

try-except clause was taken away

PR #760, opened 2016-03-01
	[ADD][sale_block_cancellation] Module to block cancellation of sales with validation added.

luistorresm:
Added a new button in Sale Orders to only allow cancel a SO when all the
pickings related with this are on `ready to transfer` or in
`waiting for another move` state, or even with pickings on different states
if the moved product quantities are correctly returned.

The original button to cancel is added to a new group `Force Sale Cancel`
to can allow cancel a Sale Order out of these cases.
- [x] Added module to block cancellation of sale orders.
- [x] Added tests


sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,53 @@
...
+        in "Waiting another move" state, the cancellation of this must be
+        allowed, or even whit pickings on different states, if the moved
+        product quantities are correctly returned.'''
+        allow = True
+        for sale in self:
+            dict_prod = {}
+            for pick in sale.picking_ids:
+                if pick.state == 'done':
+                    for move in pick.move_lines:
+                        prod_id = move.product_id.id
```
nhomar: Por favor por este algoritmo (el de las ubicaciones.) en un micro metodo externo para bajar la complejidad ciclomática.

nhomar: desde la 28 hasta la 42 puede ir en un metodo externo.

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,53 @@
...
+    @api.multi
+    def action_cancel_if(self):
+        '''If all the pickings related to a SO are on "Ready to transfer" or
+        in "Waiting another move" state, the cancellation of this must be
+        allowed, or even whit pickings on different states, if the moved
+        product quantities are correctly returned.'''
+        allow = True
+        for sale in self:
+            dict_prod = {}
+            for pick in sale.picking_ids:
```
nhomar: Una vez tengas un método externo usa una list comprehension aqui.

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,53 @@
...
+                for loc in dict_prod.get(prod):
+                    if dict_prod.get(prod).get(loc) != 0:
+                        allow = False
+            if not allow:
+                raise ValidationError(_(
+                    "To allow cancel this sale order:\n- All the pickings "
+                    "are in 'Ready to transfer' or 'Waiting another move'."
+                    "\n- If the moved product quantities in pickings are "
+                    "correctly returned."))
+        return self.action_cancel()
```
nhomar, line 112: me gusta más si sobreescribes y preprocesas el cancel original ¿Por qué un botón por separado aquí y no un super preprocesado.?

nhomar:
No veo test unitarios.

sale_block_cancellation/security/res_groups.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="0">
+
+        <record id="group_cancel_sale_blocked" model="res.groups">
+            <field name="name">Cancel Sales Blocked</field>
```
sabrinagnu: @luistorresm cambiar "Cancel Sales Blocked" >> Show default sale order cancel button

sabrinagnu: @luistorresm y en español "Mostrar botón cancelar por defecto en SO"

luistorresm:
Done @dsabrinarg

luistorresm:
@JulioSerna 

May you give me +1?

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,108 @@
...
+        'this SO are on "Ready to transfer", "Waiting Availability" or in '
+        '"Waiting another move" or if the pickings on different states have '
+        'the move of product quantities correctly returned. If this '
+        'validations are True the sale order cancellation must be allowed.')
+    picking_ids = fields.One2many(
+        compute=_get_picking_ids, method=True, relation='stock.picking',
+        string='Picking associated to this sale', search=_search_pickings)
+
+    @api.multi
+    def check_prod_by_locations(self):
```
JulioSerna, line 75: @luistorresm 

What happen is the warehouse is in three steps to delivery and the customer make a return and return is of customer to stock directly , this work as well?

luistorresm:
@JulioSerna 

The sale order allow cancel, because all the moves are correctly returned in that case.

sale_block_cancellation/tests/test_sale_block_cancellation.py
```diff
@@ -0,0 +1,83 @@
...
+        'Test that allow cancel Sale Order, with pickings returned'
+        sale = self.sale_demo.copy({'order_policy': 'picking'})
+        sale.warehouse_id.write({
+            'delivery_steps': 'pick_pack_ship',
+        })
+        sale.action_button_confirm()
+        self.proc_obj.create({}).procure_calculation()
+        for picking in sale.picking_ids:
+            if picking.state == 'confirmed':
+                picking.force_assign()
```
JulioSerna: @luistorresm 

You must avoid to force assing,

https://github.com/Vauxoo/lodigroup/pull/451

sabrinagnu:
@luistorresm "The sale order allow cancel, because all the moves are correctly returned in that case.", Which test you did to affirm this?, I believe this is not true and the system is not allowing the cancellation on the case that Julio pointed.

luistorresm:
@JulioSerna, 

@dsabrinarg have reason, if I try cancel a SO when this are Done, I get the next message:
`You cannot cancel a sale order line which is linked to a stock move already done.`

sale_block_cancellation/tests/test_sale_block_cancellation.py
luistorresm: Done @JulioSerna

JulioSerna:
Pending to review @dsabrinarg  with @nhomar

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,108 @@
...
+                    qty_d = dict_prod.get(prod_id).get(loc_d, 0.0)
+                    dict_prod.get(prod_id).update({
+                        loc_s: qty_s - move.product_uom_qty,
+                        loc_d: qty_d + move.product_uom_qty,
+                    })
+        for prod in dict_prod:
+            for loc in dict_prod.get(prod):
+                if dict_prod.get(prod).get(loc) != 0:
+                    return False
+        return True
```
josemoralesp, line 101: Hello @luistorresm 

Please can you make a refactor for this method.

Please Consider the use of sql, because if we have orders with 300 lines(like a normal day in apex) we will have more than 900 lines considering the picking returned.

This method is called when we install this module because the compute field is store=True.

This module may take many hours installing for this reason.

Please let me know any doubt with this message

Best Regards.

luistorresm, line 101: Can check again @josemoralesp .

I was change to query.

josemoralesp, line 101: Hello @luistorresm 

The main reason for the sql is avoid the excessive use of sentence 'for' in your method.

I was Checking your method and you are using the move to know if the products were correctly returned but with this you need to do something else(a lot of 'for' sentences).

I want to suggest you the use of the quant instead the move, because if the location of the quants used for your moves is different to the origin of the move(It may be stock) you can say that you have products without return.

E.G.

``` python
self._cr.execute('''
SELECT product_id, 
       sum(qty),
       location_id 
FROM stock_quant 
WHERE id IN (SELECT quant_id 
             FROM stock_quant_move_rel 
             WHERE move_id IN (SELECT id 
                               FROM stock_move 
                               WHERE picking_id IN %s)) AND
       location_id != %s 
GROUP BY product_id, location_id
''', (tuple(pick_ids.ids), main_loc_id))
```

Where main_loc_id is the location where products initiated the moves.

If this sql sentences returns something this means that you have products without returns, no matter if they return the moves directly from customer to stock.

Do you understand my point?

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @api.multi
+    @api.depends('picking_ids', 'picking_ids.state', 'state')
+    def _get_allow_cancel(self):
+        '''If all the pickings related to a SO are on "Ready to transfer" or
+        in "Waiting another move" state, the cancellation of this must be
+        allowed, or even whit pickings on different states, if the moved
+        product quantities are correctly returned.'''
+        for sale in self:
+            sale.allow_cancel = False
```
josemoralesp: By default the boolean fields are False, you can exclude this to avoid call the write method twice for each order

luistorresm: Done @josemoralesp

sale_block_cancellation/models/sale.py
```diff
@@ -0,0 +1,109 @@
...
+        WHERE picking_id IN (SELECT id FROM stock_picking WHERE id in %s) and
+            state='done' and location_dest_id in (%s, %s)
+        GROUP BY product_id;
+            """, (loc_src, loc_dest, tuple(self.picking_ids.ids), loc_src, loc_dest)
+        )
+        prods = self.env.cr.fetchall()
+        for prod in prods:
+            if sum(prod[1]) != sum(prod[2]):
+                return False
+        return True
```
josemoralesp, line 101: Perfect

luistorresm, line 101: @josemoralesp

Can you give me an +1?

Regarda

josemoralesp:
:+1:

PR #769, opened 2016-03-04
[FIX] website_variants_extra: When selecting a invalid combination of attributes the price dissapears, now the message "Product not available" appears instead.

gmlp:
fixes https://github.com/Vauxoo/yoytec/issues/816
dummy https://github.com/Vauxoo/yoytec/pull/817
When product does not exists product price disappear and then product not available product message is triggered.This happen when unavailable product variant (memory 32 gb and black color) are selected.
Before
screenshot from 2016-03-04 01-21-13
After
screenshot from 2016-03-08 21-17-13

### DIFF
--- a/website_product_availability/views/templates.xml
+++ b/website_product_availability/views/templates.xml
@@ -3,7 +3,8 @@
     <data>
          <template id="product_stock_available_website" inherit_id="website_variants_extra.product_variants_description_3columns">
              <xpath expr="//div[@class='css_quantity input-group oe_website_spinner']" position="before">
-                <t t-set="attr_value_ids" t-value="get_locations_variant_ids(product)"/>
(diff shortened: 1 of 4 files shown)


nhomar:
@gmlp before after image please! when things are visual it is a little more complicated understand the change just reading the code.

gmlp:
@nhomar , description updated.

oscarolar:
:+1: This is good to me.

gmlp:
@nhomar, Build done.

nhomar:
@gmlp 

in order to be consistent the right column should hide all buttons and links if not it looks like a bug.

I mean all icons and warehouse links should disappear.

Can you finish it on that way please?

gmlp:
@nhomar , Done.

PR #770, opened 2016-03-04
[FIX] website_product_filters: if public category is not send by get parameters, this value is getting  from product model.

gmlp:
fixes https://github.com/Vauxoo/yoytec/issues/799
Dummy https://github.com/Vauxoo/yoytec/pull/818
The breadcrumb must be always showed . Before this change the breadcrumb appears if only the user  follow the categories links flow not when open directly the flow.
Before 
screenshot from 2016-03-03 23-20-56
After
screenshot from 2016-03-03 23-10-22

### DIFF
--- a/website_product_filters/controllers/main.py
+++ b/website_product_filters/controllers/main.py
@@ -183,6 +183,8 @@ class WebsiteSale(website_sale):
         cr, uid, context, pool =\
             request.cr, request.uid, request.context, request.registry
         template_obj = pool['product.template']
+        if not category and len(product.public_categ_ids) >= 1:
+            category = product.public_categ_ids[0]
         viewed = product.views + 1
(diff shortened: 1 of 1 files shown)


nhomar:
@gmlp can you please set a before after image, I can not ee how this change fix the related error.

@oscarolar can you extend a little the issue in order to understand the poblem.?

gmlp:
@nhomar, description updated

nhomar:
It is a little more clear for me, but please you should set also a list of steps to reproduce the failure (if applies) - (which sadly is not ins the issue) - and then explain how test the feature (witht the new approach).

@oscarolar do you approve the way this feature was done?

oscarolar:
Yes, its already merged but it works.

PR #772, opened 2016-03-05
[IMP] Server Action to update Journal Items in Manufacturing Orders

hbto:
There are Manufacturing Order already started which have already created Journal Entries
because of the stock move consumed or produced,

Those Journal Entries need to be properly associated the manufacturing order to properly
create Final Journal Entries


hbto:
Server Action is working as expected

PR #774, opened 2016-03-05
8.0 mrp routing account journal hbto

hbto:
Missing Field on MRP when routing is not set


hbto:
Everything went fine with this PR on Runbot

PR #779, opened 2016-03-08
[FIX] website_rate_product: SQL sentence moved outside the loop to avoid long cumulative time on execution of calculation of rating

oscarolar:
FIXES https://github.com/Vauxoo/yoytec/issues/808
BUILD https://github.com/Vauxoo/yoytec/pull/838

### DIFF
--- a/website_rate_product/models/mail_message.py
+++ b/website_rate_product/models/mail_message.py
@@ -22,57 +22,39 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp.osv import osv, fields
+from openerp import models, fields, api
 
 
-class MailMessage(osv.Model):
+class MailMessage(models.Model):
     _inherit = 'mail.message'
-    _columns = {
-        'rating': fields.integer('Rating'),
-    }
+    rating = fields.Integer('Rating')
 
-    def _message_read_dict(self, cr, uid, message, parent_id=False,
-                           context=None):
-        res = super(MailMessage, self)._message_read_dict(cr, uid,
-                                                          message,
-                                                          parent_id=parent_id,
-                                                          context=context)
+    @api.model
+    def _message_read_dict(self, message, parent_id=False):
+        res = super(MailMessage, self)._message_read_dict(message,
+                                                          parent_id=parent_id)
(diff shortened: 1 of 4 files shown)


website_rate_product/models/mail_message.py
```diff
@@ -45,16 +45,19 @@ class ProductTemplate(osv.Model):
...
-                        and model = 'product.template' and rating > 0;" %
-                       (pid))
-            fall = cr.fetchall()
-            total = fall[0][0] if fall else total
-            res[pid] = total
+        str_tuple = str(tuple(ids)).replace(',)', ')')
+        cr.execute("""
+          select avg(rating), res_id from mail_message where res_id in %s\
+          and model = 'product.template' and rating > 0 group by res_id;
+          """ % (str_tuple))
```
moylop260: FYI [no-sql-injections-please](http://members.hellug.gr/xrg/openerp-doc/html/contribute/15_guidelines/coding_guidelines_framework.html#no-sql-injections-please)

website_rate_product/models/mail_message.py
```diff
@@ -45,16 +45,19 @@ class ProductTemplate(osv.Model):
...
+        for record in cr.fetchall():
+            total = record[0]
```
moylop260: I don't understand the focus of `total` variable.
Could you help me to understand it?

What about use:

``` python
for rating, product_template_id in cr.fetchall():
   res[product_template_id] = rating
```

website_rate_product/models/mail_message.py
```diff
@@ -45,16 +45,19 @@ class ProductTemplate(osv.Model):
     _inherit = 'product.template'
 
     def _get_rating(self, cr, uid, ids, field_name, arg, context):
+        """
+        This method gets the rating for each rated template based on the
+        comments, the rating per comment is stored in the mail.message model
+        """
         res = {}
```
moylop260: You will need initialize the dict with all ids to avoid a `KeyError`
Check follow [example](https://github.com/odoo/odoo/blob/51641d2634e48d07a3279acee405469e5d194e34/addons/product/product.py#L276)
`res = dict.fromkeys(ids, 0)`

PR #781, opened 2016-03-09
[FIX] add call to action_assign method to change the state of the picking before transfer the picking

zaoral:
This change is require for new update of odoo repository in the way that treat the pickings.


zaoral:
@josemoralesp

nhomar:
+1

PR #793, opened 2016-03-14
[REF] Add test case to mrp_workcenter_segmentation

osvalr:
# Changes:
- [X] Add test case to check journal entries after a production is made
- [x] Rebase

### DIFF
--- a/mrp_workcenter_segmentation/tests/test_workcenter_segmentation.py
+++ b/mrp_workcenter_segmentation/tests/test_workcenter_segmentation.py
@@ -6,6 +6,7 @@
 # planned by: hbto@vauxoo.com
 
 from openerp.tests.common import TransactionCase
+from collections import namedtuple
 
 
 class TestWorkcenterSegmentation(TransactionCase):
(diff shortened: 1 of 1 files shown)


mrp_workcenter_segmentation/tests/test_workcenter_segmentation.py
```diff
@@ -73,6 +73,14 @@ def create_wizard(self, mrp_production, qty=1):
         wz_brw.do_produce()
         return True
 
+    def get_account_values(self, account_name):
+        rec_ids = self.mrp_production_e.account_move_id.line_id.filtered(
+            lambda l: l.account_id.name == account_name)
+        res = []
+        for rec_id in rec_ids:
+            res.append((rec_id.debit, rec_id.credit))
```
hbto: I would be more functionally valuable to gather the values of the debit and credit into variables
than passing the whole list of values.

that is 

res = (account='ACCOUNT', debit=200, credit=100)
return res

and so on.

osvalr: oh I see, let me do that instead

PR #796, opened 2016-03-17
[FIX] website_product_filters: search criteria isn't sending when sort filter changes.

gmlp:
**search criteria isn't sending when sort filter changes.It is fixed sending search criteria using js.**
FIXES: https://github.com/Vauxoo/yoytec/issues/788
BUILD: https://github.com/Vauxoo/yoytec/pull/886

### DIFF
--- a/website_product_filters/static/src/js/send_extra_params.js
+++ b/website_product_filters/static/src/js/send_extra_params.js
@@ -1,9 +1,14 @@
 $(document).ready(function () {
 $('.oe_website_sale').each(function () {
     var oe_website_sale = this;
-
-    $('form.js_attributes input, form.js_attributes select, form.js_extra_parameters select',oe_website_sale).on('change', function () {
-        $(this).closest("form").submit();
+    $('form.js_attributes select').on('change', function () {
+        var $form =$(this).closest("form");
+        var search_query = $('.search-query').val();
+        var input = $("<input>")
+                     .attr("type", "hidden")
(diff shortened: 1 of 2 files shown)


website_product_filters/static/src/js/send_extra_params.js
```diff
@@ -1,9 +1,14 @@
...
-    $('form.js_attributes input, form.js_attributes select, form.js_extra_parameters select',oe_website_sale).on('change', function () {
-        $(this).closest("form").submit();
+    $('form.js_attributes select').on('change', function () {
+        var $form =$(this).closest("form");
+        var search_query = $('.search-query').val();
+        var input = $("<input>")
+                     .attr("type", "hidden")
+                     .attr("name", "search").val(search_query);
+        $form.append($(input));
+        $form.submit();
```
nhomar, line 11: Por que pruebas un submit "hiden"? entendí bien el selector?

gmlp, line 11: @nhomar, this add a hidden field with the search criteria into the form.

nhomar, line 11: ya sorry. +1

oscarolar:
AFAICS There was no need to add extra JS, you can simply add the `form="form_id"` to any other input outside the form that you want to send on the same HTTP request. 
see an example here
http://www.w3schools.com/html/tryit.asp?filename=tryhtml5_input_form

PR #800, opened 2016-03-19
[REF] website_rate_product: Use query with multiple ids

moylop260:
- Refactory to use a query with N ids instead of N queries with one id.
  - Continuation from https://github.com/Vauxoo/yoytec/issues/808#issuecomment-193885012

### DIFF
--- a/website_rate_product/models/mail_message.py
+++ b/website_rate_product/models/mail_message.py
@@ -46,15 +46,13 @@ class ProductTemplate(models.Model):
         This method gets the rating for each rated template based on the
         comments, the rating per comment is stored in the mail.message model
         """
-        for product in self:
-            self._cr.execute("""
-              select avg(rating), res_id from mail_message where res_id = %s\
-              and model = 'product.template' and rating > 0 group by res_id;
-              """, (product.id,))
-            record = self._cr.fetchall()
-            if record:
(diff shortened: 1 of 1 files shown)


moylop260:
I don't ran the cProfile tests and don't ran the pgbadger tests to check values of performance.
I can run if is required.

But by now I have created the PR to review.

FYI @josemoralesp @hugho-ad @oscarolar @nhomar

hugho-ad:
:+1: LGTM

oscarolar:
Just to know, what is going to happen when you add this field on a list view?

moylop260:
> Just to know, What is going to happen when you add this field on a list view?

After change:
work fine!

Before change:
work fine!

Both methods work fine!
It is just a small REF to use a query with N ids instead of N queries with one id.
Continuation from https://github.com/Vauxoo/yoytec/issues/808#issuecomment-193885012

oscarolar:
Good then.

moylop260:
Dummy PR: https://github.com/Vauxoo/yoytec/pull/907

josemoralesp:
IMHO 
The best way to do this it's using one query with multiple ids, because using the other way(one query by id) We will send several queries in a loop to the database and this is not a good practice if we talk about performance

PR #823, opened 2016-04-13
[FIX] Wrong manifest version format

luiseevaquer:
[FIX] Wrong manifest version format


moylop260:
@lescobarvx 
Para evitar conflicts traete el cherry-pick del fix en versión 8.0 con 8.0.0.x.x y luego lo aplicas el script en el branch 9.0 para cambiarlo a 9.0.0.x.x

luiseevaquer:
@moylop260 
Listo el cambio de version de 8.x.x.x.x a 9.x.x.x.x

moylop260:
The original sha was changed it :(

moylop260:
@lescobarvx 
What is the script and commans used here?

luiseevaquer:
@moylop260 
The script used is the following:
- You can fix wrong version format (x.y to 8.0.0.0.0):
  `
  $ oca-autopep8 -ri --select=CW0005 PATH
  `
- If you want to change manifest version default, you need to set MANIFEST_VERSION environtment variable, for example:
  `
  $ export MANIFEST_VERSION=9.0.0.0.0
  `

These indications are in the README.md of https://github.com/Vauxoo/maintainer-tools project

moylop260:
@nhomar 
In order to avoid a massive dummy commit... we should create a installable: False for 9.0 modules of addons-vauxoo, and enabling changing the version too... or should we change the version of all modules to 9.0 by default?

account_advance_payment/__openerp__.py
```diff
@@ -25,7 +25,7 @@
 #
 {
     "name": "Account Advance Payment",
-    "version": "1.6",
+    "version": "9.0.0.1.6",
```
osvalr, line 28: why `1.6` is taken as minor and patch version when using as full module version `9.0.0.1.6` instead of taken it as mayor and minor versions, I mean `9.0.1.6.0`?

moylop260, line 28: The true of `1.6` is that is a copy&paste of other module.
Don't represent a real version.
It just was added as reference.

luisg123v:
This Has conflicts and the branch 9.0 is not used anymore, so closing.

PR #875, opened 2016-04-21
[ADD] Adding fields sequence for model acceptability criteria

egrivero:
FIX vauxoo/instance#180
Close vauxoo/instance#265

### DIFF
--- a/user_story/model/user_story.py
+++ b/user_story/model/user_story.py
@@ -633,14 +633,13 @@ class AcceptabilityCriteria(osv.Model):
             returns the state of user.story to which belong '''
         res = {}.fromkeys(ids)
         for ac in self.browse(cr, uid, ids, context=context):
+            ac_number = ac.name.split(')')
+            if len(ac_number) > 1:
+                ac_number = ac_number[0]
+            else:
+                ac_number = ac.name[0:3]
             if ac.sequence_ac:
                 ac_number = str(ac.sequence_ac)
-            else:
-                ac_number = ac.name.split(')')
-                if len(ac_number) > 1:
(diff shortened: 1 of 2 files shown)


user_story/model/user_story.py
```diff
@@ -633,15 +633,35 @@ def _get_us_ca_numbers(self, cr, uid, ids, fieldname, arg,
...
+        elif len(ac_ids):
+            order = [x[1] for x in ac_ids]
+            read_ac = self.read(cr, uid, order,
+                                ['sequence_ac'], context=context)
+            maxi = max(read_ac, key=lambda x: x['sequence_ac'])
+            seq = maxi['sequence_ac'] + 1
+        else:
+            seq = 1
+        return seq
+
```
nhomar, line 662: this method should be:
variable seq and acc_number are defined inside IF's this is a bad practice, you need initialize always the default value before go to if.

Plase check that.

thanks.

egrivero, line 662: the variables are defined in both the if and the else, even so must be initialized ?. I understand when used alone in a condition if

nhomar, line 662: Yes.

The correct approach is:

``` python
var=1
if something:
   var=logic()
if other:
   var=logic2()
```

Use:

``` python
if something:
    var=logic()
elif other:
    var=logic2()
else:
    var=1
```

Is not considered clean.

Why?

The default behavior should be shown first and then the 1st case possible and then the second it is more readable than left the default value at the end in the else (yes I know python shit.!)

egrivero:
Travis generates errors but are not related to the changes I make

PR #890, opened 2016-04-28
Revert "[IMP] user_story: Adding task in the user story will add it at the top of the list."

zaoral:
vx#5453

---

Fix Vauxoo/instance#273
Close Vauxoo/instance#275
Reverts Vauxoo/addons-vauxoo#594

The change made generate a functional error using user story: make the task list editable in the embedded tree view and we do not want this:

We want to get inside the task to edit it all.
Sorry for the trouble

Best Regards


oscarolar:
Tested in updates.vauxoo.com please deploy to production 👍

PR #917, opened 2016-05-05
[MERGE] Merge addons-vx/8.0 into addons-vx/9.0

moylop260:


moylop260:
@nhomar 
Could you merge it?

PR #924, opened 2016-05-09
8.0 mrp routing account journal hbto

hbto:


hbto:
@nhomar @dsabrinarg 

Your attention is required

nhomar:
I Hope a migration folder if at least record has journal_id it should be moved to the new destiny.

mrp_workcenter_account_move/model/mrp.py
```diff
@@ -38,10 +38,11 @@ def test_accounting_setting(self):
...
-            msg_journal = _('Please set a Journal in BoM: {bom} to book '
-                            'Production Cost Journal Entries\n')
-            if not self.bom_id.journal_id:
-                msg += msg_journal.format(bom=self.bom_id.name)
+            msg_journal = _(
+                'Please set a Journal in Product Category: {cat} to book '
+                'Production Cost Journal Entries\n')
+            if not self.product_id.categ_id.property_stock_journal:
+                msg += msg_journal.format(cat=self.product_id.categ_id.name)
             if msg:
```
hbto, line 46: @nhomar 

There is no need for any of that,

We are just getting rid of a field and instead we are using a previous existing one.

provision exist to avoid people from booking Journal Entries without that field.

nhomar, line 46: Ok.
It is more clear now.

We have an ACL Problem:

On this line if the user does not have ACL to read at least the model Journal:

```
if not self.product_id.categ_id.property_stock_journal.
```

This will give an ACL problem and not the controlled message you are waiting for (which was the one that brings the problem alive).

Can you validate please, make this check with sudo?

hbto, line 46: The problem was an issue of view where only Financial Manager was allowed to check that feature.

That same problem was not present in mrp.routing.

mrp_routing_account_journal/view/view.xml
```diff
@@ -11,18 +11,5 @@
...
             </field>
         </record>
-        <record id="mrp_bom_journal_form_view" model="ir.ui.view">
-            <field name="name">mrp.bom.journal.form.view</field>
-            <field name="inherit_id" ref="mrp.mrp_bom_form_view"/>
-            <field name="model">mrp.bom</field>
-            <field name="arch" type="xml">
-                <xpath expr="//label[@for='routing_id']" position="before">
-                    <label for="journal_id" class="oe_inline" groups="account.group_account_manager"/>
-                    <div groups="account.group_account_manager">
```
hbto, line 21: @nhomar here lies the problem of your comments on ACL.

there were no problems of ACL but problems for ACL on Views not properly addressed

Issue #932, opened 2016-05-14
Migration accounting modules from project vauxoo/addons-vauxoo to vauxoo/account 


hbto:
Moving modules from https://github.com/Vauxoo/addons-vauxoo to https://github.com/Vauxoo/account
Related to https://github.com/Vauxoo/account/issues/3
## List of modules to migrate

| Module | TODO | WIP | DONE | BLOCKED BY |
| --- | --- | --- | --- | --- |
| account_advance_payment | x |  |  |  |
| account_aged_partner_balance_vw | x |  |  |  |
| account_amortization | x |  |  |  |
| account_analytic_active_field | x |  |  |  |
| account_analytic_btree | x |  |  |  |
| account_analytic_required | x |  |  |  |
| account_anglo_saxon_missing_key | x |  |  |  |
| account_anglo_saxon_stock_move | x |  |  |  |
| account_anglo_saxon_stock_move_purchase | x |  |  |  |
| account_anglo_saxon_stock_move_sale | x |  |  |  |
| account_asset_analytic | x |  |  |  |
| account_asset_date | x |  |  |  |
| account_asset_move_check | x |  |  |  |
| account_bank_statement_vauxoo | x |  |  |  |
| account_budget_imp | x |  |  |  |
| account_closure_preparation | x |  |  |  |
| account_currency_tools | x |  |  |  |
| account_entries_report_group_by_ref | x |  |  |  |
| account_group_auditory | x |  |  |  |
| account_group_auditory_assets | x |  |  |  |
| account_invoice_line_asset_category_required | x |  |  |  |
| account_invoice_line_currency | x |  |  |  |
| account_invoice_number | x |  |  |  |
| account_invoice_regular_validation | x |  |  |  |
| account_invoice_show_by_user | x |  |  |  |
| account_invoice_supplier_quantity_attachments | x |  |  |  |
| account_invoice_tax | x |  |  |  |
| account_ledger_report | x |  |  |  |
| account_model_plans | x |  |  |  |
| account_move_cancel | x |  |  |  |
| account_move_filters | x |  |  |  |
| account_move_folio | x |  |  |  |
| account_move_group | x |  |  |  |
| account_move_line_base_tax | x |  |  |  |
| account_move_line_extended | x |  |  |  |
| account_move_line_group_analytic | x |  |  |  |
| account_move_line_group_by_asset | x |  |  |  |
| account_move_line_group_by_extend | x |  |  |  |
[...]

PR #957, opened 2016-05-21
Migrating Accounting Modules from vauxoo/addons-vauxoo to vauxoo/account 

hbto:
This PR Deletes 81 modules and translate them to new
Project at 
https://github.com/Vauxoo/account

Related to Issue https://github.com/Vauxoo/addons-vauxoo/issues/932


luisg123v:
This seems abandoned, so closing.

PR #967, opened 2016-05-25
[IMP][account_voucher_tax] Added action server to validate when from

JulioSerna:
bank statement reconcile payment in secondary currency sometimes
create Journal Item by Rounding adjustment to high

The Behavior cannot reproduce and is related the vauxoo issue#1460

PR DUMMY https://github.com/Vauxoo/lodigroup/pull/563

### DIFF
--- a/account_voucher_tax/__openerp__.py
+++ b/account_voucher_tax/__openerp__.py
@@ -45,7 +45,8 @@
     "data": [
         "account_tax_view.xml",
         "account_voucher_tax_view.xml",
-        "security/ir.model.access.csv"
+        "security/ir.model.access.csv",
+        "data/ir_actions_server.xml"
     ],
     "test": [
         "test/account_voucher_taxes.yml",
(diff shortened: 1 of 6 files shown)


nhomar:
@JulioSerna 

Can you put this logic into a method and just call teh method and add a proper unittest to the method.. please.!

account_voucher_tax/__openerp__.py
```diff
@@ -45,7 +45,8 @@
     "data": [
         "account_tax_view.xml",
         "account_voucher_tax_view.xml",
-        "security/ir.model.access.csv"
+        "security/ir.model.access.csv",
+        "data/ir_actions_server.xml"
```
hbto: Please add the trailing comma at the end of the new 
element being added,
next time new change will involve only one line.

JulioSerna:
@nhomar @hbto 

Changes applied

PR #970, opened 2016-05-27
[ADD] Controller to be able to print the brochure of a product with public URL HU#1006 CA#47

oscarolar:
Build https://github.com/Vauxoo/yoytec/pull/1305

### DIFF
--- a/website_variants_extra/controllers/main.py
+++ b/website_variants_extra/controllers/main.py
@@ -1,0 +1,41 @@
+# coding: utf-8
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2014-Today OpenERP SA (<http://www.openerp.com>).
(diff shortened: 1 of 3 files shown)


website_variants_extra/controllers/main.py
```diff
@@ -0,0 +1,53 @@
...
+        if docids:
+            docids = [int(i) for i in docids.split(',')]
+        if data.get('context'):
+            # Ignore 'lang' here, because the context in data is the one
+            # from the webclient *but* if the user explicitely wants to
+            # change the lang, this mechanism overwrites it.
+            data_context = simplejson.loads(data['context'])
+            if data_context.get('lang'):
+                del data_context['lang']
+            context.update(data_context)
```
nhomar: I did the lang should not be from the user at least?

PR #975, opened 2016-06-01
[IMP][mrp_partial_production] Add depends to recompute function

JulioSerna:
when is called from tests and avoid issue vauxoo/lodigroup#569


JulioSerna:
@moylop260 

Could you merge this?  it is about to fix the error by vauxoo/odoo#99

PR Dummy,
https://github.com/Vauxoo/lodigroup/pull/565

Only error with the PAC,
job 1843 1 vauxoo lodigroup travis ci

PR #981, opened 2016-06-08
[FIX] Fixing date intervals in computing range values for due days

hbto:
Fixing date intervals in computing range values for due days


hbto:
This change has gone green on functional test and will be merged to be applied on customers instances

PR #989, opened 2016-06-11
[IMP] allow sale order copy line with orders in sent state

cvinh:
Hello
I allow copying lines with order in sent state
Best regards


luisg123v:

PR #990, opened 2016-06-11
[FIX] Quick and Dirty fix for account_model_plans

cvinh:
I did not manage to overwrite original generate method... so I totally rewrited it... dirty isn't it ?


hbto:
Thanks for the PR @invitu 

I would be better if that dirty Improve you have proposed could be 
cleaned and modularised, those improving MacCabe index rating,
Though to be legit I haven't test your code, nor functionally nor technically.

Regards.

PR #998, opened 2016-06-15
[FIX] website_product_filters: Conceptual change.

nhomar:
1.  [ ] s/template/product/g
2.  [ ] list_price (in db) by lst_price (computed with price list).
3.  [ ] Redesign we ended to manage sort kind of Caché for prices and counting because it is consuming a lot of resources look page by page this values.
4.  [ ] Only unit tests it is not necessary test all the sale process for this little feature what we need is simple set a bunch of demo data look for an specific set of url and see if the filters returns what we expected in number of elements and computed values in the frontend.

### DIFF
--- a/website_product_filters/models/product.py
+++ b/website_product_filters/models/product.py
@@ -60,11 +60,36 @@ class WebsiteProductMetadata(models.Model):
         "product_public_category_id")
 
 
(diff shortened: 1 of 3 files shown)


luisg123v:

PR #1001, opened 2016-06-17
[DEL] migrated to cliente app

hugho-ad:
- website_order_confirm
  Deleted, this module only overwrite the function payment_validate in
  order to enable direct creation of Sale Order directly from website
  instead of creating a Quotation.


nhomar:
- Nobody depends of it?
- Can you extend a little here what this was doing in order to have it as comment please.

hugho-ad:
Talking with @oscarolar , this module is Not necessary to Delete.

nhomar:
Can you enlight me Why? what were the conclusions of that meeting please?

hugho-ad:
The module works, and is not necessary  remove them, just delete the dependency

PR #1002, opened 2016-06-17
[IMP] Adding method to convert numbers which were formated with

hbto:
thousands separators into float

### DIFF
--- a/controller_report_xls/controllers/main.py
+++ b/controller_report_xls/controllers/main.py
@@ -70,6 +70,8 @@ def is_formatted_number(value, thousands_sep, decimal_point):
 
     if value.count(thousands_sep) > 0:
         value = value.replace(thousands_sep, '')
+    if decimal_point in value:
+        value = value.replace(decimal_point, '.')
     if is_integer(value):
         return True
     elif is_float(value):
@@ -93,6 +95,8 @@ def unformat_number(value, lang_sep):
 
     if value.count(thousands_sep) > 0:
         value = value.replace(thousands_sep, '')
+    if decimal_point in value:
+        value = value.replace(decimal_point, '.')
 
     if is_integer(value):
         value = int(value)
--- a/controller_report_xls/controllers/main.py
+++ b/controller_report_xls/controllers/main.py
@@ -54,8 +54,8 @@ def is_integer(value):
         return False
 
 
-def is_formatted_number(value):
-    set_sign = set([',', '.', '-'])
+def is_formatted_number(value, thousands_sep, decimal_point):
+    set_sign = set([thousands_sep, decimal_point, '-'])
     set_val = set(list(value))
     if set_val == set():
         return False
@@ -70,10 +70,8 @@ def is_formatted_number(value):
     elif '-' in set_val and not value[0] == '-':
         return False
 
-    # /!\ NOTE: Assumption. ',' is thousands separator
-    # /!\ NOTE: Assumption. '.' is decimal separator
-    if value.count(',') > 0:
-        value = value.replace(',', '')
+    if value.count(thousands_sep) > 0:
+        value = value.replace(thousands_sep, '')
     if is_integer(value):
         return True
     elif is_float(value):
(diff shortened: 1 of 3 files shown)


controller_report_xls/controllers/main.py
```diff
@@ -38,6 +38,96 @@
...
+    except ValueError:
+        return False
+
+
+def is_integer(value):
+    try:
+        int(value)
+        return True
+    except ValueError:
+        return False
```
nhomar, line 54: Why not the official isinstance?

controller_report_xls/controllers/main.py
```diff
@@ -38,6 +38,96 @@
...
+
+
+def is_formatted_number(value):
+    set_sign = set([',', '.', '-'])
```
nhomar: is it really necessary the set if the values are wired? this values can come from the lang in Odoo.

hbto: @nhomar Please check here, this is the first approach,

that is considered here

https://github.com/vauxoo-dev/addons-vauxoo/blob/236bb8277113a9548df25d6097915c533658b4cd/controller_report_xls/controllers/main.py#L109-L118

controller_report_xls/controllers/main.py
hbto, line 54: @nhomar Because value is a string that is trying to be converted into a number, e.g, casting into integer or float,

That is the reason

hbto, line 54: @nhomar it is always

isinstance(value, (int, float)) => False

because:

type(value) => string.

Regards

nhomar, line 54: Understood.

I personally prefer by far regex:

``` python`
import re
if re.match("^\d+?\.\d+?$", element) is None:
    print "Not float"
```

hbto, line 54: You have to realize that sometimes numbers are formatted as

> u'-7,777,777.77' 
> or 
> u'-7.777.777,77' 

this string should regarded as a number with your regex that won't happen

What we have intended is to unformat string into integer/float
reverse the process performed by `formatLang`

Which is complained here 

https://www.python.org/dev/peps/pep-0378/

> Some commenters do not like the idea of format strings at all and find them to be unreadable. Suggested alternatives include the COBOL style PICTURE approach or a convenience function with keyword arguments for every possible combination.

feketemihai:
@hbto @nhomar We talk in initial pr that i made aboout this...if you set Separator value to something like [3,3,3,3] and the language from Odoo corespond with the Locale of the client PC i think that the values were presented ok in excel...Can you check it like this...Thanks...

hbto:
@feketemihai 

>  We talk in initial pr that i made aboout this...if you set Separator value to something like [3,3,3,3] and the language from Odoo corespond with the Locale of the client PC i think that the values were presented ok in excel...Can you check it like this...Thanks...

Yes I have checked that and it is not working properly,

We have to unformat the text that has already been formatted with formatLang

Regards

oscarolar:
test

hbto:
@nhomar @moylop260 @feketemihai 

Your input / review / support is important here

@nhomar PR includes the use of res.lang to find Language Separator.

nhomar:
:+1:

controller_report_xls/controllers/main.py
```diff
@@ -38,6 +38,86 @@
...
+    except ValueError:
+        return False
+
+
+def is_formatted_number(value, thousands_sep, decimal_point):
+    set_sign = set([thousands_sep, decimal_point, '-'])
+    set_val = set(list(value))
+    if set_val == set():
+        return False
+    elif any([val.isalpha() for val in set_val]):
```
moylop260: What about `s/any([val.isalpha() for val in set_val]):/any([val.isalpha() or val.isspace() for val in set_val]):/g`?

moylop260:
I should prefer use dynamic `regex` base on `res_lang.decimal_separator` and `res_lang.thousand_separator` but I can't time to create a valid example and test it.

Then LGTM

(Just add tests in order to valid a wrong future refactoring)

feketemihai:
@hbto @nhomar @moylop260 For me it's ok even like this,  not tested what will improve the use of regex on decimal_separator  or thousand_separator, probably a more safe method than replace...

feketemihai:
:+1:

hbto:
`string_to_number` method got UnitTested
@moylop260 Could you check if that fulfil your expectations

Best Regards

moylop260:
@hbto 
Thank you man!
Should we create a dummy PR for projects where is used?

hbto:
@moylop260 Just need to refresh runbot for each project, 
as they have been PR'ed, I will just check for any missing 
project.

Regards

PR #1006, opened 2016-06-22
[8.0][message_post_model] Modified the value shown when a selection field is modified 

josemoralesp:
When a field selection is changed the value shown in the messages must be the string not the database value

FIX https://github.com/Vauxoo/yoytec/issues/1559


nhomar:
can you create a dummy in any project we use this module in order to test it?

josemoralesp:
@nhomar I did it https://github.com/Vauxoo/yoytec/pull/1564

PR #1018, opened 2016-07-11
[CC] Taking away code that is not implemented yet

hbto:
Taking away code that is not implemented yet


hbto:
@feketemihai 

As this code is not yet implemented 
I would like to get rid of it.

If you are to implement it later, you can retrieve the code from the CVS.

Best Regards.

feketemihai:
@hbto It's ok from my point of view...Thanks for announcing me...

PR #1019, opened 2016-07-12
[PYLINT & FLAKE8] Fixes

hbto:
Son new fixes due to changes on Pylint checks

### DIFF
--- a/controller_report_xls/controllers/main.py
+++ b/controller_report_xls/controllers/main.py
@@ -37,9 +37,7 @@ def is_formatted_number(value, thousands_sep, decimal_point):
     set_sign = set([thousands_sep, decimal_point, '-'])
     set_val = set(list(value))
     is_text = False
-    if set_val == set():
-        is_text = True
-    elif any([val.isalpha() or val.isspace() for val in set_val]):
+    if any([val.isalpha() or val.isspace() for val in set_val]):
         is_text = True
     elif not all([val.isdigit() for val in set_val - set_sign]):
         is_text = True
--- a/controller_report_xls/controllers/main.py
+++ b/controller_report_xls/controllers/main.py
@@ -40,7 +40,7 @@ def is_number(value):
         return False
 
 
-def is_string(value, thousands_sep, decimal_point):
+def is_string(value, thousands_sep=',', decimal_point='.'):
     """Tries to determine if value is not a numeric value, i.e. int or float"""
     set_sign = set([thousands_sep, decimal_point, '-'])
     set_val = set(list(value))
@@ -52,13 +52,15 @@ def is_string(value, thousands_sep, decimal_point):
         is_text = True
     elif value.count('-') > 1:
         is_text = True
+    elif value.count(decimal_point) > 1:
+        is_text = True
     elif '-' in set_val and not value[0] == '-':
         is_text = True
 
     return is_text
 
 
-def is_formatted_number(value, thousands_sep, decimal_point):
+def is_formatted_number(value, thousands_sep=',', decimal_point='.'):
     res = True
     if is_string(value, thousands_sep, decimal_point):
         res = False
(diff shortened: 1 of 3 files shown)


controller_report_xls/controllers/main.py
```diff
@@ -57,16 +57,20 @@ def is_integer(value):
...
+        is_text = True
     elif value.count('-') > 1:
-        return False
+        is_text = True
     elif '-' in set_val and not value[0] == '-':
-        return False
+        is_text = True
+
+    if is_text:
+        return value
```
nhomar: I think you do not need to check If is_text due to it is a boolean itself just retur the value. If not you will return  None.

hbto: You are right @nhomar, `lapsus mentis`
though it actually has to return `False`

hbto:
@nhomar @moylop260 

your input will be appreciated

Regards

hbto:
@nhomar @moylop260 

Merge at will

controller_report_xls/controllers/main.py
```diff
@@ -1,27 +1,4 @@
 # -*- coding: utf-8 -*-
-###########################################################################
```
moylop260, line 2: Please add the short license format: https://github.com/OCA/maintainer-tools/blob/master/template/module/controllers/main.py#L1-L3

zaoral, line 2: @moylop260 I think we could create an issue in the main vauxoo odoo repositories to set update python files with the short license format and also update the README.rst/index.html file to add the developer/planner/auditor of the module (extracted from the old license) to the Contributors [section](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst#contributors) in the module description. Are you agree?

moylop260, line 2: > Are you agree?

Yes I'm

Note: Just for the record, I don't requested replace the large-format by short-format... I just requested add a license... maybe short one (It was deleted)

zaoral, line 2: @moylop260  Mmm.. interesting. I have only one question: We should use AGPL-3 or LGPL-3 when creating odoo modules? I use LGL-3 and in the example yu mentioned the module template has AGPL-3

moylop260, line 2: AGPL-3.0 <- Public projects
LGPL-3.0 <- Private projects

moylop260, line 2: FYI A AGPL module can't depend of a LGPL

controller_report_xls/controllers/main.py
```diff
@@ -57,15 +34,19 @@ def is_integer(value):
 def is_formatted_number(value, thousands_sep, decimal_point):
     set_sign = set([thousands_sep, decimal_point, '-'])
     set_val = set(list(value))
+    is_text = False
     if set_val == set():
```
moylop260:
``` python
...
set_val = set(list(value))
if set_val == set():  # Is even True?
...
```

hbto:
@nhomar & @moylop260 Please check and provide guidance/feedback
@feketemihai, can you provide UnitTest for methods `get_css_style` & `get_odoo_style` in main.py file
and all those methods in xfstyle.py file?
@nhomar most of the code to be UnitTested is within class `ReportController` where PhanthomJS Tests are required.

Here is the latest coverage by itself of module `controller_report_xls`.
Coverage that could be provided by other modules is not shown here.

git_ _docker_exec_-it_afr-8_0-2_script_-q_-c_term_xterm_tmux_-s_afr-8_0-2_new_-s_afr-8_0-2_term_xterm_tmux_-s_afr-8_0-2_attach_-d_-t_afr-8_0-2__bin_bash__dev_null_ _176x51

Best Regards.

moylop260:
LGTM

controller_report_xls/controllers/main.py
zaoral, line 2: Good to know/
Thank you @moylop260

nhomar:
+1 for me just rebase.

PR #540, opened 2016-07-20
[IMP] modules version do not follow the OCA guideline

KarenKawaii:
module: user_story
version: 2.6

This module version number do not match with the OCA guidelines.
The version is  `2.6` but should be `8.0.2.6.0` 

Change here https://github.com/Vauxoo/addons-vauxoo/blob/8.0/user_story/__openerp__.py#L28

Maybe another modules are in the same condition.

NOTE: This [script](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/.version.py) was created to update the version of modules of this repository. Maybe this script can be modificate and use to update this module version and another modules in the same situation.

### DIFF
--- a/account_analytic_active_field/__openerp__.py
+++ b/account_analytic_active_field/__openerp__.py
@@ -21,7 +21,7 @@
 ##########################################################################
 {
     "name": "Field Active In Analytic Accounts",
-    "version": "1.6",
+    "version": "8.0.0.1.6",
     "author": "Vauxoo",
     "category": "Accouting",
     "website": "http://www.vauxoo.com/",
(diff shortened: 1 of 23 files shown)


KarenKawaii:
Hi @osvalr 
About of your doubt, the fixed versions in this PR follow the format applied in: #823 

I have my doubts too like you, but as @moylop260 said in: https://github.com/Vauxoo/addons-vauxoo/pull/823#discussion_r70474333
most versions are a copy&paste so, I think is better let the reference in the version for now just for standardize (in new modules we have to take care of using the [semantic versioning](http://semver.org/lang/es/))

@moylop260 these are the missing modules that didn't have the version format: `8.0.x.y.z`

### [#VX5706](https://www.vauxoo.com/web#id=5706&view_type=form&model=project.task&menu_id=136&action=138)

KarenKawaii:
@moylop260 could you check this PR please?

PR #1026, opened 2016-07-22
[FIX] Missing @api.v8 definition for action_done as it was defined as @api.v7

hbto:
Missing @api.v8 definition was causing some problem with UnitTests at downstream projects as they are using @api.v8 code.


hbto:
On Customer Project this PR went well, errors are unrelated to diff

Issue #465, opened 2015-09-29
Module 'product_do_merge' will not work when module 'delivery' is installed


JordiBForgeFlow:
Dear Vauxoo team,

We have identified that the module 'product_do_merge' will not be able to merge two products when the module 'delivery' is installed.

The root cause of this issue is that the module 'delivery' adds deliverately the unit of measure 'kg' in stock moves, and this UoM does not necessarily have to be in the same category as the product. 

As you know, your module will raise an error message if it detects a UoM in any object that refers to a product, that is not in the same UoM category.

Since this weight UoM added to stock moves is really dummy (Kg by default, with no possibility to change), we have added a module that will default this field weight unit to blank in stock moves.
https://github.com/Eficent/eficent-odoo-addons/tree/7.0/delivery_no_weight_unit

Once you install this module and make an update SQL query to delete all data from this field in the database, it will be possible to merge products.

I'm not sure if you have had this issue before, and perhaps you know a better way to approach this, but this is what we've done.

Regards,
Jordi.

PR #1029, opened 2016-07-27
[IMP][8.0][product_do_merge] Fix issue when delivery module is installed

JordiBForgeFlow:
Fixes issue #465 

With this PR..

It is possible to exclude some UoM fields in certain models from the check of conflicting UoM's. 

For example, the field Weight UoM introduced by module 'Delivery' will be often set in a UoM for the field 'Weight Unit of Measure' with a dimension different than that of the product. For example, deliver 1 Unit(s), with weight = 700 Kg.

In normal circumstances the merge process will fail because "Kg" is not in the same dimension as the product's base Uom (Unit).

In order to exclude a specific field, please go to 'Settings / Technical / Database structure / Exclude UoM from Product Merge', then add the field related to the product UoM that you want to exclude from the merge criteria.

### DIFF
--- a/product_do_merge/README.rst
+++ b/product_do_merge/README.rst
@@ -16,5 +16,13 @@ We can select which product will be the main product.
 This feature do not change anything if the products to be merged have
 operations in different units of measure.
 
+It is possible to exclude some UoM fields in certain models from the above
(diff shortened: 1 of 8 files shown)


luisg123v:
This seems abandoned, so I'll close it.

PR #1033, opened 2016-08-08
[ADD] Feature of autorized service centers related to the right model and template improvements HU#1006 CA#63

oscarolar:
Build https://github.com/Vauxoo/yoytec/pull/1734

Result:

authorized_service_providers___www_yoytec_com


website_authorized_service_centers/demo/brands_and_partners_demo.xml
```diff
@@ -0,0 +1,196 @@
+<?xml version="1.0" encoding="utf-8"?>
```
oscarolar, line 1: @josemoralesp all the demo data created here must be created as legacy data, I brought it from the actual yoytec website, this info was not in the database.

PR #1036, opened 2016-08-11
[IMP] default_warehouse_from_sale_team: defined 'picking_type_id'

JulioSerna:
default as well from the sale team werahouse

### DIFF
--- a/default_warehouse_from_sale_team/__openerp__.py
+++ b/default_warehouse_from_sale_team/__openerp__.py
@@ -14,6 +14,7 @@
         'sale_stock',
         'sales_team',
         'base_action_rule',
+        'purchase_requisition',
     ],
     'test': [
     ],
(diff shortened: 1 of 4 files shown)


nhomar:
@JulioSerna can you add  a test for your new method please?

JulioSerna:
@nhomar 
Test done
and the PR will be tested in this [DUMMY](https://github.com/Vauxoo/yoytec/pull/1811)

nhomar:
@JulioSerna can you point me out the User Story and/or the explanation/use case about this? it is so light the explanation you share here and I do not understand at first view why this is needed.

JulioSerna:
@nhomar 

it is about the [HU#1129](https://www.vauxoo.com/web#id=1552&view_type=form&model=project.issue&action=578)
And the [Issue#1552](https://www.vauxoo.com/web#id=1552&view_type=form&model=project.issue&action=578)

When user make a purchase order the picking type must be one related to sale team warehouse

PR #1042, opened 2016-08-23
[IMP] default_warehouse_from_sale_team: propagate SUPERUSER_ID in read

JulioSerna:
when user create from sale pickings chained in another warehouse without
access to read

e.g.
Warehouse A
Warehouse B
User A with access to Warehouse A
User B with access to Warehouse B
Route A with resupply Warehouse B → Warehouse A
Create sale order with **Route A** with **User A**

**Sale order created the pickings:**
- Warehouse B → Transito
- Transito → Warehouse A

**User A** must see only the picking **Transito → Warehouse A**
**User B** must see only the picking **Warehouse B → Transito**

When **User B** validated  **Warehouse B → Transito**  the picking **Transito → Warehouse A** must change to **ready to transfer** and if User A cancelled the picking  **Transito → Warehouse A** the picking **Warehouse B → Transito** must change to cancelled as well,
and then, **User A** and **User B** need the **write** and **read** in another warehouse but they can not see the records in web

### DIFF
--- a/default_warehouse_from_sale_team/models/sales_team.py
+++ b/default_warehouse_from_sale_team/models/sales_team.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 
 from openerp import api, fields, models
-
+from openerp import SUPERUSER_ID
 
 class InheritedCrmSaseSection(models.Model):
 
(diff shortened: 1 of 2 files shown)


default_warehouse_from_sale_team/models/sales_team.py
```diff
@@ -88,3 +88,27 @@ def default_get(self, fields_list):
...
+                cr, SUPERUSER_ID, ids, fields=fields, context=context,
+                load=load)
+        return super(WarehouseDefault, self).read(
+            cr, user, ids, fields=fields, context=context, load=load)
+
+    @api.v8
+    def read(self, fields=None, load='_classic_read'):
+        """This method is overwrite because we need to propagate SUPERUSER_ID
+        when picking are chained in another warehouse without access to read"""
+        if self.env['ir.model.access'].check_groups(
```
zaoral: @JulioSerna 

in v8 why do you not use self.env.user.has_group('default_warehouse_from_sale_team.group_limited_default_warehouse_sp')

JulioSerna: @zaoral 
What is the different?
Why one is better than other?

JulioSerna:
This pr solves issue vauxoo/yoytec#1879

default_warehouse_from_sale_team/models/sales_team.py
zaoral: Hi @JulioSerna 

As you can see in the next compare image of the two methods both has the same sql sentence, however:
1. has_group will throw you a message if the group do not exist or do a bad reference while check_groups don't.
2. has_group has cache extra management.

image

Best Regards

JulioSerna:
@zaoral  changes applied

@nhomar 
Could you merge this pr?

The pr was tested in this [PR DUMMY](https://github.com/Vauxoo/yoytec/pull/1892)

PR #1043, opened 2016-08-23
[REF] website_authorized_service_centers: Show correctly all the desc…

hugho-ad:
…ription of each product.brand because it was no showed by a alc problem

FIX https://github.com/Vauxoo/yoytec/issues/1896

Build https://github.com/Vauxoo/yoytec/pull/1897

### DIFF
--- a/website_authorized_service_centers/views/authorized_service_centers.xml
+++ b/website_authorized_service_centers/views/authorized_service_centers.xml
@@ -16,7 +16,7 @@
                     </section>
                     <section class="mt16 mb16">
                         <div class="container">
-                            <t t-set="providers" t-value="env['product.brand'].search([('authorized_partner_ids', '!=', False)], order='name desc')"/>
+                            <t t-set="providers" t-value="env['product.brand'].sudo().search([('authorized_partner_ids', '!=', False)], order='name desc')"/>
                             <t t-as="provider" t-foreach="providers">
                                 <t t-call="website_authorized_service_centers.authorized_centers"/>
                             </t>


hugho-ad:
@oscarolar what do you think ?
Instead of the `sudo()` stament, I could use a new controller, to return the set of partners

oscarolar:
👍  keep it simple.

oscarolar:
Dude, I dont see any runbot build here

hugho-ad:
@oscarolar see the Description 
https://github.com/Vauxoo/yoytec/pull/1897

oscarolar:
I mean runbot buuld for addons-vauxoo, only travis build here, is there a problem here @moylop260 ?

moylop260:
The addons vauxoo don't use runbot.
It just use travis but without tests

hugho-ad:
Runbot Green, this is ready to be merge

website_authorized_service_centers/views/authorized_service_centers.xml
```diff
@@ -16,7 +16,7 @@
                     </section>
                     <section class="mt16 mb16">
                         <div class="container">
-                            <t t-set="providers" t-value="env['product.brand'].search([('authorized_partner_ids', '!=', False)], order='name desc')"/>
+                            <t t-set="providers" t-value="env['product.brand'].sudo().search([('authorized_partner_ids', '!=', False)], order='name desc')"/>
```
moylop260, line 19: @JulioSerna @nhomar 
Should we use `sudo` or should we use a new ACL to allow read `product.brandh` for portal users?

hugho-ad: @moylop260 the ACL that you suggest us, already exists:
https://github.com/Vauxoo/product-attribute/blob/8.0/product_brand/security/ir.model.access.csv#L3

The ACL requiered is for the model res.partner, because the info aren't returned and rendered correctly caused by the lack of the permissions over the res.partner model.

I like the use of the stament `sudo()` in the template instead of ACL soluction because it's not adding an extra granting permissions also , the `sudo()` stament only works in backend, returning pure html after qweb process.

moylop260: I got it
Thank you

PR #1044, opened 2016-08-24
[IMP] stock_landed_costs_average: the invoices created from landed

JulioSerna:
should be created like supplier invoices instead customer invoices


JulioSerna:
@hbto 

Could you check and merge this pr?

PR #1047, opened 2016-08-29
[REF] website_authorized_service_centers: Set attr store = True to product_count field to avoid calculation.

hugho-ad:


website_authorized_service_centers/model/product_brand.py
```diff
@@ -27,3 +27,8 @@ class ProductBrand(models.Model):
 
     authorized_partner_ids = fields.Many2many('res.partner')
     url = fields.Char(string="URL")
+    products_count = fields.Integer(
+        string='Number of products',
+        compute='_get_products_count',
+        store=True,
+    )
```
nhomar, line 34: Why are you overwritting a field here, instead correct the original module?

hugho-ad: You rigth, this change must be in the original module
Thanks

PR #1050, opened 2016-09-03
[REF] stock_quant_cost_segmentation: Faster 'segmentation_cost' computed

moylop260:
Fix https://github.com/Vauxoo/yoytec/issues/1962
Dummies PR: https://github.com/Vauxoo/yoytec/pull/1965 https://github.com/Vauxoo/lodigroup/pull/605

### DIFF
--- a/stock_quant_cost_segmentation/model/stock.py
+++ b/stock_quant_cost_segmentation/model/stock.py
@@ -119,7 +119,7 @@ class StockQuant(models.Model):
     @api.multi
     def write(self, vals):
         res = super(StockQuant, self).write(vals)
-        if not (set(SEGMENTATION_COST).issubset(set(vals)) or
+        if not (set(vals) & set(SEGMENTATION_COST) or
                 self.env.context.get('force_segmentation_cost')):
             return res
         # TODO: Validate sql injection from SEGMENTATION_COST variable
--- a/stock_quant_cost_segmentation/model/stock.py
+++ b/stock_quant_cost_segmentation/model/stock.py
@@ -109,10 +109,18 @@ class StockMove(models.Model):
 class StockQuant(models.Model):
     _inherit = "stock.quant"
 
+    @api.model
+    def create(self, vals):
+        segmentation_cost = sum([
+            getattr(self, field_name) for field_name in SEGMENTATION_COST])
+        vals.update({'segmentation_cost': segmentation_cost})
+        return super(StockQuant, self).create(vals)
+
     @api.multi
     def write(self, vals):
         res = super(StockQuant, self).write(vals)
-        if not set(SEGMENTATION_COST).issubset(set(vals)):
+        if not set(SEGMENTATION_COST).issubset(set(vals)) or \
+           not self.env.context.get('force_segmentation_cost'):
             return res
         # TODO: Validate sql injection from SEGMENTATION_COST variable
         # Because other module could add a monkey patch with sql injection
--- a/stock_quant_cost_segmentation/model/stock.py
+++ b/stock_quant_cost_segmentation/model/stock.py
@@ -119,8 +119,8 @@ class StockQuant(models.Model):
     @api.multi
     def write(self, vals):
         res = super(StockQuant, self).write(vals)
-        if not set(SEGMENTATION_COST).issubset(set(vals)) or \
-           not self.env.context.get('force_segmentation_cost'):
+        if not (set(SEGMENTATION_COST).issubset(set(vals)) or
+                self.env.context.get('force_segmentation_cost')):
             return res
         # TODO: Validate sql injection from SEGMENTATION_COST variable
         # Because other module could add a monkey patch with sql injection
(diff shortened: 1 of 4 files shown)


nhomar:
@moylop260

Can you add a build into Lodi instances @sabrinagnu can give you the points.

Question: I suppose it is more secure this way than change the computed method right?

IF YES:

I think this should have:

1.- Increase the version number of the module.
2.- add a migration process which pre-compute (pure SQL) all the values before upgrade (it is executed in the middle of the -u all and will give us prepared all the values in the database).

In production evironments the value with just install the module will left the elements empty.

stock_quant_cost_segmentation/model/stock.py
```diff
@@ -109,12 +109,28 @@ def product_segmentation_update_after_done(
...
+        if not set(SEGMENTATION_COST).issubset(set(vals)) or \
+           not self.env.context.get('force_segmentation_cost'):
+            return res
+        # TODO: Validate sql injection from SEGMENTATION_COST variable
+        # Because other module could add a monkey patch with sql injection
+        sum_query = ' + '.join(["COALESCE(%s, 0)" % field_name
+                                for field_name in SEGMENTATION_COST])
+        # NOTE: Cache is too slow to manage many records from computed field
+        query = "UPDATE stock_quant SET segmentation_cost = (" + sum_query + \
+            ") WHERE id IN %s"
```
nhomar, line 131: I think in order to avoid a write we should have a little IF:

IF computation is equal to what the field already has then do not write.

josemoralesp: @nhomar the write was changed by an update of postgres. The answer of the update is quick. I tried with a database of apex and the behavior was the expected. I think that this change is correct, only update the field qusers using postgres Engine

moylop260: You are right!
We have a early return  if we don't have the values from `SEGMENTATION_COST` fields with a `union`:
`set(vals) & set(SEGMENTATION_COST)`

hbto:
When installing a fresh instance --without-demo=all
following error appears @moylop260 do you have a proposal to solve it.

```
2016-09-06 02:52:46,655 1587 INFO asset openerp.modules.module: module stock_quant_cost_segmentation: creating or updating database tables
2016-09-06 02:52:47,063 1587 INFO asset openerp.modules.loading: loading stock_quant_cost_segmentation/view/view.xml
2016-09-06 02:52:47,344 1587 INFO asset openerp.modules.loading: loading stock_quant_cost_segmentation/data/data.xml
2016-09-06 02:52:47,358 1587 ERROR asset openerp.sql_db: Programming error: syntax error at or near ")"
LINE 1: ...rial_cost, 0) + COALESCE(production_cost, 0)) WHERE id IN ()
                                                                      ^
, in query UPDATE stock_quant SET segmentation_cost = (COALESCE(landed_cost, 0) + COALESCE(subcontracting_cost, 0) + COALESCE(material_cost, 0) + COALESCE(production_cost, 0)) WHERE id IN %s
2016-09-06 02:52:47,358 1587 CRITICAL asset openerp.service.server: Failed to initialize database `asset`.
  File "/.repo_requirements/odoo/openerp/service/server.py", line 936, in preload_registries
    registry = RegistryManager.new(dbname, update_module=update_module)
  File "/.repo_requirements/odoo/openerp/modules/registry.py", line 370, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/.repo_requirements/odoo/openerp/modules/loading.py", line 355, in load_modules
    loaded_modules, update_module)
  File "/.repo_requirements/odoo/openerp/modules/loading.py", line 255, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
  File "/.repo_requirements/odoo/openerp/modules/loading.py", line 188, in load_module_graph
    getattr(py_module, post_init)(cr, registry)
  File "/root/addons-vauxoo/stock_quant_cost_segmentation/hooks.py", line 12, in post_init_hook
    env['stock.quant'].search([]).write({})
  File "/.repo_requirements/odoo/openerp/api.py", line 267, in wrapper
    with profiling(): return new_api(self, *args, **kwargs)
  File "/root/addons-vauxoo/stock_quant_cost_segmentation/model/stock.py", line 132, in write
    self.env.cr.execute(query, (tuple(self.ids),))
  File "/.repo_requirements/odoo/openerp/sql_db.py", line 158, in wrapper
    return f(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
ProgrammingError: syntax error at or near ")"
LINE 1: ...rial_cost, 0) + COALESCE(production_cost, 0)) WHERE id IN ()
                                                                      ^

2016-09-06 02:52:47,359 1587 INFO asset openerp.service.server: Initiating shutdown
2016-09-06 02:52:47,359 1587 INFO asset openerp.service.server: Hit CTRL-C again or send a second signal to force the shutdown.
```

Best Regards

moylop260:
Just add a `if` to validate that `len(self.ids) > 0`

hbto:
Dealing with this last error on:

https://github.com/Vauxoo/addons-vauxoo/pull/1053

PR #1053, opened 2016-09-06
[FIX] Avoid rewrite if no ids to write

hbto:


hbto:
@moylop260 could please review / sanction / merge

Best Regards

moylop260:
👍 

``` python
    if not (set(vals) & set(SEGMENTATION_COST) or
               self.env.context.get('force_segmentation_cost')) or
              self.ids
```

My early return basic tests:
screen shot 2016-09-05 at 10 30 33 pm

PR #1055, opened 2016-09-07
[FIX] When Creating records values are not yet written in records only on dict

hbto:
So fetching the segmentation origin values can not be possible at early stages.

current_stock


hbto:
@moylop260 Please review / sanction/ censor / merge

Best regards

moylop260:
+1

PR #1056, opened 2016-09-07
[FIX] stock_quant_cost_segmentation: Fix sum Null value

moylop260:
screen shot 2016-09-06 at 9 33 35 pm

Executed in last green of runbot with this branch:

screen shot 2016-09-06 at 9 44 09 pm

Fix https://github.com/Vauxoo/yoytec/issues/1982

### DIFF
--- a/stock_quant_cost_segmentation/model/stock.py
+++ b/stock_quant_cost_segmentation/model/stock.py
@@ -113,7 +113,8 @@ class StockQuant(models.Model):
     def create(self, vals):
         vals.update(
             {'segmentation_cost': sum(
-                [vals.get(field_name, 0) for field_name in SEGMENTATION_COST])})
+                [vals.get(field_name) or 0
+                 for field_name in SEGMENTATION_COST])})
         return super(StockQuant, self).create(vals)
 
     @api.multi


moylop260:
@hbto your feedback is required here, please

hbto:
@moylop260 you saw and old error, It was fixed when I wrote 

`vals.get(field_name, 0)` => this yields a number or zero

previously

`vals.get(field_name)` => this could yield a number or None

There was the problem,

so there was no need for a fix.

regards

hbto:
@moylop260 you saw and old error, It was fixed when I wrote 

`vals.get(field_name, 0)` => this yields a number or zero

previously

`vals.get(field_name)` => this could yield a number or None

There was the problem,

so there was no need for a fix.

regards

moylop260:
@hbto 
Could you execute the following code please?

``` python
SEGMENTATION_COST = ['field1']
vals = {'field1': None}

print "result with or: ",
result = sum([
    vals.get(field_name) or 0
    for field_name in SEGMENTATION_COST])
print result


print "result without or: ",
result = sum([
    vals.get(field_name, 0)
    for field_name in SEGMENTATION_COST])
print result
```

hbto:
@moylop260 

I've got your point and I have come to realise why
I couldn't foresee your point,

When creating vals does not provide provide a
key with value None,

that was the reason I overlooked it,

Regards

Issue #1054, opened 2016-09-06
[product_available_by_warehouse] domain of routes must be routes in sale order line


JulioSerna:
When use **Show message** to see product available the domain in route are modfied but the domain must be consider the router that only are available in sale order line
## with **Show message**

![sale_order_route_domain](https://cloud.githubusercontent.com/assets/7595505/18285974/5172e64a-7436-11e6-9141-b081ae35f818.png)

---
## without **Show message**

![sale_order_route](https://cloud.githubusercontent.com/assets/7595505/18285981/5a53e9c6-7436-11e6-8a7a-64dfe676f1a5.png)
- Expected behavior

Shows routes available in **sale order line** and there are routes from locations usage **supplier** like **Dropshipping** that must be available as well

suniagajose:
I got it @JulioSerna... 

i will work on the following TODO list :
- [x] avoid route_ids domain in 'show message' onchange.
- [x] use sudo in warehouse search to show always the same info no matter what permission has the user.
- [x] swap positions between show_message field and product_id field

If i forgot anything or you have a additional requirement, please share it...

Regards,

PR #1059, opened 2016-09-14
[REF] product_available_by_warehouse: remove domain route in show message

suniagajose:
## Summary

When it was used 'show message' field to see product available the domain route should consider only the routes that are available for sale order line.

To solve #1054 
## TODO
- [x] avoid route_ids domain in 'show message' onchange.
- [x] use sudo in warehouse search to show always the same info no matter what permission has the user.
- [x] swap positions between show_message field and product_id field

### DIFF
--- a/product_available_by_warehouse/models/sale.py
+++ b/product_available_by_warehouse/models/sale.py
@@ -45,9 +45,11 @@ class SaleOrderLine(models.Model):
             product_qty_by_wh = product.\
                 get_product_available_by_warehouse()[product.id][colname]
         for warehouse, product_qty in product_qty_by_wh:
-            if product_qty > 0.0 and \
-                    colname in ['qty_available', 'virtual_available']:
-                route_ids += [warehouse.delivery_route_id.id]
+            # TODO: uncomment until figure out a better way to choice only
+            # the routes with availability
+            # if product_qty > 0.0 and \
+            #         colname in ['qty_available', 'virtual_available']:
+            #     route_ids += [warehouse.delivery_route_id.id]
             warning_msgs += _("- %s . Quantity: %.2f \n" %
                               (warehouse.name, product_qty))
         res = {}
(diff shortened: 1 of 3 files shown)


suniagajose:
Hi @JulioSerna 

Can you review this, please?

product_available_by_warehouse/models/sale.py
```diff
@@ -62,7 +62,8 @@ def onchange_show_message(self, show_message, product_id):
         # update route domain
         if route_ids:
             domain = res.get('domain', {})
-            domain['route_id'] = [('id', 'in', route_ids)]
+            # TODO: uncomment until figure out a better way
+            # domain['route_id'] = [('id', 'in', route_ids)]
             res.update({'domain': domain})
```
nhomar, line 68: sorry but I do not understand this.

You are passing a blank domain and wnat's the intention of this part.

I expect a unittest here if not we will be doing error-test until every fail generate hours of support.

Please create a unittest that given the border case return the quantity expected./

product_available_by_warehouse/views/sale_view.xml
```diff
@@ -7,7 +7,7 @@
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale_stock.view_order_form_inherit" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='order_line']/form//field[@name='product_id']" position="before">
+            <xpath expr="//field[@name='order_line']/form//field[@name='product_id']" position="after">
```
nhomar, line 10: why are you changing the view? at least this is not mandatory in the requirement described in the task/issue.

product_available_by_warehouse/models/sale.py
JulioSerna: @nhomar 

this code do not passing blank domain, in fact is passing the exactly same domain returned by the original onchange

this change is because before it was modifying in wrong way and that is the reason of the change and  `TODO`  

[related issue](https://github.com/Vauxoo/addons-vauxoo/issues/1054)

product_available_by_warehouse/views/sale_view.xml
JulioSerna: @nhomar 
in fact, it is requirement for the customer, in usability is better to choose first the product
produtc_after
 than this
produt_before

JulioSerna:
LGTM
👍

PR #1061, opened 2016-09-15
[REF] sale_order_due_date: Use another demo sale order

KarenKawaii:
This is to avoid the following error when this module is
installed in isolated way and then is use
-u sale_order_due_date --test-enable

2016-09-09 03:17:32,999 9358 ERROR openerp_test_sale_order_due_date
openerp.addons.sale_order_due_date.tests.test_sale_due_date: `
ValueError: Wrong value for sale.order.order_policy: u'prepaid'`

**PR dumy:** https://github.com/Vauxoo/yoytec/pull/2004

**Log error:** http://runbot.vauxoo.com/runbot/static/build/17488-1880-34756f/logs/job_20_test_all.txt

### DIFF
--- a/sale_order_due_date/models/sale_order.py
+++ b/sale_order_due_date/models/sale_order.py
@@ -7,8 +7,8 @@
 #    info Vauxoo (info@vauxoo.com)
 #    coded by: Luis Torres <luis_t@vauxoo.com>
 ############################################################################
-from openerp import models, fields, api
 from datetime import date, timedelta
+from openerp import models, fields, api
 
 
 class SaleOrder(models.Model):
--- a/sale_order_due_date/tests/test_sale_due_date.py
+++ b/sale_order_due_date/tests/test_sale_due_date.py
@@ -7,8 +7,8 @@
 #    info Vauxoo (info@vauxoo.com)
 #    coded by: Luis Torres <luis_t@vauxoo.com>
 ############################################################################
-from openerp.tests import common
 from datetime import date, timedelta, datetime
+from openerp.tests import common
 
 
 class TestSaleDueDate(common.TransactionCase):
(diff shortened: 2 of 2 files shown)


KarenKawaii:
Good morning @luistorresm and @moylop260 

This PR is to avoid the error in the `sale_order_due_date` module.
Are you agree with this change?
Also I repaired the pylint errors.

The PR dummy in yoytec is in the description and was green in all the tests.

Best Regards.

moylop260:
What about create a new sale order record?

luistorresm:
@KarenKawaii 

What is the problem with the order_policy as "prepaid"?

This error is found only for yoytec?

Regars

KarenKawaii:
Hi @luistorresm yes this happen if you use a db based in openerp_template for yoytec, so instead of use a sale_order without policy i used another that have a policy_order to avoid the error.

@moylop260 Use a new record created like data demo will solve the problem too. I can add the order_policy in that new record.

So, what do you think?

KarenKawaii:
After debugging and explaining to @luistorresm the error that appears in `sale_order_due_date` related to `prepaid` `order_policy`we agreed to create a `sale_order` demo as @moylop260 suggested but via python into the test.

This change is already applied and the error isn't triggered.
captura de pantalla 2016-09-21 a las 12 31 59 a m

The explaining of the error is in:
https://docs.google.com/document/d/1JWA607bQWKHiBCp45kcq5486cSozoOHqqnzwGgqMtGU/edit#

Could you review it again, please?
Regards.

luistorresm:
Thanks @KarenKawaii 

👍

KarenKawaii:
You're welcome

@moylop260 The latest results in the yoytec pr dummy are green.

Regards

PR #1067, opened 2016-09-28
[IMP] partner_validate_vat: Migrated to v9

luistorresm:
Migrated method to v9, with new api.

Added tests in module


luistorresm:
@moylop260 

Could you check this PR please?

Regards

PR #1069, opened 2016-09-29
[IMP] Print stock_card at logger

hbto:
[IMP] Print stock_card at logger in order to debug later when failing and find cause of random failures of tests

2_ _docker_exec_-it_yoytec-8_0_script_-q_-c_term_xterm_tmux_-s_yoytec-8_0_new_-s_yoytec-8_0_term_xterm_tmux_-s_yoytec-8_0_attach_-d_-t_yoytec-8_0__bin_bash__dev_null_ _168x48

### DIFF
--- a/stock_card/tests/test_stock_card_negative_stock.py
+++ b/stock_card/tests/test_stock_card_negative_stock.py
@@ -27,6 +27,15 @@ from openerp.tests.common import TransactionCase
 
 _logger = logging.getLogger(__name__)
 
+try:
+  import pandas
+except ImportError:
+  _logger.debug('Cannot `import pandas`.')
+
+try:
+  import tabulate
+except ImportError:
+  _logger.debug('Cannot `import tabulate`.')
 
 class TestStockCardNegativeStock(TransactionCase):
 
--- a/stock_card/tests/test_stock_card_negative_stock.py
+++ b/stock_card/tests/test_stock_card_negative_stock.py
@@ -18,24 +18,21 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-import logging
 from datetime import datetime, timedelta
-from tabulate import tabulate
-import pandas as pd
-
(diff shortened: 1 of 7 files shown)


stock_card/tests/test_stock_card_negative_stock.py
```diff
@@ -177,6 +189,15 @@ def test_01_do_inouts(self):
...
 
         card_lines = self.get_stock_valuations()
+
+        df = pd.DataFrame(card_lines)
+        tbl_sc = tabulate(df, headers='keys', tablefmt='psql')
+        _logger.info('Gotten Stock Card \n%s', tbl_sc)
+
+        df = pd.DataFrame(self.inv_ids)
+        tbl_sc = tabulate(df, headers='keys', tablefmt='psql')
+        _logger.info('Expected Stock Card \n%s', tbl_sc)
```
zaoral, line 199: Hi @hbto 
I just see this and that remember me a change I made that is also used as debug info.
Was made in the same context: print debug information.

The difference is that instead print the information inside the method using logger, I used a method that return
and string.

``` diff
 self.assertEqual(len(self.inv_ids), len(card_lines),
-                           "Both lists should have the same length(=12)")
+                          "Both lists should have the same length(=12)" + self.debug_info())
```

This way you will not need to import logger, and the debug information will be print only when the error occurs. This information will be print in the sever log with the `ERROR...` prefix, so will be easy to find, You do not need to print the debug info in the tests and then look down for the unit test that fails and see what tests match: the result will be clean, easy to find and only printed when necessary indicating you what tests is failing.

NOTE: If you like you can have the example of the debug method I create in [here](https://github.com/Vauxoo/yoytec/blob/8.0/sale_cash_register/tests/test_sale_cash_register.py#L113)

hbto: @zaoral It is intended that the message be printed regardless of the failure happening.

Regards.

PR #1086, opened 2016-10-17
[ADD] default_warehouse_from_sale_team: Constrain to default_warehouse field.

ysantiago:
Add constrain to validate that does not exist more that one sale team with the same default_warehouse.

Close [Vauxoo/typ#80](https://github.com/Vauxoo/typ/pull/80)
Close [Vauxoo/yoytec#2054](https://github.com/Vauxoo/yoytec/pull/2054)


luisg123v:
This is abandoned and typ is not on v8 anymore, so closing.

PR #1089, opened 2016-10-21
[FIX] Segmention on Average Products and Material Cost when Segmentation is filled

hbto:
- Server action for updating segmentation fields properly in AVG products in order to fetch right values in STD products when updating with Product Extended Segmentation.
- Initialization script for quants when material cost is zero but other segmentation fields sum is not zero and cost in quant isn't zero either, retrieve material cost from those values.

### DIFF
--- a/product_extended_segmentation/data/data.xml
+++ b/product_extended_segmentation/data/data.xml
@@ -22,5 +22,26 @@ self.update_material_cost_on_zero_segmentation(cr, uid, [])
             </field>
         </record>
 
+        <record model="ir.actions.server" id="update_product_segmentation_from_stock_card">
+            <field name="name">Update Segmentation on Average Product from Stock Card</field>
+            <field name="model_id" eval="ref('product.model_product_product')"/>
+            <field name="state">code</field>
+            <field name="condition"></field>
+            <field name="sequence">5</field>
+            <field name="code">
+# Available locals:
+#  - time, datetime, dateutil: Python libraries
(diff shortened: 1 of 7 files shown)


product_extended_segmentation/data/data.xml
```diff
@@ -22,5 +22,26 @@ self.update_material_cost_on_zero_segmentation(cr, uid, [])
             </field>
         </record>
 
+        <record model="ir.actions.server" id="update_product_segmentation_from_stock_card">
+            <field name="name">Update Segmentation on Average Product from Stock Card</field>
+            <field name="model_id" eval="ref('product.model_product_product')"/>
+            <field name="state">code</field>
+            <field name="condition"></field>
```
nhomar, line 29: can you add a proper condition in order to avoid run this method in 100% of product when -u all?

nhomar (changes requested):

product_extended_segmentation/data/data.xml
hbto: @nhomar 
This Server Action is to be called on demand, it will not be triggered on `-u all` command.

hbto:
@nhomar 
Except for McCabe rating at 23 for `_calc_price` method
This PR does not present any other relevant issue,
Dummy PR on Customer Runbot is Yellow, because of some
unrelated warnings.

If you don't have any other advice please merge.

Best Regards

PR #1092, opened 2016-10-26
[BUG] account_voucher_tax: Frozendict error on context mutations

deivislaya:
**Impacted versions:**
8.0

**Steps to reproduce:**
1. Create Unit Test calling  `process_reconciliation` method

**Current behavior:**

The context is now immutable, for new API,  this causes by calling the method generates the following error:

``` bug
NotImplementedError: '__setitem__' not supported on frozendict
```

caused by

``` python
        context['date'] = st_line.date
        context['st_line_currency_rate'] = st_line_currency_rate
```

**Expected behavior:** 

To call the method from unit tests successfully since version 8 and greater

complete original code [here](https://github.com/Vauxoo/addons-vauxoo/blob/8.0/account_voucher_tax/account_statement_bank.py#L122)

A viable solution is to place the context in method parameters as follows:

``` diff
+    context_2 = context.copy()
```

### DIFF
--- a/account_voucher_tax/account_statement_bank.py
+++ b/account_voucher_tax/account_statement_bank.py
@@ -124,33 +124,35 @@ class AccountBankStatementLine(osv.osv):
         if context is None:
             context = {}
 
+        context_2 = context.copy()
         move_line_obj = self.pool.get('account.move.line')
         move_obj = self.pool.get('account.move')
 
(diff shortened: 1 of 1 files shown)


deivislaya:
Hi @JulioSerna 

This PR is ready for review

JulioSerna:
👍 
[pr dummy](https://github.com/Vauxoo/typ/pull/94)

Issue #1093, opened 2016-10-27
[9.0] Error installing mrp_production_analytic_acc


kmlew:
Hi!  

I'm using Odoo version 9.0c-20160620 (Community Edition).  While installing mrp_production_analytic_acc, I encounter the following below.  Please let me know if there's a fix for it.

![mrp account analytics odoo](https://cloud.githubusercontent.com/assets/17688618/19761831/ab5e9e6c-9c61-11e6-9e61-883a46610353.png)

zaoral:
Hi @kmlew 
Thank you for reporting.

It seems that this module was created in 8.0 and do not have any problem so far.
Since you are using 9.0 version maybe this is a new restriction for this particular version.

I make a quick review and modify the view code in order to avoid this error just making the xpath more simple. You can review the change made in #1094: Please you can try to install mrp_production_analytic_acc module with the change made and let me know if the error disappears

kmlew:
Hello @zaoral,

There are a few more lines similar to the one you changed further which also causes the installation to fail.  I have manually corrected them in my copy and it installs.  Unfortunately, I am new to GitHub and am not sure how to share my changes.  Perhaps you can give me some tips?

KM

zaoral:
Hi @kmlew 
Yes, you can:

You can do it yourself following the vauxoo's contribution guide [here}(https://github.com/Vauxoo/addons-vauxoo/blob/8.0/CONTRIBUTING.md#how-add-a-new-feature) If you feel comfortable using git commands: 
1. from github fork vauxoo/addons-vauxoo to your user account.
2. clone and configure locally your new repository kmlew/addons-vauxoo
3. create a new branch with the changes, commit and then push your new branch to kmlew/addons-vauxoo.
   4  Now you can create the pull request with your change to vauxoo/addons-vauxoo repository, just go to the pull request tab, and click in the create _Pull Request_ button.
   ![image](https://cloud.githubusercontent.com/assets/7593953/19790207/103629ee-9c7c-11e6-8097-1072f372f255.png)
4. Now you need to indicate where do you want to make the new add (the merge) and from what branch to add the new changes (your branch in kmlew/addons-vauxoo). In order to do this you need to first click in the link **compare across forks**
   ![image](https://cloud.githubusercontent.com/assets/7593953/19790227/45ff2256-9c7c-11e6-9425-315a9a7da642.png)
   
[...]

PR #1094, opened 2016-10-27
[FIX] error in view when trying to install mrp_production_analytic_acc module

zaoral:
This PR modify mrp_production_analytic_acc view in order to solve the issue
reported in #1093.


luisg123v:
The branch 9.0 is not used anymore, so closing.

PR #1100, opened 2016-11-01
[IMP] stock_card: when set warehouse in wizard we are searching

JulioSerna:
locations in base to parent left/right but when wizard is not set some
locations has not parent left/right and search return locations and try
to create stock card in that locations but we want to create stock card
with all moves when warehouse is not set in wizard

in order to solves:
**Asked Product has not Moves to show**


JulioSerna:
@hbto 

Could you check this change?

hbto:
+1 LGTM
@JulioSerna

PR #1101, opened 2016-11-03
[IMP] default_warehouse_from_sale_team: Modify search of warehouse from the move

deivislaya:
Verify that sale team is related to warehouse from move or picking type

### DIFF
--- a/default_warehouse_from_sale_team/models/stock.py
+++ b/default_warehouse_from_sale_team/models/stock.py
@@ -28,7 +28,11 @@ class StockQuant(models.Model):
             StockQuant, self)._get_accounting_data_for_valuation(move)
 
         sale_team = self.env['crm.case.section'].search(
-            [('default_warehouse', '=', move.warehouse_id.id)], limit=1)
+            [('default_warehouse',
+              '=',
+              move.warehouse_id.id if move.warehouse_id
+              else move.picking_type_id.warehouse_id.id)],
+            limit=1)
 
         if sale_team.journal_stock_id:
             journal_id = sale_team.journal_stock_id.id
--- a/default_warehouse_from_sale_team/views/stock_view.xml
+++ b/default_warehouse_from_sale_team/views/stock_view.xml
@@ -1,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<openerp>
+    <data>
+
+       <record id="stock_warehouse_sale_teams" model="ir.ui.view">
+            <field name="name">stock.warehouse.sale.teams</field>
+            <field name="model">stock.picking.type</field>
+            <field name="inherit_id" ref="stock.view_picking_type_form"/>
+            <field name="arch" type="xml">
+                <field name="warehouse_id" position="attributes">
+                    <attribute name="required">1</attribute>
+                </field>
+            </field>
+        </record>
+
+    </data>
+</openerp>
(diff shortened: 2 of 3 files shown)


default_warehouse_from_sale_team/models/stock.py
```diff
@@ -27,8 +27,12 @@ def _get_accounting_data_for_valuation(self, move):
         journal_id, acc_src, acc_dest, acc_valuation = super(
             StockQuant, self)._get_accounting_data_for_valuation(move)
 
+        warehouse_valuation = move.warehouse_id.id if move.warehouse_id else \
+            move.picking_type_id.warehouse_id.id if move.picking_type_id else \
+            False
+
         sale_team = self.env['crm.case.section'].search(
-            [('default_warehouse', '=', move.warehouse_id.id)], limit=1)
+            [('default_warehouse', '=', warehouse_valuation)], limit=1)
```
nhomar: you only need  move.warehouse_id remove the .id and is the same no previous validation needed.

default_warehouse_from_sale_team/views/stock_view.xml
```diff
@@ -0,0 +1,17 @@
...
+<openerp>
+    <data>
+
+       <record id="stock_warehouse_sale_teams" model="ir.ui.view">
+            <field name="name">stock.warehouse.sale.teams</field>
+            <field name="model">stock.picking.type</field>
+            <field name="inherit_id" ref="stock.view_picking_type_form"/>
+            <field name="arch" type="xml">
+                <field name="warehouse_id" position="attributes">
+                    <attribute name="required">1</attribute>
```
nhomar, line 11: I am not sure if this validation is necessary here, a picking type is not common to be edited manually.

default_warehouse_from_sale_team/models/stock.py
deivislaya: In this line we are validating, which sometimes is quant of stock.move not find a warehouse, if this case is true, we look for the warehouse in the picking type associated with stock.move. That is why we are applying a prior validation before search of crm.case.section

nhomar:
On Thu, Nov 3, 2016 at 10:47 AM, Deivis Laya notifications@github.com
wrote:

> In this line we are validating, which sometimes is quant of stock.move not
> find a warehouse, if this case is true, we look for the warehouse in the
> picking type associated with stock.move. That is why we are applying a
> prior validation before search of crm.case.section
> 
> I understand but the warehouse_valuation variable is not necessary, you
> can use directly the recordset in the domain if not setted this will be
> False no need to pass the .id (all this was not true in odl api).

## 

Nhomar Hernandez
CEO Vauxoo.
Site: http://vauxoo.com
Twitter: @nhomar
Blog: http://nhomar.com
Github User: https://github.com/nhomar
Odoo Gold Partner
Skype: nhomar00 (Envia mail previo no lo superviso siempre).
HangOut: nhomar@vauxoo.com
Móvil Venezuela:
+58 4144110269
Móvil México:
+52 1 4773933942

deivislaya:
@nhomar  

Remove the  `warehouse_valuation` variable and try to assign without .id, but fails to assert

nhomar:
which assert? can you show me the error please?

deivislaya:
stock operations odoo

deivislaya:
Hi @JulioSerna

This PR is ready for review

PR #1102, opened 2016-11-05
[IMP] default_warehouse_from_sale_team: adapt class to allow get

JulioSerna:
sequence in base to sale tema in models like:
* sale.order
* purchase.order
* purchase.requisition


JulioSerna:
[DUMMY PR](https://github.com/Vauxoo/typ/pull/124)

PR #1103, opened 2016-11-07
[9.0][ADD] new module currency_print_name

osvalr:
##  Summary
When needing to print invoices, POS tickets, sale orders or any printed document and also indicate the full name in your or your partner's language for the currency you're handling, this module includes that and makes it translatable getting up to 50 characters in the print name long enough for your needs.


osvalr:
Hi @josemoralesp would you help me to review this new module, your feedback is welcome

osvalr:
@nhomar your feedback is welcomed here

nhomar:
I think we need the es.po file.

osvalr:
@nhomar `es.po` is already there

PR #1108, opened 2016-11-08
[REF] workalendar_holidays: modify holiday import to allow a given tz

suniagajose:
Summary
-------------

In cases when not should import holidays using user timezone, now we can send by context a custom timezone.


suniagajose:
hi @moylop260 can merge this please??

thanks

PR #1109, opened 2016-11-08
[REF] addons-vauxoo: mute pylint notification by function duplicated

jorgescalona:
Add disable pylint message


JulioSerna:
I close pr by the name branch is not related 8.0-typ-landed-cost-mute-pylint-jorgescalon

PR #1110, opened 2016-11-08
[FIX] addons-vauxoo: add disable message by duplicate methods

jorgescalona:
disable continous integration message respect to duplicated methods by different apis


JulioSerna:
LGTM
👍

PR #1114, opened 2016-11-11
[IMP] Do not check all location only where products are available

hbto:
Use only locations where there are available quants to post
Journal Entries on Change of Standard Price on products


hbto:
@moylop260 @nhomar @josemoralesp @JulioSerna 

Could you review this PR.

PR #1115, opened 2016-11-11
Do not check all location only where products are available

hbto:
Use only locations where there are available quants to post
Journal Entries on Change of Standard Price on products

Related to https://github.com/Vauxoo/addons-vauxoo/pull/1114


hbto:
@moylop260 @nhomar @josemoralesp @JulioSerna

Could you review this PR.

moylop260 (approved):

hbto:
@nhomar could you review too?

PR #1116, opened 2016-11-12
[FIX] Fixing asked model is different from provide model at SQL

hbto:
At SQL Query whereabouts model used is product.template, so records are of that model
and not from product.product, hence the _new_ INNER JOIN clause


sabrinagnu:
👍

PR #1125, opened 2016-11-23
[FIX] Passing unexpected non boolean on boolean field

hbto:
[FIX] Passing unexpected non boolean value '0' (is string) 
in boolean field 'attachment_use'. You should use eval='False'.


hbto:
Please @moylop260 @nhomar 

merge this minor fix 

it is causing warnings in other projects.

Regards

PR #1126, opened 2016-11-24
[REF] product_available_by_warehouse: Show only qty available in internal locations

ysantiago:
Close [Vauxoo/typ#181](https://github.com/Vauxoo/typ/pull/181)
Close [Vauxoo/yoytec#2174](https://github.com/Vauxoo/yoytec/pull/2174)

### DIFF
--- a/product_available_by_warehouse/models/product.py
+++ b/product_available_by_warehouse/models/product.py
@@ -40,3 +40,11 @@ class ProductProduct(models.Model):
                 'outgoing_qty': outgoing_qty,
             }
         return res
+
+    @api.multi
+    def _get_domain_locations(self):
+        res = super(ProductProduct, self)._get_domain_locations()
+        if self._context.get('qty_available', False):
+            for domain in res:
+                domain.insert(0, ('location_id.usage', '=', 'internal'))
+        return res
--- a/product_available_by_warehouse/models/sale.py
+++ b/product_available_by_warehouse/models/sale.py
@@ -42,7 +42,9 @@ class SaleOrderLine(models.Model):
         product_qty_by_wh = []
         route_ids = []
         if colname in titles:
-            product_qty_by_wh = product.\
+            ctx = ((colname == 'qty_available') and
+                   {'qty_available': True} or {})
+            product_qty_by_wh = product.with_context(ctx).\
                 get_product_available_by_warehouse()[product.id][colname]
         for warehouse, product_qty in product_qty_by_wh:
             # TODO: uncomment until figure out a better way to choice only


product_available_by_warehouse/models/product.py
```diff
@@ -40,3 +40,11 @@ def get_product_available_by_warehouse(self):
                 'outgoing_qty': outgoing_qty,
             }
         return res
+
+    @api.multi
+    def _get_domain_locations(self):
```
JulioSerna, line 45: @ysantiago 

I do not know if is the correct make this super, it suppose when you call the product_available return quantity only of internal location and you do not need modify into native method in odoo like this,

@suniagajose 
what do you think?

suniagajose: Hi @JulioSerna 

I do not agree with the super, it is a risk change that can affect other features, maybe should do the computing independent of the context to avoid the change in a common use function like `_get_domain_locations`

But to be honest, actually i understand that qty_available only show product quantities on internal locations...  i don't know if i'm wrong, but will useful that show a case (with a video or screenshot) when qty available, sum from a not internal location

suniagajose: In help of `qty_available` field:

> In a context with a single Warehouse, this includes goods stored in the Stock Location of this Warehouse, or any of its children. stored in the Stock Location of the Warehouse of this Shop,  or any of its children Otherwise, this includes goods stored in any Stock Location  with 'internal' type.

ysantiago: Hello @suniagajose 

Field ```qty_available``` shows quantities of location that are not internal locations, how it can see [here](https://www.youtube.com/watch?v=6RxHck7KGtw&feature=youtu.be)

suniagajose:
hi @ysantiago 

With the video i see the problem, my opinion is that this PR should be made on client instance, not in addons-vauxoo because the locations for this case on video is not a properly design... In a properly design all locations child of warehouse location with 'view' type should be 'internal' type.

IMHO, the solution (if is possible) that i would propose is redesign the locations properly.

ysantiago:
@suniagajose 

Therefore, I don't think that help field ```qty_available```(which you quote above) apply to this locations configuration.

JulioSerna:
@ysantiago  @suniagajose 

I think in basis of the video this is absolutely  wrong, it is no depends of the customer configuration locations, You can not never consider like on hand product in locations of type different to 'internal'

josemoralesp:
We have three main of view locations(Odoo Base)

Physical Locations.

Virtual Locations.

Partner Locations.

A location of type inventory should not be child of physical location.

The compute of qty on hand uses all locations child of physical locations. It is conceptually wrong have an inventory location, child of our physical location.

Yes I could be agree with you and only consider the location of internal type but this bug comes from a conceptually wrong configuration of our locations and this could be fixed without waste more time in code.

We could have an similar error with our journal items and our accounts but we will not change the entire balance logic to specify account types.

I only want to express an idea @JulioSerna  I leave you the final decision of this topic.

Best Regards.

suniagajose:
Exactlly @josemoralesp, it is not a written rule: if location is a child of a physical location should be counted as on hand quantity, (basically all locations down this hierarchy should be internal). Otherwise, if technically is not a physical location, then should be on virtual location or partner location hierarchy

suniagajose:
Extracted from Odoo documentation:

> Here are the different available types of locations:
> 
> - Supplier Location: virtual location representing the source location for products received from suppliers,
> - View: shows that the location is only an organizational node for the hierarchical structure, and cannot be involved in stock moves itself. The view type is not made into a leaf node in a structure – it usually has children.
> - Internal Location: physical location inside your own stock,
> - Customer Location: virtual location representing the destination for products sent to customers,
> - Inventory: virtual location serving as the counterpart for inventory operations used to correct stock levels (physical inventories),
> - Procurement: virtual location serving as temporary counterpart for procurement operations when you do not yet know the source (supplier or production). Products in this location should be zero after the scheduler run completes,
> - Production: virtual counterpart location for production operations; consuming raw material and sending finished products,
> - Transit Location for Inter-Companies Transfers: used as an intermediate location in a multi-company environment.
>
> You can have several locations of the same type. In that case, your product, supplier and warehouse configurations determine the location that is to be used for any given operation.

[source](http://openerpbook.com/warehouse/5_14_Stock_inv.html#location)

suniagajose:
the following is extracted from official 9.0 documentation:

> There are 3 types of locations:
> 
> **The Physical Locations** are internal locations that are part of the warehouses for which you are the owner. They can be the loading and unloading area of your warehouse, a shelf or a department, etc.
> **The Partner Locations** are spaces within a customer and/or vendor's warehouse. They work the same way as Physical Locations with the only difference being that you are not the owner of the warehouse.
> **The Virtual Locations** are places that do not exist, but in which products can be placed when they are not physically in an inventory yet (or anymore). They come in handy when you want to place lost products out of your stock (in the Inventory loss), or when you want to take into account products that are on their way to your warehouse (Procurements).

[source](http://www.odoo.com/documentation/user/9.0/inventory/settings/warehouses/difference_warehouse_location.html)

ysantiago:
@josemoralesp @suniagajose thanks for your time. 

A redesign of the locations will be done.

PR #1127, opened 2016-11-25
[REF] workalendar_holidays: added section 'Usage' in README

suniagajose:
TODO
---------

- [x] Explained with images how use 'Import Holidays' wizard
- [x] Added example of `compute_working_days` function implementation
- [x] Added docstring to `compute_working_days` function
- [x] Indicate the reason why the timezone should be the same as the time of the system


suniagajose:
hi @nhomar this is ready to merge

PR #1129, opened 2016-11-27
[IMP] default_warehouse_from_sale_team: add priority to get warehouse

JulioSerna:
from warehouse of picking type of stock move because warehouse_id of
stock move can be another different by stock move propagated like
re-supply move from another warehouse


JulioSerna:
PR DUMMY https://github.com/Vauxoo/typ/pull/190

PR #1138, opened 2016-12-20
[FIX] Fixing stock_card because it was partially covering Odoo Purchase returns case

hbto:
[FIX] Fixing stock_card because it was partially covering
purchase returns when stock_available was equal to zero.

Partially because were replicating exactly the case exposed by Odoo
which was returning a quant which was already sold.
Now we take into account the quant available which lead to a qty=0

Case Exposed by Odoo: https://goo.gl/6vfvgz
https://www.odoo.com/documentation/user/9.0/accounting/others/inventory/avg_price_valuation.html#purchase-return-use-case

Case Exposed by Odoo show with Stock Card:
https://drive.google.com/file/d/0B1w9gokcJL9hUUVmNDNRLWhfMHM/view?usp=sharing

The case when returning Purchase left qty > 0:
https://drive.google.com/file/d/0B1w9gokcJL9hWlJsT0hrOFRqTTQ/view?usp=sharing

### DIFF
--- a/stock_card/model/stock_card.py
+++ b/stock_card/model/stock_card.py
@@ -138,7 +138,6 @@ class StockCardProduct(models.TransientModel):
         vals['product_qty'] += (vals['direction'] * row['product_qty'])
         sm_obj = self.env['stock.move']
         move_id = sm_obj.browse(row['move_id'])
-        product_id = self.env['product.product'].browse(row['product_id'])
         # Cost is the one record in the stock_move, cost in the
         # quant record includes other segmentation cost: landed_cost,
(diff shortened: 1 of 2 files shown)


hbto:
@nhomar 

could you review it.


This has gone green on Customers instance
https://travis-ci.com/Vauxoo/lodigroup/builds/36993376

JulioSerna:
👍

hbto:
@nhomar @moylop260 @JulioSerna 
Customer has agree on feature/fix.

Best Regards.

PR #1140, opened 2016-12-20
[IMP] Setting Border Date to Stock Card on Product

hbto:
Chances are that you are migrating data from one system to another or from one version to another, and you could end up in not trusting the Original Stock Card, because, there are returns from previous date than that when setup your Initial Inventory Values that could mess up your stock average.

Then what we intend is to provide you with a Tool where you could set Border Conditions for your stock card, i.e., you said from this moment on my Stock Card is legal, right, true reliable.

To achieve this goal there are some assumptions to be made, it will be considered a Universal True that you will be performing an Initialization of Inventory Values, like sending all the products to Inventory Loss to deplete all the products setting then to zero and then set your right average on your products and then Creating new Stock Move which will create new Inventory Stock on your Warehouse replenishing your Inventories, all this been done at a specific day.

Feature Exposed:
https://drive.google.com/file/d/0B1w9gokcJL9hRElwZEl2a2lRaXc/view

### DIFF
--- a/stock_card/model/product.py
+++ b/stock_card/model/product.py
@@ -1,10 +1,14 @@
 # -*- coding: utf-8 -*-
 
-from openerp import models, api
+import logging
+from openerp import models, api, fields
+_logger = logging.getLogger(__name__)
 
 
 class ProductProduct(models.Model):
     _inherit = 'product.product'
+
+    date_stock_card_border = fields.Datetime(string='Stock Card Border Date')
 
     @api.multi
     def stock_card_move_get(self):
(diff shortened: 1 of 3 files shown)


hbto:
@JulioSerna @nhomar please review

Regards

JulioSerna:
👍 

How we can fill the date_stock_card_border field?

hbto:
@JulioSerna 
> How we can fill the date_stock_card_border field?

https://drive.google.com/file/d/0B1w9gokcJL9hRElwZEl2a2lRaXc/view

PR #1141, opened 2016-12-22
[FIX] Allow other cost_method than standard to created Journal Entries

hbto:
[FIX] Allow other cost_method than standard to created Journal Entries
for price adjustment.
Add key in Journal Item to track which product was updated.


JulioSerna:
👍

product_extended_variants/model/template.py
```diff
@@ -117,7 +119,7 @@ def do_change_standard_price(self, cr, uid, ids, new_price, context=None):
 
                 for prod_variant in product.product_variant_ids:
                     qty = prod_variant.qty_available
-                    if prod_variant.cost_method != 'standard' or not qty:
+                    if not qty or make_journal_entry:
```
hbto: This condition should be like:
`if prod_variant.cost_method != 'standard' and not make_journal_entry or not qty:`

Otherwise it will be applied to all Products regardless of the Cost Method.

Best Regards

GavyMG: Ups! My fault. It's

stock_unfuck/models/res_config.py
```diff
@@ -0,0 +1,16 @@
...
+
+class StockConfiguration(models.TransientModel):
+
+    _inherit = 'stock.config.settings'
+
+    company_id = fields.Many2one(
+        'res.company', string='Company', required=True,
+        default=lambda self: self.env.user.company_id)
+
+    make_journal_entry = fields.Boolean(
```
JulioSerna: @hbto 

Are you agree in add this change in this module?

I thought that this configuration it would be into product_extended_variants module

hbto: @JulioSerna I Agree, 

What do you think @nhomar, Do you Agree?

JulioSerna: @GavyMG 

I talked with @hbto and we are agree in keep all this change into product_extended_variants
Could you make the change and move your changes from stock_unfuck to product_extended_variants?

GavyMG: Ok, no problem.

PR #1146, opened 2016-12-28
[ADD] stock_account_unfuck: return product supplier/customer to real value of transaction

ebirbe:
This code is a migration of code introduced in https://github.com/Vauxoo/typ/pull/228 applying the coding recommendations made in that PR.

### DIFF
--- a/stock_account_unfuck/models/stock.py
+++ b/stock_account_unfuck/models/stock.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import division
+
+from openerp import api, models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    @api.multi
+    def _recalculation_average_cost_price(self, signal=1):
+        """ move is a browse record """
+        for move in self:
+            if move.product_qty == 0:
+                return
+            for quant in move.reserved_quant_ids:
+                if quant.qty <= 0:
+                    return
+            average_valuation_price = 0.0
+            qty_available = move.product_id.product_tmpl_id.qty_available
+            average_valuation_price = sum(
+                [quant.qty * quant.cost for quant in move.reserved_quant_ids])
+            amount_unit = move.product_id.standard_price
+            new_std_price = (
+                (amount_unit * qty_available) +
+                average_valuation_price * signal) / (
+                    qty_available + move.product_qty * signal)
+            move.product_id.sudo().write({'standard_price': new_std_price})
+
+    @api.multi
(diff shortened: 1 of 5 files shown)


stock_account_unfuck/models/stock.py
```diff
@@ -0,0 +1,45 @@
...
+        for move in self:
+            if move.product_qty == 0:
+                return
+            for quant in move.reserved_quant_ids:
+                if quant.qty <= 0:
+                    return
+            average_valuation_price = 0.0
+            qty_available = move.product_id.product_tmpl_id.qty_available
+            average_valuation_price = sum(
+                [quant.qty * quant.cost for quant in move.reserved_quant_ids])
```
hbto, line 24: This is right when returning merchandise to Suppliers.
```
+            average_valuation_price = sum(
+                [quant.qty * quant.cost for quant in move.reserved_quant_ids])
```

When returning merchandise from `customer` or `transit`, the value to be used is the 
average price when the transaction was performed, i.e., `price_unit ` on
stock.move record.
https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L1811

hbto (changes requested):

stock_account_unfuck/models/stock.py
hbto: @JulioSerna Do you agree on my suggestions?

hbto:
UnitTests are to be added to this module for following cases:
- [ ] [Purchase Returns - Case Remaining Qty equals zero](https://github.com/vauxoo-dev/addons-vauxoo/blob/9a89032f1943f5d6fce3e00d84fc6949d211b3bb/stock_account_unfuck/README.md#odoos-approach-on-purchase-returns---case-remaining-qty-equals-zero)
- [ ] [Purchase Returns - Case Remaining Qty greater than zero](https://github.com/vauxoo-dev/addons-vauxoo/blob/9a89032f1943f5d6fce3e00d84fc6949d211b3bb/stock_account_unfuck/README.md#new-approach-on-purchase-returns---case-remaining-qty-greater-than-zero)
- [ ] [Sales Returns](https://github.com/vauxoo-dev/addons-vauxoo/blob/9a89032f1943f5d6fce3e00d84fc6949d211b3bb/stock_account_unfuck/README.md#new-approach-on-sales-returns)

hbto:
@JulioSerna the accounting fix is not yet scheduled for this branch/PR?

stock_account_unfuck/models/stock.py
JulioSerna: 👍

PR #1151, opened 2017-01-10
[FIX] [stock_landed_costs_average] Default values for debit & credit in aml records

hbto:
[FIX] Odoo sometimes could be touchy when not passing default values for 
debit and credit fields in account.move.line model


JulioSerna:
👍

PR #1153, opened 2017-01-12
[IMP] Take into account the date_stock_account_border to compute average

hbto:
Returns from Customers or Transit will be performed at current 
average instead of previous cost on transaction when initialization date
is set on the product

NOTE
====
stock.py: generates/computes the `average` and,
stock_account.py: takes a value (from stock.move or product, it depends) and book a journal entry.

UnitTest
========
- [x] Accounting - Journal Entries are to be make a Cost taking into account Historical Cost and Accounting Border Date.
- [ ] Logistics - Average is to be computed taking into account Historical Cost and Accounting Border Date.
- [ ] Provide a thorough UnitTest (average / journal items)- testing each step in the process of Receiving merchandise from Supplier, Sending it to Customer, Returning Merchandise to Supplier, Returning Merchandise from Customer, Sending/Receiving Merchandise to/from Transit.
- [ ] Provide UnitTest for border date.
- [ ] Review `price_unit` on `action_done` check if *before* or *after*

### DIFF
--- a/stock_account_unfuck/__openerp__.py
+++ b/stock_account_unfuck/__openerp__.py
@@ -21,6 +21,7 @@
     ],
     # always loaded
     'data': [
+        'views/view.xml',
     ],
     # only loaded in demonstration mode
     'demo': [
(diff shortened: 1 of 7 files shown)


JulioSerna:
👍

luisg123v:
@hbto,

Could we close this?

PR #1155, opened 2017-01-16
Thorough Unit Testing on Stock Account Unfuck

hbto:
Because of a better structured Unit Test on `stock_account_unfuck` a bug was fixed


hbto:
It went Green on Customer's instance merging

PR #1165, opened 2017-01-27
Set installable True the hr_payroll_multicompany and sync_youtube modules

zaoral:
[task#6803](https://www.vauxoo.com/web?debug=?#id=6803&view_type=form&model=project.task&menu_id=136&action=138)

This is required to run our instance vauxoo.


nhomar:
This 2 modules were out of scope at the very begining, why this merge?

PR #1166, opened 2017-01-27
[IMP] product_extended_variants: in order to allow propagate context

JulioSerna:
with location to others modules and make the super and change the accounting
values


hbto (approved):

JulioSerna:
PR dummy related https://github.com/Vauxoo/lodigroup/pull/654

PR #1170, opened 2017-02-01
[IMP] partner_credit_limit: Complete coverage

egrivero:
complete sale.py coverage
complete account_invoice.py coverage
complete partner.py coverage

### DIFF
--- a/partner_credit_limit/model/partner.py
+++ b/partner_credit_limit/model/partner.py
@@ -58,7 +58,8 @@ class ResPartner(models.Model):
             # credit = 0.0
             debit_maturity, credit_maturity = 0.0, 0.0
             for line in movelines:
-                if line.date_maturity and line.partner_id.grace_payment_days:
+                limit_day = line.date_maturity
+                if line.partner_id.grace_payment_days:
                     maturity = fields.Datetime.from_string(
                         line.date_maturity)
                     grace_payment_days = timedelta(
@@ -66,10 +67,6 @@ class ResPartner(models.Model):
                     limit_day = maturity + grace_payment_days
                     limit_day = limit_day.strftime("%Y-%m-%d")
 
-                elif line.date_maturity:
-                    limit_day = line.date_maturity
-                else:
-                    limit_day = fields.Date.today()
                 if limit_day <= fields.Date.today():
                     # credit and debit maturity sums all aml
                     # with late payments
--- a/partner_credit_limit/tests/test_sale_credit_limit.py
+++ b/partner_credit_limit/tests/test_sale_credit_limit.py
@@ -52,6 +52,15 @@ class TestSalesCreditLimits(common.TestCommon):
         with self.assertRaises(exceptions.Warning):
             sale_id.action_confirm()
 
+        # CASE WHERE PARTNER HAVE CREDIT
+        # set credit limit in 500
+        self.partner_china.credit_limit = 700.00
+        self.assertTrue(self.partner_china.allowed_sale,
+                        "Allowed Sale should be True")
+        sale_id.action_confirm()
+        self.assertEqual(sale_id.state, 'sale',
+                         'State of the sales order must be in sale')
+
     def test_partner_with_late_payments(self):
         """This test validate that the partner has not late payments
 
--- a/partner_credit_limit/tests/test_invoice_credit_limit.py
+++ b/partner_credit_limit/tests/test_invoice_credit_limit.py
@@ -50,7 +50,6 @@ class TestCreditLimits(common.TestCommon):
         # credit limit exceded
         # credit_limit = 500
         # amount_total = 600
-        print invoice_id.state
         with self.assertRaises(exceptions.Warning):
             invoice_id.action_invoice_open()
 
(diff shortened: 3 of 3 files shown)


partner_credit_limit/tests/test_sale_credit_limit.py
```diff
@@ -52,6 +52,13 @@ def test_credit_limit_overloaded(self):
         with self.assertRaises(exceptions.Warning):
             sale_id.action_confirm()
 
+        # CASE WHERE PARTNER HAVE CREDIT
+        # set credit limit in 500
+        self.partner_china.credit_limit = 700.00
+        self.assertTrue(self.partner_china.allowed_sale,
+                        "Allowed Sale, should be True")
```
nhomar: remove the comma.

egrivero:
I need to complete this, I do not know how to reproduce that

screenshot from 2017-01-31 21 22 32
screenshot from 2017-01-31 21 26 22

egrivero:
Achieve almost all coverage, minus line 72 from file partner.py
[partner_credit_limit](https://github.com/Vauxoo/addons-vauxoo/blob/10.0/partner_credit_limit/model/partner.py#L72)

But I do not think that validation is necessary, check the account code, where they declare the field data_madurity, and say it is required, and to give that condition that date_madurity is not full, I do not see it likely.
[account_move](https://github.com/Vauxoo/odoo/blob/10.0/addons/account/models/account_move.py#L392)

In the last commit I made where I put the test to do the coverage of lines 62 to 67, try to leave the date_madurity in balnco, but when posting the same account entry was filled in the field.
[commit](https://github.com/Vauxoo/addons-vauxoo/pull/1170/files#diff-9829129cc70ab53b122f22ae06f55f67R176)

I propose to remove that condition

nhomar:
On Tue, Jan 31, 2017 at 10:25 PM, Edgar Rivero [Vauxoo] <
notifications@github.com> wrote:

> Achieve almost all coverage, minus line 72 from file partner.py
> partner_credit_limit
> <https://github.com/Vauxoo/addons-vauxoo/blob/10.0/partner_credit_limit/model/partner.py#L72>
>
> But I do not think that validation is necessary, check the account code,
> where they declare the field data_madurity, and say it is required, and to
> give that condition that date_madurity is not full, I do not see it likely.
> account_move
> <https://github.com/Vauxoo/odoo/blob/10.0/addons/account/models/account_move.py#L392>
>
> In the last commit I made where I put the test to do the coverage of lines
> 62 to 67, try to leave the date_madurity in balnco, but when posting the
> same account entry was filled in the field.
> commit
> <https://github.com/Vauxoo/addons-vauxoo/pull/1170/files#diff-9829129cc70ab53b122f22ae06f55f67R176>
>
> I propose to remove that condition
>

remove it and explain the ratilonale in the commit.

Nhomar Hernandez
CEO Vauxoo.
Site: http://vauxoo.com
Twitter: @nhomar
Blog: http://nhomar.com
Github User: https://github.com/nhomar
Odoo Gold Partner
Skype: nhomar00 (Envia mail previo no lo superviso siempre).
HangOut: nhomar@vauxoo.com
Móvil Venezuela:
+58 4144110269
Móvil México:
+52 1 4773933942

moylop260:
What is your dummy pr to check the travis testing this module?

egrivero:
> What is your dummy pr to check the travis testing this module?

vauxoo/imeqmo#61

moylop260:
Remove the else and initialize the variable `limit_day = fields.Date.today()` before of the first `if`
In order to reduce cyclomatic complexity rate and qty of tests required to cover it

egrivero:
Uhm, I had thought of it, in the following way, since it is certain that the field date_madurity
```
limit_day = line.date_maturity
if line.date_maturity and line.partner_id.grace_payment_days:
    maturity = fields.Datetime.from_string(line.date_maturity)   
    grace_payment_days = timedelta(
        days=line.partner_id.grace_payment_days)
    limit_day = maturity + grace_payment_days
    limit_day = limit_day.strftime("%Y-%m-%d")

**Comment by moylop260** [comment]
Oh I got it is a required field.
Thanks for clarify

**Comment by josemoralesp** [comment]
If you want, you can use one line for the for limit_day format
```python
limit_day = (maturity + grace_payment_days).strftime("%Y-%m-%d")
```

egrivero:
I do not know if it is negative with twice the same variable, but with the change that is proposed I would stay in two lines not to exceed 80 characters

```
limit_day = (
    maturity + grace_payment_days).strftime("%Y-%m-%d")

**Comment by josemoralesp** [comment]
I don't know the space available you have to use, all text has approximately 60(y pico) characters. If you don't have enough space you can keep the old line.

That can be used in your future codes.

**Comment by egrivero** [comment]
Sure thanks

**Comment by egrivero** [comment]
@nhomar @moylop260 

This PR is ready for Merge

PR Dummy vauxoo/imeqmo#61

File: partner_credit_limit/model/partner.py [hunk 2]
```diff
@@ -58,6 +58,7 @@ def _get_overdue_credit(self):
             # credit = 0.0
             debit_maturity, credit_maturity = 0.0, 0.0
             for line in movelines:
+                limit_day = line.date_maturity
                 if line.date_maturity and line.partner_id.grace_payment_days:
```
moylop260: Could you remove the `if` where the field `date_maturity` is used to check if exists?

egrivero: I changed it, just leave the condition of
If line.partner_id.grace_payment_days

partner_credit_limit/model/partner.py
```diff
@@ -58,18 +58,15 @@ def _get_overdue_credit(self):
             # credit = 0.0
             debit_maturity, credit_maturity = 0.0, 0.0
             for line in movelines:
-                if line.date_maturity and line.partner_id.grace_payment_days:
+                limit_day = line.date_maturity
```
moylop260, line 61: Maybe you could delete other `if` using other logic way and reduce the mccabe rate.
```python
limit_day = fields.Datetime.from_string(line.date_maturity) + timedelta(days=line.partner_id.grace_payment_days or 0)
```

partner_credit_limit/model/partner.py
```diff
@@ -58,18 +58,15 @@ def _get_overdue_credit(self):
...
+                if line.partner_id.grace_payment_days:
                     maturity = fields.Datetime.from_string(
                         line.date_maturity)
                     grace_payment_days = timedelta(
                         days=line.partner_id.grace_payment_days)
                     limit_day = maturity + grace_payment_days
                     limit_day = limit_day.strftime("%Y-%m-%d")
```
moylop260, line 68: We should use the odoo global variable for this format
Or you could use datetime object directly to avoid format issues.
Something like as:
```python
# Using "limit_day" like as datetime object (not string)
`if (fields.Datetime.from_string(fields.Date.today() - limit_day).days <= 0`

nhomar:
Everybody:

I will merge this because the scope was covered, then @egrivero will make the fixes and little refactors in other PR.

PR #1172, opened 2017-02-02
[IMP] payment_term_type: Added test to verify when changing the configuration of the field payment_term_type and complete coverage

egrivero:
payment_term_type: Coverage Complete


egrivero:
@nhomar This PR is ready for review

PR Dummy: vauxoo/imeqmo#70

PR #1175, opened 2017-02-07
[ADD] product_cost_usd: handle cost in USD on products

suniagajose:
The approach is to have the USD as the base currency, so that the sale prices in company currency can be kept up to date through the exchange rate.

Features:
------------
- New Cost USD field on the Product form.
- Validate Cost USD so that it is not less than the
  list price of the supplier.
- Avoid save a 'Cost in USD' when product does not has
  assigned a supplier with price in USD.
- Allowed pricelist computation based on Cost in USD.
- Added unit tests to validate constrains and pricelists
  computation.
- Compatibility with module sale_margin.


suniagajose:
@nhomar 

this is ready to merge

PR #1176, opened 2017-02-10
[REF] product_cost_usd: allow multi-level complexity

suniagajose:
Summary
-------------

Before this feature, the 'Cost in USD' not work properly when pricelist is based on another pricelist (based on 'Cost in USD'). 
The complexity of two-levels of pricelists was not possible. This feature only worked when the base cost in dollars is in the first level.

Example:
------------
1. Create a Pricelist: Minimal 5 MXN, to give a 10% discount when quantity at least 5
img

2. Create a Pricelist: Public Pricelist 15%, where a product has 10% discount on cost when quantity at least 5.
img2


luisg123v:

PR #1180, opened 2017-02-18
[IMP] account_voucher_tax: in order to avoid error by expected singleton

JulioSerna:
we need to validate if the journal entry has several journal items like
liquidity


JulioSerna:
PR dummy https://github.com/Vauxoo/typ/pull/298

PR #1181, opened 2017-02-24
[IMP] Optimization of process to get stock_card values

josemoralesp:


hbto:
Could create a PR for two outstanding customer instances in order to check that 
we are not breaking anything

hbto (approved):

hbto (changes requested):
There are some changes to apply yet, I have seen some actual data that differs, does not match, in two different instances

hbto (approved):
I have check and the cause of the mismatching values is not in this PR, but in another PR the one in TyP, so, there was changed the code to fix the problem.

Go ahead with merging

stock_card/model/stock_card.py
```diff
@@ -416,6 +417,10 @@ def _stock_card_move_history_get(self, product_id, locations_ids=None):
         query = '''
             SELECT distinct
                 sm.id AS move_id, sm.date, sm.product_id, prod.product_tmpl_id,
+                sm.origin_returned_move_id AS origin_returned_move_id,
+                sm.move_dest_id AS move_dest_id,
+                sm.move_dest_id AS move_dest_id,
```
hbto: Delete This line is written twice

josemoralesp:
PR: 
TYP: https://github.com/Vauxoo/typ/pull/306
LODI: https://github.com/Vauxoo/lodigroup/pull/658

hbto:
as both are ok I will endorse this PR
+1

JulioSerna (approved):

JulioSerna:
👍

PR #1185, opened 2017-03-09
[IMP] account_anglo_saxon_stock_move: Added journal and period

luistorresm:
Added journal and period when try reconcile the stock accrual, to cases
when reconcile have offset.

Added validation that journal is set in product category


luistorresm:
@JulioSerna @hbto 

Could you merge this PR please?

Regards

hbto:
👍

PR #1189, opened 2017-03-27
[FIX] sale_margin_percentage: migrating to new sale.config.settings view

suniagajose:
Summary
-------------

The `sale.config.settings` view was changed completely after v10.0, in this fix we have moved `margin_threshold` field from old sale settings view to the new one.


suniagajose:
hi @nhomar 

Can you merge this please?

Thanks

PR #1192, opened 2017-05-09
[FIX] stock_account_unfuck: Use price_unit of move_orig_ids field to calculate standard_price of product

ysantiago:
Use price_unit of move_orig_ids field insteed of price_unit of move_dest_id field to calculate standard_price of product.

Close [Vauxoo/lodigroup#694](https://github.com/Vauxoo/lodigroup/pull/694)


hbto:
👍 

Be aware: Returns are not covered with this approach.

We have to try to deal with this issue in a later development though it seems it is low priority

Best Regards

PR #1197, opened 2017-06-05
[IMP] account_move_line_base_tax: Removed amount validation to diot

luistorresm:
movements

In this module was added a validation to verify that movements that will
be considered in DIOT report must have amount in the credit. But to invoice
refunds the logic change, and the validation is incorrect.

The DIOT report already have the validation to movements with amount 0,
if this are created manually. By this reason is removed the validation
here.

### DIFF
--- a/account_move_line_base_tax/account.py
+++ b/account_move_line_base_tax/account.py
@@ -24,8 +24,6 @@
 #
 ##############################################################################
 from openerp.osv import osv, fields
-from openerp.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


luistorresm:
@JulioSerna 

Are you agree with this PR?

Fix this Issue https://drive.google.com/file/d/0BzSA94_IwlAUMGlWaDE5TzZnMDQ/view

Regards

JulioSerna:
👍

PR #1198, opened 2017-06-07
[FIX] stock_landed_costs_average: Rounding values of multiple valuation adjustments lines in landed cost for positive and negative values of cost lines

ysantiago:


hbto:
👍

stock_landed_costs_average/model/stock_landed_costs.py
```diff
@@ -587,7 +588,13 @@ def compute_landed_cost(self, cr, uid, ids, context=None):
                         else:
                             value = (line.price_unit / total_line)
 
-                        if digits:
+                        # When the valuation is the last record, assign
+                        # line.price_unit less the accumulated value in
+                        # value_split
+                        if digits and \
```
JulioSerna: you can use
 ```pyhton
if (digits and 
    valuation == cost_adjust_lines[-1]):
```

File: stock_landed_costs_average/model/stock_landed_costs.py [hunk 2]
```diff
@@ -587,7 +588,13 @@ def compute_landed_cost(self, cr, uid, ids, context=None):
...
+                                valuation == cost_adjust_lines[-1]:
```
JulioSerna: I do not understand this `cost_adjust_lines` is using by `for` but in this validation
`cost_adjust_lines[-1]` always is returning the last position , why?

stock_landed_costs_average/model/stock_landed_costs.py
ysantiago: @JulioSerna Change made

stock_landed_costs_average/model/stock_landed_costs.py
ysantiago: @JulioSerna because we have to know if ```valuation``` is the last register in ```cost_adjust_lines``` to assigned the remainder of ```line.price_unit```.

hbto: @JulioSerna This way you make sure that the remaining amount and the all previous amount that have already been allocated will sum as line.price_unit.

for example:

if you have line.price_unit = 100.

and there are three lines when divided: 100 / 3 = 33.33

the last one will have to be 33.34, 

using this approach will ensure this will happen.

Regards

JulioSerna: Thanks

hbto:
This issue is somehow related to:
https://github.com/odoo/odoo/pull/8716
https://github.com/odoo/odoo/pull/8714

and so this fixing should be carried on to upstream branch.

Regards.

JulioSerna (approved):

PR #1199, opened 2017-06-08
[REF] stock_landed_costs_average: Refactored code in function compute_landed_cost

ysantiago:


stock_landed_costs_average/model/stock_landed_costs.py
```diff
@@ -566,45 +566,46 @@ def compute_landed_cost(self, cr, uid, ids, context=None):
             for line in cost.cost_lines:
                 value_split = 0.0
                 cost_adjust_lines = cost.valuation_adjustment_lines
+                fnc = min if line.price_unit > 0 else max
                 for valuation in cost_adjust_lines:
+                    if (not valuation.cost_line_id or
+                            valuation.cost_line_id.id != line.id):
+                        continue
```
josemoralesp: @ysantiago What do you think about add this condition in the above loop, using filtered method?
I have measured this kind of situation and it's better use filtered method instead of an if statement within the loop.

ysantiago: @josemoralesp Do you refered something like that [33c4bed](https://github.com/Vauxoo/addons-vauxoo/pull/1199/commits/33c4bed2ddb3fc3fc1a8237daf310332b492c52a)?

josemoralesp: Yes, that works too

ysantiago:
@josemoralesp @hbto Could you please approve this change?

hbto:
👍

PR #1202, opened 2017-06-15
[FIX] stock_account_unfuck: Use current standard_price to calculate the new value of standard_price in move from transit to stock when there is no origin in the picking

ysantiago:


hbto:
:+1:

PR #1204, opened 2017-06-16
8.0 controller report xls performance imp hbto

hbto:
Memoizing frequently asked method.

`ormcache`  as it is only applicable to orm methods
as stated [here](https://github.com/odoo/odoo/blob/10.0/odoo/tools/cache.py#L34)

### DIFF
--- a/aging_due_report/wizard/wizard.py
+++ b/aging_due_report/wizard/wizard.py
@@ -41,13 +41,6 @@ class AccountAgingWizardDocument(models.TransientModel):
     _rec_name = 'partner_id'
     _order = 'date_due'
 
-    @api.depends('date_due')
-    def _get_due_days(self):
-        for record in self:
-            today = datetime.now()
-            date_due = datetime.strptime(record.date_due, '%Y-%m-%d')
-            record.due_days = (today - date_due).days
-
     partner_id = fields.Many2one('res.partner', 'Partner')
     invoice_id = fields.Many2one('account.invoice', 'Invoice')
     aml_id = fields.Many2one('account.move.line',
@@ -58,8 +51,7 @@ class AccountAgingWizardDocument(models.TransientModel):
     tax = fields.Float('Tax')
     total = fields.Float('Total')
     payment = fields.Float('Payment')
-    due_days = fields.Integer(string='Due Days',
-                              compute="_get_due_days")
+    due_days = fields.Integer(string='Due Days')
     date_emission = fields.Date('Emission Date')
     date_due = fields.Date('Due Date')
     company_id = fields.Many2one('res.company', u'Company')
@@ -109,31 +101,6 @@ class AccountAgingWizardPartner(models.TransientModel):
             if lt_lim and gt_lim:
                 res_line[span] += doc.residual
 
-    @api.depends('document_ids', 'document_ids.residual',
-                 'document_ids.payment', 'document_ids.total')
-    def _get_amount(self):
-        field_sum = ['residual', 'payment', 'total', 'not_due']
-        field_spans = ['span01', 'span02', 'span03', 'span04', 'span05']
-        field_names = field_spans + field_sum
-        for record in self:
-            res = {}
-            direction = record.aaw_id.direction == 'past'
-            spans = [record.aaw_id.period_length * x * (direction and 1 or -1)
-                     for x in range(5)]
-            for fn in field_names:
-                res[fn] = 0.0
-            for doc in record.document_ids:
-                for fsum in field_sum:
-                    if fsum != 'not_due':
-                        res[fsum] += doc.read([fsum])[0][fsum]
-                if not direction and doc.due_days > 0 \
-                        or direction and doc.due_days <= 0:
-                    res['not_due'] += doc.residual
-                else:
-                    record._get_amount_span(
-                        field_spans, res, doc, spans, direction)
-            record.update(res)
-
     partner_id = fields.Many2one('res.partner', string='Partner',
                                  required=True)
     name = fields.Char(string="Name", related='partner_id.name',
@@ -149,32 +116,23 @@ class AccountAgingWizardPartner(models.TransientModel):
                                help='Journal Items')
     currency_id = fields.Many2one('res.currency', 'Currency',
                                   required=True)
-    total = fields.Float(string='Total', digits=dp.get_precision('Account'),
-                         compute='_get_amount', store=True)
+    total = fields.Float(string='Total', digits=dp.get_precision('Account'))
     payment = fields.Float(string='Payment',
-                           digits=dp.get_precision('Account'),
-                           compute='_get_amount', store=True)
+                           digits=dp.get_precision('Account'))
     residual = fields.Float(string='Residual',
-                            digits=dp.get_precision('Account'),
-                            compute='_get_amount', store=True)
+                            digits=dp.get_precision('Account'))
     not_due = fields.Float(string='Not Due',
-                           digits=dp.get_precision('Account'),
-                           compute='_get_amount', store=True)
+                           digits=dp.get_precision('Account'))
     span01 = fields.Float(string='span01',
-                          digits=dp.get_precision('Account'),
-                          compute='_get_amount', store=True)
+                          digits=dp.get_precision('Account'))
     span02 = fields.Float(string='span02',
-                          digits=dp.get_precision('Account'),
-                          compute='_get_amount', store=True)
+                          digits=dp.get_precision('Account'))
     span03 = fields.Float(string='span03',
-                          digits=dp.get_precision('Account'),
-                          compute='_get_amount', store=True)
+                          digits=dp.get_precision('Account'))
     span04 = fields.Float(string='span04',
-                          digits=dp.get_precision('Account'),
-                          compute='_get_amount', store=True)
+                          digits=dp.get_precision('Account'))
     span05 = fields.Float(string='span05',
-                          digits=dp.get_precision('Account'),
-                          compute='_get_amount', store=True)
+                          digits=dp.get_precision('Account'))
     aawc_id = fields.Many2one('account.aging.wizard.currency',
                               'Account Aging Wizard Currency',
                               help='Account Aging Wizard Currency Holder')
@@ -185,30 +143,6 @@ class AccountAgingWizardCurrency(models.TransientModel):
     _description = 'Account Aging Wizard Currency'
     _rec_name = 'currency_id'
 
-    @api.depends('partner_ids', 'partner_ids.residual',
-                 'partner_ids.not_due', 'partner_ids.span01',
-                 'partner_ids.span02', 'partner_ids.span03',
-                 'partner_ids.span04', 'partner_ids.span05')
-    def _get_amount(self):
-        for record in self:
-            residual = not_due = span01 = span02 = 0.00
-            span03 = span04 = span05 = 0.00
-            for part in record.partner_ids:
-                residual += part.residual
-                not_due += part.not_due
-                span01 += part.span01
-                span02 += part.span02
-                span03 += part.span03
-                span04 += part.span04
-                span05 += part.span05
-            record.residual = residual
-            record.not_due = not_due
-            record.span01 = span01
-            record.span02 = span02
-            record.span03 = span03
-            record.span04 = span04
-            record.span05 = span05
-
     currency_id = fields.Many2one('res.currency', 'Currency',
                                   required=True)
     aaw_id = fields.Many2one('account.aging.wizard',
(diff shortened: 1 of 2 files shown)


moylop260:
What about use lru cache of python?
like as https://github.com/Vauxoo/cfdilib/pull/49

hbto:
@moylop260 Yes we have got rid of the previous implementation
and we have deployed the standard implementation of python.

> What about use lru cache of python?
> like as Vauxoo/cfdilib#49

Regards

aging_due_report/wizard/wizard.py
```diff
@@ -185,10 +184,7 @@ class AccountAgingWizardCurrency(models.TransientModel):
     _description = 'Account Aging Wizard Currency'
     _rec_name = 'currency_id'
 
-    @api.depends('partner_ids', 'partner_ids.residual',
-                 'partner_ids.not_due', 'partner_ids.span01',
-                 'partner_ids.span02', 'partner_ids.span03',
-                 'partner_ids.span04', 'partner_ids.span05')
+    @api.depends('aaw_id.user_id')  # /!\ NOTE: avoid multi-triggering
     def _get_amount(self):
```
moylop260: This is a global `get_amount` vs `get_amount` by line above.
And both are using the same operations but one is cumulative and other one is just by line.
What about use just `line._get_amount` and assign a cumulative sum to parent.

Something like as:
```python
_name = 'account.aging.wizard.partner'

def _get_amount(self):
    line['span01'] = 10
    line.parent_id['span01'] += line['span01']
```

With this feature you won't need to compute the same value 2 times to totalize and to get the line value.

moylop260:
The method `get_odoo_style` if you use a hashable objects (like as string) in the parameter received  of the method you could cache it too and the method should tranform this string to html object and could be faster. (Slow the first time but faster the second ones)

moylop260:
The method `_get_due_days` could be improvement with the following patch:
 
```diff
+       today = datetime.now()
        for record in self:
-            today = datetime.now()
             date_due = datetime.strptime(record.date_due, '%Y-%m-%d')
             record.due_days = (today - date_due).days
```

moylop260:
The method [_get_invoice_by_partner](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L477) looks like as:
```python
invoices = set(inv.search(...))
```

You don't need a set for a search if you want get unique records because sql return unique record for `SELECT ... FROM ... WHERE ...`

moylop260:
For the method [_get_invoice_by_partner](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L489-L491) you are using `[aml.id for aml in invoice.payment_ids]` but you could use just `invoice.payment_ids.ids`

moylop260:
For the method [_get_invoice_by_partner](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L497-L502) you are using to get `date_due`:

```python
            date_due = [amx_brw.date_maturity
                        for amx_brw in invoice.move_id.line_id
                        if amx_brw.account_id.type in (
                            'receivable', 'payable')]
            date_due = date_due and min(date_due)
            date_due = date_due or invoice.date_due or invoice.date_invoice
```

I  mean, you get all date_maturity of all records to use just one (the minimum one).
This consume too much resources.

You should use search with limit and order to faster date_due process.

```python
           date_maturity = aml.search_read(
                [('account_id.type', 'in', ['receivable', 'payable']),
                 ('date_maturity', '!=', None),
                 ('move_id', '=', invoice.move_id.id)], ['date_maturity'],
                limit=1, order='date_maturity ASC')
            date_due = (date_maturity[0]['date_maturity'] if date_maturity
                        else invoice.date_due)  or invoice.date_invoice
```

This will be faster because you don't have many small queries to get record by record the value.

moylop260:
You have a [global get_amount](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L114) plus a [get_amount by line](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L192)

Both are using the same operations but one is cumulative and other one is just by line.
What about use just `line._get_amount` and assign a cumulative sum to parent?

Something like as:
```python
_name = 'account.aging.wizard.partner'

def _get_amount(self):
    line.span01 = 10
    line.parent_id.span01 += line.span01
```

With this feature you won't need to compute the same value 2 times to totalize

moylop260:
For the following logic:

```python
values = [line.value for line in record.lines]
count = 0
for value in values:
   count += value
```

You should use `read_group` like as @josemoralesp recommend us.
I mean,
```python
count = record.read_group([('id', 'in', record.lines.ids)], ['value'], [])  # Use sum(value) of sql
count = count[0]['value'] if count else 0
```

By each `line.value` you are executing a small query too many times to database and this is slower.
Better use a big query with the result.

E.g. with this logic:
 - [_get_amount](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L123-L131)
 - [_get_aml](https://github.com/Vauxoo/addons-vauxoo/blob/27d831f449ef08f29722534a91cf68c1e2560458/aging_due_report/wizard/wizard.py#L452-L463)
  (I don't know if there are more similar methods but you could search them)

moylop260:
FYI I have created the following patches using `read_group` in order to have a better practical example of use them.

https://gist.github.com/moylop260/9e0211ee02cd440101d733f0e7f337c3

Please, test and fix before to use them directly

hbto:
> The method get_odoo_style if you use a hashable objects (like as string) in the parameter received of the method you could cache it too and the method should tranform this string to html object and could be faster. (Slow the first time but faster the second ones)

@moylop260 uno de los parámetros no es hashable

hbto:
> The method _get_due_days could be improvement with the following patch:
> 
> +       today = datetime.now()
>         for record in self:
> -            today = datetime.now()
>              date_due = datetime.strptime(record.date_due, '%Y-%m-%d')
>              record.due_days = (today - date_due).days


@moylop260 Taking into account

hbto:
@moylop260 Thanks for you suggestions they will be taken into account.

Thanks so much.

sabrinagnu:
@moylop260 technically this LGTY?, the functional tests 👍

josemoralesp:
Hello @hbto @moylop260 .

These would be my changes for _get_amount method using read_group:

```python
from collections import defaultdict
@api.depends('document_ids', 'document_ids.residual',
                'document_ids.payment', 'document_ids.total')
def _get_amount(self):
    field_sum = ['payment', 'total', 'aawp_id']
    field_spans = ['span01', 'span02', 'span03', 'span04', 'span05']
    for record in self:
        res = defaultdict(float)
        direction = record.aaw_id.direction == 'past'
        spans = [record.aaw_id.period_length * x * (direction and 1 or -1)
                    for x in range(5)]
        vals = record.document_ids.read_group(
            [('aawp_id', '=', record.id)], field_sum, ['aawp_id'])

        vals = vals[0] if vals else {}
        
        res['payment'] += vals[0].get('payment', 0)
        res['total'] += vals[0].get('total', 0)

        vals = record.document_ids.read_group(
            [('aawp_id', '=', record.id),
             ('due_days', '<=' if direction else '>', 0)],
            ['residual', 'aawp_id'], ['aawp_id'])
        
        if vals:
            res['not_due'] += vals[0].get('residual', 0)
        else:
            record._get_amount_span(
                field_spans, res, doc, spans, direction)
        record.update(res)
```

I have used two new things here:

This method defaultdict allows set an object by default, then when you try to get an unexisting key, this return the object. This is useful to avoid using extra loops in our code to fill values in a dict.

Another method used was read_group. This method is useful when you want to get sum of numeric values in many records(The same behavior of use a sql with "sum" and "group by"). In our case we use it to get the sum of three fields in the whole documents related with the wizard.

In sql this would be something like this:

```sql
SELECT
   aawp_id,
   sum(residual),
   sum(payment)
FROM
   account_aging_wizard_document
WHERE
   aawp_id = our_current_id
GROUP BY
   aawp_id
```

I didn't test the python method above. But it's only an idea(how it could be done).
Let me know if you need my help for something. I want to apologize if I said something extremely obvious for you, but I would want to leave the message to be reused in other cases for anyone in the team

aging_due_report/wizard/wizard.py
```diff
@@ -487,19 +484,21 @@ def _get_invoice_by_partner(self, partner_ids,
                            invoice.company_id.currency_id.id and
                            invoice.currency_id.id)
             payment = self._get_aml(
-                [aml.id for aml in invoice.payment_ids],
+                [invoice.payment_ids.ids],
```
moylop260: Remove the `[` from `[invoice.payment_ids.ids]` because .ids returns a list

moylop260:
@hbto 
I found the following OCA module to create xls reports:
 - https://github.com/OCA/reporting-engine/tree/8.0/report_xls

You can see a implementation from:
 - https://github.com/OCA/account-financial-reporting/blob/8.0/account_journal_report_xls/report/nov_account_journal_xls.py

Maybe this module could help us to improve performance.

hbto:
> I found the following OCA module to create xls reports

@moylop260  our way of implementing a report in XLS is less cumbersome than theirs.

We don't have to deal with fields nor definitions of any kind whatsoever.

Actually, `controller_report_xls` is cleaner to deploy 

Did you see how many lines with `xlwt` they had to write to implement that module?
It could be that one glance is not enough to judge an amazing judge.

Meanwhile using `controller_report_xls` is less intrusive, you have to declare one context
and the report will come in xls, nothing else, (behind the job you have to do to get it done).

Regards.

moy:
I guess you meant @moylop260, not @moy ;-).

aging_due_report/wizard/wizard.py
```diff
@@ -585,9 +511,61 @@ def compute_lines(self, partner_ids):
...
             line['aawp_id'] = aawp_ids[partner_id, currency_id]
             line['aaw_id'] = wzd_brw.id
+            line['due_days'] = self._get_due_days(today, line)
             aawd_obj.create(line)
+
+        # Only to trigger _get_amount
+        wzd_brw._get_amount()
         return True
 
+    def _get_amount(self):
```
hbto, line 524: @moylop260 I have taken away computed fields and consolidated `_get_amount` method into only one.

Though I would like to have implemented when creating records, I have not being able to do it without 
modifying the _get_amount_span().

Could you please further advise on this PR and give us your +1 if apply.

Regards.

aging_due_report/wizard/wizard.py
```diff
@@ -443,24 +370,18 @@ def _get_lines_by_partner_without_invoice(self, rp_ids):
...
+                for aml_brw in aml_ids)
         else:
-            aml_gen = (
-                aml_brw.debit and (aml_brw.debit * sign) or
-                aml_brw.credit * (-1 * sign)
-                for aml_brw in aml_obj.browse(aml_ids))
-            for iii in aml_gen:
-                res += iii
+            res = sum(
+                (aml_brw.debit * sign) or aml_brw.credit * (-1 * sign)
```
moylop260: What about use read_group here?

```python
    if currency_id:
            amount = aml_obj.read_group(
                [('id', 'in', aml_ids), ('amount_currency', '!=', 0)],
                ['amount_currency'], [])
            amount = amount[0] if amount else {}
            res = amount.get('amount_currency', 0) * sign
    else:
            amount = aml_obj.read_group(
                [('id', 'in', aml_ids)], ['debit', 'credit'], [])
            amount = amount[0] if amount else {}
            res = (amount.get('debit', 0) * sign +
                   amount.get('credit', 0) * sign * -1)
```

Note:
You can use directly a sum `aml.debit*sign + aml.credit*sign*-1 ` because production data don't have values in both fields at same time.

You can verify it using:
`SELECT debit, credit FROM account_move_line WHERE debit!=0 and credit!=0 LIMIT 1`
(same case for null values)

aging_due_report/wizard/wizard.py
```diff
@@ -585,9 +511,61 @@ def compute_lines(self, partner_ids):
...
+
+        for wzd_cur in self.currency_ids:
+            residual = not_due = span01 = span02 = span03 = span04 = 0.0
+            span05 = 0.0
+            for part in wzd_cur.partner_ids:
+                res = {}.fromkeys(field_names, 0.0)
+                for doc in part.document_ids:
+                    for fsum in field_sum:
+                        if fsum != 'not_due':
+                            res[fsum] += doc.read([fsum])[0][fsum]
```
moylop260: We have too many small queries here yet.

What about use the following patch with `read_group`?
https://gist.github.com/moylop260/9e0211ee02cd440101d733f0e7f337c3#file-read_group-patch

aging_due_report/wizard/wizard.py
```diff
@@ -585,9 +511,61 @@ def compute_lines(self, partner_ids):
...
+        field_sum = ['residual', 'payment', 'total', 'not_due']
+        field_spans = ['span01', 'span02', 'span03', 'span04', 'span05']
+        field_names = field_spans + field_sum
+        direction = self.direction == 'past'
+        spans = [
+            self.period_length * x * (direction and 1 or -1)
+            for x in range(5)]
+
+        for wzd_cur in self.currency_ids:
+            residual = not_due = span01 = span02 = span03 = span04 = 0.0
```
moylop260: You can use directly `_cache` in order to avoid define these local variables
`wiz_cur._cache['residual'] = 0` <- This avoid a write for database
`wiz_cur.residual` <- This avoid a read for database, read directly from the cache

This is similar to https://github.com/odoo/odoo/pull/14005/files#diff-0efb3242408405e7b60f00cdb80f83cfR25

moylop260:
@hbto Please, check my last commit

moylop260 (approved):

sabrinagnu:
@josemoralesp merge please as soon as the CI ends, @moylop260 approves this already, the functional test say that is taking only 2 minutes, great job team @hbto

josemoralesp:
Roger that

PR #1205, opened 2017-06-19
[REF] supplier_invoice_number_unique: Use unique sql constraint instead.

moylop260:
For a customer instance we have the following profiling results validating a invoice:

```bash
ncalls     tottime    tt_percall cumtime    ct_percall file:lineno (method)
2/1        0.000      0.000      21.459     21.459     /home/odoo/addons-vauxoo/account_anglo_saxon_stock_move/model/invoice.py:123 (invoice_validate)
9/8        0.000      0.000      12.770     1.596      /home/odoo/addons-vauxoo/invoice_datetime/invoice.py:119 (write)
1          0.000      0.000      11.045     11.045     /home/odoo/addons-vauxoo/account_tax_importation/model/account_invoice.py:221 (invoice_validate)
1          0.000      0.000      10.231     10.231     /home/odoo/addons-vauxoo/account_invoice_line_asset_category_required/models/ail_asset_category_required.py:65 (invoice_validate)
1          0.000      0.000      10.206     10.206     /home/odoo/addons-vauxoo/supplier_invoice_number_unique/model/account_invoice.py:65 (invoice_validate)
...
```

The `invoice_validate` use too much resources to detect a supplier_invoice_number unique.

This PR change that constraint a sql unique constraint faster

NOTE: Before to install this module you need verify that the following sql don't emit results:
`SELECT supplier_invoice_number,commercial_partner_id,count(*) AS repeated FROM account_invoice WHERE supplier_invoice_number IS NOT NULL AND state NOT IN ('draft', 'cancel') AND type IN ('in_invoice', 'in_refund') GROUP BY commercial_partner_id,supplier_invoice_number, company_id HAVING count(*) >=2`

PR #1211, opened 2017-07-13
[IMP] partner_validate_vat: Migration to api new

deivislaya:
vx#8249

**TODO**

- [x] Validate that Partner's VAT must be a unique value in each company


partner_validate_vat/res_partner.py
```diff
@@ -24,47 +24,37 @@
...
+from openerp.exceptions import ValidationError
 
 
-class ResPartner(osv.Model):
+class ResPartner(models.Model):
     _inherit = 'res.partner'
 
-    def _check_vat_uniqueness(self, cr, uid, ids, context=None):
+    @api.multi
+    @api.constrains('vat')
```
moylop260: Is possible create a sql unique constraint?

moylop260 (changes requested):
- Add tests
 - Use standard structure module (models)

partner_validate_vat/res_partner.py
deivislaya: For this case it is not possible, since we must validate not only the field  ```vat``` is not repeated in model ```res.partner```, it is already valid that a parent_id of  a partner have the same ```vat```. He is in the same model

partner_validate_vat/models/__init__.py
```diff
@@ -0,0 +1,27 @@
+# coding: utf-8
```
JulioSerna, line 1: use short version

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,55 @@
+# coding: utf-8
```
JulioSerna, line 1: same here

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,55 @@
...
+        """ Check that the vat is unique in the level
+            where the partner in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self:
+            current_vat = partner.vat
+
+            if all([user_company.country_id == partner.country_id, current_vat,
```
JulioSerna: replace `all` by (... `and` ...)

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,55 @@
...
+                    partner.search([
+                        ('vat', '=', current_vat),
+                        ('id', 'not in', [partner.parent_id.id, partner.id]),
+                        ('parent_id', '!=', partner.parent_id.id)
```
JulioSerna: you can replace this domain
```
('id', 'not in', [partner.parent_id.id, partner.id]),
('parent_id', '!=', partner.parent_id.id)
```
using `commercial_partner_id`

partner_validate_vat/models/__init__.py
deivislaya: Change applied. Using short version

partner_validate_vat/models/res_partner.py
deivislaya: Change applied. Using short version

partner_validate_vat/models/res_partner.py
deivislaya: change applied. Replaced by conditionals

partner_validate_vat/models/res_partner.py
deivislaya: Change applied. Using ```commercial_partner_id``` field

deivislaya:
Hi @moylop260 @JulioSerna

This PR is ready for review

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+
+        user_company = self.env.user.company_id
+
+        for partner in self:
+            current_vat = partner.vat
+            current_commercial_partner = partner.commercial_partner_id.id
+            if (user_company.country_id == partner.country_id and
+                current_vat and current_commercial_partner and partner.search([
+                    ('vat', '=', current_vat),
+                    ('id', 'not in', [partner.commercial_partner_id.id,
```
JulioSerna: Could you check if is necessary this validation?
I think only with this validation is enough
`('commercial_partner_id', '!=', partner.commercial_partner_id.id)`

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+            where the partner in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self:
+            current_vat = partner.vat
+            current_commercial_partner = partner.commercial_partner_id.id
+            if (user_company.country_id == partner.country_id and
+                current_vat and current_commercial_partner and partner.search([
```
JulioSerna: in what case `commercial_partner_id` could not be set?

deivislaya: When the ```res.partner``` record is created, the computed field ```commercial_partner_id.id``` is ```False``` in a first iteration in the constraints method, then the computed method is called and the field is calculated, then it is called in a second iteration the constraint method and the ```commercial_partner_id``` field , It is seted

partner_validate_vat/models/res_partner.py
deivislaya: Change applied. Deleted validation innecessary

deivislaya:
Hi @JulioSerna @moylop260 

This PR is ready for review

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    vat = fields.Char(copy=False)
+
+    @api.constrains('vat', 'country_id')
+    def _check_vat_uniqueness(self):
+        """ Check that the vat is unique in the level
```
moylop260: strip docstring

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+            where the partner in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self:
+            current_vat = partner.commercial_partner_id.vat
+            if (user_company.country_id == partner.country_id and
```
moylop260: I don't understand the rational of this validation `user_company.country_id == partner.country_id`

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+                current_vat and partner.search([
+                    ('commercial_partner_id', '!=',
+                     partner.commercial_partner_id.id)], limit=1)):
```
moylop260: Where are you validating the vat?

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.constrains('vat', 'country_id')
+    def _check_vat_uniqueness(self):
+        """ Check that the vat is unique in the level
+            where the partner in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self:
```
moylop260: Use `for partner in self.mapped('commercial_partner_id'):` in order to group by commercial and avoid execute the same validation 2 or more times

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+from openerp import models, api, fields, _
+from openerp.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    vat = fields.Char(copy=False)
+
+    @api.constrains('vat', 'country_id')
```
moylop260: commercial_partner_id

partner_validate_vat/models/res_partner.py
moylop260: Better adds `.filtered('vat')` too

partner_validate_vat/models/res_partner.py
deivislaya: change  applied

partner_validate_vat/models/res_partner.py
deivislaya: This is to validate that the user's country is the same as partner where the vat is being applied

partner_validate_vat/models/res_partner.py
deivislaya: change applied

partner_validate_vat/models/res_partner.py
deivislaya: This does not work for all tests, because ```commercial_partner_id``` is a computed field and its method is called after the ```@api.contrains``` in version 8

partner_validate_vat/models/res_partner.py
deivislaya: change applied

partner_validate_vat/models/res_partner.py
deivislaya: @moylop260 This test [Failed](https://travis-ci.com/Vauxoo/typ/jobs/83683685#L2271), applying this change but witn last commit is [resolved](https://github.com/Vauxoo/addons-vauxoo/pull/1211/commits/8b5ce9ff68ac01d9ff1348b176633e8fffddaef5)

partner_validate_vat/models/res_partner.py
moylop260: Yes, that told me the code.
Why you need validate the country of the user?
What is the rationale of `validate that the user's country is the same as partner where the vat is being applied`?

partner_validate_vat/res_partner.py
moylop260: Really, that is not a problem because you could use a extra field computed with store=True just for parent_id is False and use it directly in the sql-constraint
Something like as https://github.com/Vauxoo/addons-vauxoo/pull/1205

partner_validate_vat/models/res_partner.py
moylop260: Verify your tests because they don't detect this error and this is the main focus of this module.

partner_validate_vat/models/res_partner.py
moylop260: You can use `self.mapped('commercial_partner_id').filtered('vat')` because the `commercial_partner_id` is a stored field.

To `api.constrains` you need use the field that trigger the `commercial_partner_id`

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self.filtered(
+                lambda r: r.country_id == user_company.country_id and
+                r.vat is not False):
+            current_commercial = partner.commercial_partner_id
+            if not current_commercial:
+                current_commercial = partner.browse(
```
moylop260: Which cases `commercial_partner_id` is not assigned?

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+                    partner._commercial_partner_compute(
```
moylop260: Why you are calling the method directly if this field is computed?

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+                        self.env.cr, 'commercial_parent_id')[partner.id])
+            if (partner.search([
+                    ('vat', '=', current_commercial.vat),
+                    ('commercial_partner_id', '!=',
```
moylop260: You can see that you can use a sql search for this field (because is a stored one like as my comment above)

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+    def _check_vat_uniqueness(self):
+        """Check that the vat is unique in the level where the partner is
+        in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self.filtered(
+                lambda r: r.country_id == user_company.country_id and
+                r.vat is not False):
```
moylop260: What is the reason to validate specify `if r.vat is not False` instead of use just `if r.vat`?

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+    @api.constrains('vat', 'country_id', 'commercial_partner_id')
+    def _check_vat_uniqueness(self):
+        """Check that the vat is unique in the level where the partner is
+        in the tree
+        """
+
+        user_company = self.env.user.company_id
+
+        for partner in self.filtered(
+                lambda r: r.country_id == user_company.country_id and
```
moylop260: Add in the docstring the rationale of use `r.country_id == user_company.country_id`
Please, don't add the `what`, add the `why`

partner_validate_vat/models/res_partner.py
moylop260: Use `.mapped('commercial_partner_id')`

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+from openerp import models, api, fields, _
+from openerp.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    vat = fields.Char(copy=False)
+
+    @api.constrains('vat', 'country_id', 'commercial_partner_id')
```
moylop260: Check if here you won't have a error with quickly creation:
 - https://github.com/odoo/odoo/pull/18053

moylop260 (changes requested):
Check how you could use the following logic here:
https://github.com/Vauxoo/addons-vauxoo/pull/1205?

partner_validate_vat/models/res_partner.py
deivislaya: @moylop260 This is the [case](https://youtu.be/bMMsncBKpl0) where ```commercial_partner_id``` is not assigned

partner_validate_vat/models/res_partner.py
deivislaya: I try to validate the country of the partner in relation to the user's  country , because it may happen that he places the same vat in another partner located in another country without relation between them. For this case if it is accepted that the same vat is repeated for both partners.  This validation is important when placing the vat in the partner

partner_validate_vat/models/res_partner.py
deivislaya: Because the method is called after of  ```@api.constrains``` only version 8 with a recorset new

partner_validate_vat/models/res_partner.py
deivislaya: This case is allowed, a ```_sql_constraints``` would avoid it, and this case is possible

sql_constraints

partner_validate_vat/models/res_partner.py
deivislaya: Change applied

partner_validate_vat/models/res_partner.py
deivislaya: Change apllied by ```if r.vat```
The field ```commercial_partner_id``` is not assigned for recors new

partner_validate_vat/models/res_partner.py
deivislaya: quickly creation is success by is allowed partner without vat (rfc)

deivislaya:
Hi @JulioSerna @moylop260 

This PR is ready for review

partner_validate_vat/models/res_partner.py
moylop260: Are you sure?
(I don't see this docstring...)

deivislaya: Sorry the docstrins was incompleted, change applied

deivislaya:
Hi @moylop260

This PR is ready for review

partner_validate_vat/models/res_partner.py
deivislaya: This [test](https://github.com/vauxoo-dev/addons-vauxoo/blob/12ba054b76fa1916e9cb4d27f663080ab2da89a0/partner_validate_vat/tests/test_partner_vat_unique.py#L59) and [case](https://www.youtube.com/watch?v=bMMsncBKpl0&feature=youtu.be) are equal

deivislaya:
Hi @JulioSerna

This PR is ready for review

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+        but is from different countries.
+        3.- The partners that are related have the same fields: vat,
+        commercial_parent_id, country_id
+        """
+
+        user_company = self.env.user.company_id
+
+        # Check if it belongs to the same company in relation to point 2
+        for partner in self.filtered(
+                lambda r: r.country_id == user_company.country_id and r.vat):
```
JulioSerna: What about only validate the company_id?

I think the case of the same vat twice with two different customer from different country is not possible

what do you think?

deivislaya: I understand that it is unlikely but it is not impossible

Another reason I think now takes more importance, is that this allows an external partner to use a generic VAT identifier, as indicated by this [newspaper](https://contabilidad.tumentor.net/2017/01/rfc-generico-sat-2017.html)

deivislaya: Hi @JulioSerna

This PR is ready for review

moylop260: @luistorresm is using a method to use the generic VAT where is required use it.
But this module is not a `l10n_mx` one so you shouldn't add extra validation to get a generic VAT for MX.

And odoo use a method to validate the VAT based on the country.

If you want use a unique constraint use a minimal compute method and add a sql unique constraint like as:
  - https://github.com/Vauxoo/addons-vauxoo/pull/1205

partner_validate_vat/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+            current_commercial = partner.commercial_partner_id
+            if not current_commercial:
+                current_commercial = partner.browse(
+                    partner._commercial_partner_compute(
```
moylop260: You don't need a calling to compute method for a compute field. You should use directly partner.computed_field instead.
Dejavú I feeling that you are ignoring my suggestions or I didn't write them

partner_validate_vat/models/res_partner.py
moylop260: I mean, you could use:

```python

commercial_vat = fields.Char(compute=compute_commercial_vat, store=True)

@api.depends('vat', 'parent_id.commercial_partner_id', 'is_company')
def compute_commercial_vat(self):
   for partner in self.filtered(lambda p: (p.is_company or not p.parent_id) and p.vat):  # Detect just global partners (companies) avoiding children (contacts)
      partner.commercial_vat = re.sub(r'\W+|\_', '', partner.vat.upper())  # Replace all special chars and use uppercase in order to raise the exception if a small variation is used.

_sql_constraints = [
  UNIQUE('commercial_vat', 'company_id'),
]
```

NOTE: This code is just a example and there are syntax errors. Use it to understand the logic.

It code is so so so faster.

deivislaya:
Hi @JulioSerna @moylop260 

This PR is ready for review

JulioSerna:
@deivislaya 

I think we need to add something [like this](https://github.com/Vauxoo/addons-vauxoo/pull/1205/files#diff-6aaa5be6a0ba58e5b1b2b73f53af4c68R12)

moylop260:
One month ago I told you that
https://github.com/Vauxoo/addons-vauxoo/pull/1211#discussion_r128902264

And other comments I repeated this comment.
But you are ignoring my comments and request feedback again and again.

I'm tired of this situation.
Please don't ping me if you are thinking to ignore my comments

partner_validate_vat/models/res_partner.py
deivislaya: Change applied following your comments. Thanks

partner_validate_vat/models/res_partner.py
deivislaya: called to computed method deleted

partner_validate_vat/res_partner.py
deivislaya: This change have been apllied in the commit and we use a extra field compute with ```store=True```. This we allow us use sql_contraint. Thanks for you comment,has been of great use

partner_validate_vat/models/res_partner.py
deivislaya: ok thanks

luisg123v:

PR #1212, opened 2017-07-18
[ADD] delivery_available_domain: show available carriers

suniagajose:
Summary
-------------

In sale order when select a partner or change the sale order lines the delivery carriers proposed will change
according to its config.


luisg123v:
This seems abandoned, so I'll close it. Feel free to reopen it if not.

PR #1213, opened 2017-07-19
8.0 account_anglo_saxon_stock_move_purchase migration

hbto:
When a functional field moves from store=False to store=True,
`at -u all -d $BD ` it takes too long to perform that task, 
then a migration script prior to perform the update of the module
is execute to avoid the huge amount of time be wasted when updating.


hbto:
@moylop260 please review.

a MP will be created on customers project.

moylop260:
See VXT#7980 for more info about

PR #1218, opened 2017-08-03
[IMP] account_anglo_saxon_stock_move: Improvement performance

hbto:
Use batches to process reconciling and avoid trigger the computed field related many times.

<s>This PR will try to fix a performance issue when validating invoices with many invoice lines
at account_move_line aml creation will try to reconcile the lines and at the same time will try to update functional fields that are aml sensitive in invoice that were introduced by commission_payments.</s> This part was superseded by https://github.com/Vauxoo/addons-vauxoo/pull/1219

### DIFF
--- a/account_anglo_saxon_stock_move/model/invoice.py
+++ b/account_anglo_saxon_stock_move/model/invoice.py
@@ -43,87 +43,70 @@ class AccountInvoice(models.Model):
 
     @api.multi
     def action_cancel(self):
-        aml_obj = self.env['account.move.line']
-        aml_ids = []
-        for inv in self:
-            if not inv.move_id:
-                continue
-            for aml_brw in inv.move_id.line_id:
-                # Unreconcile all non-receivable and non-payable lines
-                if aml_brw.account_id.reconcile and aml_brw.account_id.type \
-                        not in ('receivable', 'payable'):
-                    aml_ids.append(aml_brw.id)
-        if aml_ids:
-            aml_obj._remove_move_reconcile(aml_ids)
-
+        for inv in self.filtered(lambda inv: inv.move_id):
+            amr_ids = inv.invoice_line.mapped('move_id.aml_ids.reconcile_id')
+            amr_ids.unlink()
         return super(AccountInvoice, self).action_cancel()
 
-    def reconcile_stock_accrual(self, cr, uid, ids, context=None):
-        context = dict(context or {})
-        ids = isinstance(ids, (int, long)) and [ids] or ids
-        aml_obj = self.pool['account.move.line']
-        amr_obj = self.pool['account.move.reconcile']
-        for inv_brw in self.browse(cr, uid, ids, context=context):
-            for ail_brw in inv_brw.invoice_line:
-                ail_brw.refresh()
-                if not ail_brw.move_id:
+    @api.multi
+    def reconcile_stock_accrual(self):
+        aml_obj = self.env['account.move.line']
+        for inv_brw in self:
+            # We just care about lines which have a stock_move.aml_ids related
+            all_aml_ids = aml_obj
+            aml_ids = inv_brw.invoice_line.mapped('move_id.aml_ids')
+
+            # In order to keep every single line reconciled we will look for
+            # all the lines related to a purchase/sale order
+            all_aml_ids |= aml_ids.mapped('purchase_id.aml_ids')
+            all_aml_ids |= aml_ids.mapped('sale_id.aml_ids')
+
+            categ_ids = all_aml_ids.filtered(
+                lambda m:
+                m.product_id and
+                not m.product_id.categ_id.property_stock_journal)
+            if categ_ids:
+                raise ValidationError(_(
+                    'The Stock Journal is missing on following '
+                    'product categories: %s' % (', '.join(
+                        categ_ids.mapped('name')))
+                ))
+
+            res = {}
+            # Only stack those that are fully reconciled
+            amr_ids = all_aml_ids.mapped('reconcile_id')
+            amr_ids.unlink()
+
+            # Let's group all the Accrual lines by Purchase/Sale Order, Product
+            # and Account
+            for aml_brw in all_aml_ids.filtered('account_id.reconcile'):
+                doc_brw = aml_brw.purchase_id or aml_brw.sale_id
+                account_id = aml_brw.account_id.id
+                product_id = aml_brw.product_id
+                res.setdefault((doc_brw, account_id, product_id), aml_obj)
+                res[(doc_brw, account_id, product_id)] |= aml_brw
+
+            for (doc_brw, account_id, product_id), aml_ids in res.items():
+                if not len(aml_ids) > 1:
                     continue
-                if ail_brw.move_id.product_id != ail_brw.product_id:
-                    continue
-
-                res = {}
-                amr_ids = [
-                    aml_brw1.reconcile_id.id or
-                    aml_brw1.reconcile_partial_id.id
-                    for aml_brw1 in ail_brw.move_id.aml_ids
-                    if aml_brw1.reconcile_id or aml_brw1.reconcile_partial_id
-                ]
-
-                if amr_ids:
-                    amr_ids = list(set(amr_ids))
-                    amr_obj.unlink(cr, uid, amr_ids, context=context)
-
-                ail_brw.refresh()
-                aml_brws = [
-                    aml_brw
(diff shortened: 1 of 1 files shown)


account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -58,65 +58,50 @@ def action_cancel(self):
...
-                    continue
-                if ail_brw.move_id.product_id != ail_brw.product_id:
+    @api.multi
+    def reconcile_stock_accrual(self):
+        aml_obj = self.env['account.move.line']
+        for inv_brw in self:
+            # We just care about lines which have a stock_move related
+            move_ids = inv_brw.invoice_line.mapped('move_id')
+
+            categ_ids = move_ids.mapped('product_id.categ_id').filtered(
```
moylop260: Use directly `filtered('property_stock_journal')` for positive fields

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -58,65 +58,50 @@ def action_cancel(self):
...
+
+            res = {}
+            # Only stack those that are fully reconciled
+            amr_ids = move_ids.mapped('aml_ids.reconcile_id')
+            amr_ids.unlink()
+
+            for aml_brw in aml_obj.search([('sm_id', 'in', move_ids.ids)]):
+                sm_id = aml_brw.sm_id
+                account_id = aml_brw.account_id.id
+                res.setdefault((sm_id, account_id), aml_obj)
```
moylop260: `setdefault` returns the value if exists then you could use it directly since that is a object used with pointers, I mean

```python
value = res.setdefault((sm_id, account_id), aml_obj)
value |= aml_brw
```

This avoid double indexed
screen shot 2017-08-08 at 23 29 06

moylop260 (changes requested):

moylop260 (approved):

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
-        amr_obj = self.pool['account.move.reconcile']
-        for inv_brw in self.browse(cr, uid, ids, context=context):
-            for ail_brw in inv_brw.invoice_line:
-                ail_brw.refresh()
-                if not ail_brw.move_id:
-                    continue
-                if ail_brw.move_id.product_id != ail_brw.product_id:
+    @api.multi
+    def reconcile_stock_accrual(self):
+        aml_obj = all_aml_ids = self.env['account.move.line']
```
moylop260: Shouldn't be initialized `all_aml_ids` for each iteration of the `for` loop?

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
+                ))
+
+            res = {}
+            # Only stack those that are fully reconciled
+            amr_ids = all_aml_ids.mapped('reconcile_id')
+            amr_ids.unlink()
+
+            # Let's group all the Accrual lines by Purchase/Sale Order, Product
+            # and Account
+            for aml_brw in all_aml_ids.filtered(
```
moylop260: I just tested `all_aml_ids.filtered('account_id.reconcile')` and is working fine!

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
+                    lambda l: l.account_id.reconcile):
+                doc_brw = aml_brw.purchase_id or aml_brw.sale_id
+                account_id = aml_brw.account_id.id
+                product_id = aml_brw.product_id
```
moylop260: Shouldn't `product_id = aml_brw.product_id.id` or `account_id = aml_brw.account_id` to unified objects or integers?

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
+                res.setdefault((doc_brw, account_id, product_id), aml_obj)
+                res[(doc_brw, account_id, product_id)] |= aml_brw
+
+            for (doc_brw, account_id, product_id), aml_ids in res.items():
+                journal_id = product_id.categ_id.property_stock_journal.id
```
moylop260: Define this variable after the `continue` because you could save time if the continue is executed so don't need assign the variable

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
-            for aml_brw in inv.move_id.line_id:
-                # Unreconcile all non-receivable and non-payable lines
-                if aml_brw.account_id.reconcile and aml_brw.account_id.type \
-                        not in ('receivable', 'payable'):
-                    aml_ids.append(aml_brw.id)
-        if aml_ids:
-            aml_obj._remove_move_reconcile(aml_ids)
-
+        for inv in self.filtered(lambda inv: inv.move_id):
+            amr_ids = inv.invoice_line.mapped('move_id.aml_ids.reconcile_id')
```
moylop260, line 47: I saw too much changes from original to this one.
Are you sure that is working with the same behaviour?

If is yes: 👍 Excellent!

account_anglo_saxon_stock_move/model/invoice.py
hbto: indeed

account_anglo_saxon_stock_move/model/invoice.py
hbto: Let's do it

account_anglo_saxon_stock_move/model/invoice.py
hbto: Yes cancellation is working as expected and is faster than before

account_anglo_saxon_stock_move/model/invoice.py
hbto: `account_id = aml_brw.account_id.id` in here I use it group by account_id (integer) 
but here `product_id = aml_brw.product_id` I use it as an object because I want it 
to get advantage of object [here](https://github.com/Vauxoo/addons-vauxoo/pull/1218/files#diff-f6d45b6288783ab3ee76bc9962939e08R90)

account_anglo_saxon_stock_move/model/invoice.py
hbto: indeed

moylop260 (approved):

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -43,87 +43,70 @@ def line_get_convert(self, line, part, date):
...
+
+            # In order to keep every single line reconciled we will look for
+            # all the lines related to a purchase/sale order
+            all_aml_ids |= aml_ids.mapped('purchase_id.aml_ids')
+            all_aml_ids |= aml_ids.mapped('sale_id.aml_ids')
+
+            categ_ids = all_aml_ids.filtered(
+                lambda m:
+                m.product_id and
+                not m.product_id.categ_id.property_stock_journal)
```
luistorresm: I think that if not `m.product` then `m.product_id.categ_id.property_stock_journal` also is empty.

Maybe only is necessary `m.product_id.categ_id.property_stock_journal`

luistorresm: Sorry, I`m bad

hbto:
This MP/PR has been squashed / rebased and is ready to be merged.

@moylop260

moylop260:
LGTM

hbto:
Tests in customer's instance have been quite successful:
- Cancelation of a Invoice with over 300 lines has gone from more than 20mins to 5 mins.
- Validation of a Invoice with over 300 lines has gone from more than 1 hour to just 15mins.

Now we can visualize many Journal Items in a single Reconciliation:
journal_items_-_odoo

hbto:
Let's keep an eye in this branch as fore-coming changes in master appears to
include reconciliation created by the stock-in & stock-out accounts.

https://github.com/odoo-dev/odoo/blob/master-stock-val-cur-rate-reconcile-oco/addons/stock_account/models/account_invoice.py#L101-L126

sabrinagnu (approved):

PR #1219, opened 2017-08-05
[REF] stock_move_entries, invoice_discount, commission_payment: Improvement performance on computed fields

moylop260:
Add api.depends and store=True to computed fields and migrated to new api

Dummy PR: https://github.com/Vauxoo/lodigroup/pull/719

### DIFF
--- a/invoice_discount/model/invoice.py
+++ b/invoice_discount/model/invoice.py
@@ -78,11 +78,13 @@ class AccountInvoice(models.Model):
                  'invoice_line.discount_amount')
     def _compute_subtotal_wo_disc(self):
         line = self.env['account.invoice.line']
+        res = line.read_group(
+            domain=[('invoice_id', 'in', self.ids)],
+            fields=['invoice_id', 'subtotal_wo_discount', 'discount_amount'],
+            groupby='invoice_id')
+        invoice_sum = dict([(item['invoice_id'][0], item) for item in res])
         for invoice in self:
-            res = line.read_group(
-                domain=[('invoice_id', '=', invoice._cache['id'])],
-                fields=['invoice_id', 'subtotal_wo_discount',
-                        'discount_amount'],
-                groupby='invoice_id')[0]
-            invoice.subtotal_wo_discount = res.get('subtotal_wo_discount', 0)
-            invoice.discount_amount = res.get('discount_amount', 0)
+            invoice.subtotal_wo_discount = invoice_sum.get(invoice.id, {}).get(
+                'subtotal_wo_discount')
+            invoice.discount_amount = invoice_sum.get(invoice.id, {}).get(
+                'discount_amount')
--- a/invoice_discount/model/invoice.py
+++ b/invoice_discount/model/invoice.py
@@ -77,11 +77,12 @@ class AccountInvoice(models.Model):
     @api.depends('invoice_line.subtotal_wo_discount',
                  'invoice_line.discount_amount')
     def _compute_subtotal_wo_disc(self):
-        # TODO: Use read_group method here
+        line = self.env['account.invoice.line']
         for invoice in self:
-            invoice._cache['subtotal_wo_discount'] = 0
-            invoice._cache['discount_amount'] = 0
-            for line in invoice.invoice_line:
-                invoice._cache['subtotal_wo_discount'] += (
-                    line.subtotal_wo_discount)
-                invoice._cache['discount_amount'] += line.discount_amount
+            res = line.read_group(
+                domain=[('invoice_id', '=', invoice._cache['id'])],
+                fields=['invoice_id', 'subtotal_wo_discount',
+                        'discount_amount'],
+                groupby='invoice_id')[0]
+            invoice.subtotal_wo_discount = res.get('subtotal_wo_discount', 0)
+            invoice.discount_amount = res.get('discount_amount', 0)
--- a/invoice_discount/model/invoice.py
+++ b/invoice_discount/model/invoice.py
@@ -23,133 +23,65 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-"""File to add functionalitity in account.invoice.line to get the amount without
+"""File to add functionality in account.invoice.line to get the amount without
 discount and the value of the discount
 """
-from openerp.osv import osv, fields
+from openerp import api, fields, models
 
 
-class AccountInvoiceLine(osv.osv):
+class AccountInvoiceLine(models.Model):
 
     """Inherit from account.invoice.line to get by line the amount without
     discount and the amount of this
     """
     _inherit = 'account.invoice.line'
 
-    def _get_subtotal_without_discount(self, cr, uid, ids, args, field_name,
-                                       context=None):
+    subtotal_wo_discount = fields.Float(
+        store=True, string="SubTotal w/o Discount",
+        help=('Amount without apply the discount of the line, '
+              'is calculated as Qty * Price Unit'),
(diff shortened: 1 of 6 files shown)


account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -59,63 +59,42 @@ def action_cancel(self):
...
-
+            categ_ids = inv_brw.invoice_line.filtered(
+                lambda l: not l.product_id.categ_id.property_stock_journal)
             if categ_ids:
                 raise ValidationError(_(
                     'The Stock Journal is missing on following '
-                    'product categories: %s' % (
-                        ', '.join([cat.name for cat in categ_ids]))
+                    'product categories: %s' % (', '.join(
+                        categ_ids.mapped('product_id.categ_id.name')))
```
hbto: better it be like
`categ_ids.mapped('name')))`
?

moylop260:
@josemoralesp @luistorresm Could you give me your technical revision, please?

invoice_discount/model/invoice.py
```diff
@@ -23,133 +23,65 @@
...
-                                                invoice.'),
-        'discount_amount': fields.function(_get_discount, string='Discount',
-                                           store=False, type='float',
-                                           help='Total of discount apply in \
-                                           each line of the invoice.'),
-    }
+    @api.depends('invoice_line.subtotal_wo_discount',
+                 'invoice_line.discount_amount')
+    def _compute_subtotal_wo_disc(self):
+        # TODO: Use read_group method here
```
moylop260: @luisg123v Could you help to fix this TODO, please?
@josemoralesp can help you to understand the read_group to sum values

moylop260:
@luisg123v Could you help me to check if migrations script is working fine vs the original behaviour?

invoice_discount/model/invoice.py
```diff
@@ -23,133 +23,65 @@
...
+        for invoice in self:
```
luistorresm: @moylop260 

Is best _cache[] that an python variable?

To have only one write, you could assign the sum in a variable, 

What do you think about
`
subtotal_wo_discount = sum([line.mapped('subtotal_wo_discount')])`

moylop260: That is a good point. Thanks!

But we will use `read_group` is faster and cleaned.

hbto:
@moylop260 Please rebase your branch on new commits from
https://github.com/Vauxoo/addons-vauxoo/pull/1218

Regards

invoice_discount/model/invoice.py
hbto: So far I haven't been able to use this method `read_group` as would do `pandas`
I think that library could do this job better.

hbto (commented):
Would it be better to use `pandas` library than use `read_group` method?

invoice_discount/model/invoice.py
josemoralesp: Hello @moylop260 
I think the same way that you do. read_group is the fastest solution.

I did a test in a copy of apex using two similar invoice_line fields
PSQL:
desktop screenshot 1

read_group with odoorpc
desktop screenshot 2
In our case would be something like this:
```python
line = self.env['account.invoice.line']
for invoice in self:
    results = line.read_group([('invoice_id', '=', invoice.id)], ['discount_amount', 'subtotal_wo_discount', 'invoice_id'], ['invoice_id'])
    invoice.subtotal_wo_discount = results and results[0].get('subtotal_wo_discount') or 0
    invoice.discount_amount = results and results[0].get('discount_amount') or 0
```
There is an important question. 
Are these fields stored in database?
 The read_group method will be useful, only if these fields are stored in database, if one of them isn't we will have to use panda as @hbto has said before

stock_move_entries/model/stock_move_entries.py
```diff
@@ -30,11 +30,19 @@
...
-    """
-    """
+    def _auto_init(self, cr, context=None):
+        res = super(AccountMoveLine, self)._auto_init(cr, context=context)
+        index = 'account_move_line_sm_id_account_id_index'
+        cr.execute("SELECT indexname FROM pg_indexes WHERE indexname = %s",
+                   (index,))
+        if not cr.fetchone():
+            cr.execute(
+                'CREATE INDEX %s ON account_move_line (account_id, sm_id)' %
```
josemoralesp, line 40: @moylop260  from postgres 9.5 it was added the option IF NOT EXISTS, you can avoid using the first select to verify if this doesn't exist.
```python
cr.execute(
    'CREATE INDEX IF NOT EXISTS %s ON account_move_line (account_id, sm_id)' % index)
```
More about that [here](https://www.postgresql.org/docs/9.5/static/sql-createindex.html)

invoice_discount/model/invoice.py
luisg123v: Wouldn't be possible to adapt method `filtered` to do the job? i.e. getting all subtotals to be summed. As far as I know, the efficiency of the `filtered` method is very good.

luisg123v: Please take into account that, since the `read_group` operates only with values already stored on database, if we edit an invoice (e.g. adding a new line), the computed value will not be properly calculated until that invoice is saved.

invoice_discount/model/invoice.py
luisg123v: @josemoralesp I believe that code wouldn't work because of two reasons:
- When an invoice is being modified, its ID is not accessible, it's a NewId object. We could workaround that using `invoice._cache['id']`
- If we use `groupby='invoice_id'`, then there would be only one result, because all lines belongs to the same invoice.

Am I wrong?

josemoralesp: @luisg123v The field  belongs to account invoice model and it is ```store=True```, the field should be computed when the invoice is saved(theoretically)
Did you do this test ?

invoice_discount/model/invoice.py
```diff
@@ -23,133 +23,66 @@
...
+    @api.depends('invoice_line.subtotal_wo_discount',
+                 'invoice_line.discount_amount')
+    def _compute_subtotal_wo_disc(self):
+        line = self.env['account.invoice.line']
+        for invoice in self:
+            res = line.read_group(
+                domain=[('invoice_id', '=', invoice._cache['id'])],
+                fields=['invoice_id', 'subtotal_wo_discount',
+                        'discount_amount'],
+                groupby='invoice_id')[0]
```
josemoralesp: @luisg123v I don't agree about assume that the lines exist. You have to be sure before use __getitem__ method(Asking for the first value)

hbto:
I would like to have these changes stabilized in order to make them available when
https://github.com/Vauxoo/addons-vauxoo/pull/1218 get merged

@josemoralesp @moylop260 @luisg123v @luistorresm

invoice_discount/model/invoice.py
moylop260: Using read group you can use all ids in one call or not?
And you dont need use cache for get the ids

stock_move_entries/model/stock_move_entries.py
moylop260: That is cool!
But we are not sure if this module is used with postgresql >=9.5 for all cases.

IMHO we should use the full compatibility way.

moylop260:
LGTM

hbto (approved):

hbto:
LGTM

luisg123v:
I've tested the fields from the invoice model, they're working as expected.

josemoralesp (approved):

PR #1220, opened 2017-08-14
[FIX] supplier_invoice_number_unique: Fixed tests 

luistorresm:
Test that verify that reference is unique by supplier invoices was fixed, because in the last change the constraint was changed


luistorresm:
@dsabrinarg @josemoralesp 

Could you check this PR, please?

Regards

josemoralesp (approved):

PR #1221, opened 2017-08-23
[8.0] [IMP]Take away Accrual Reconciliation from Invoice to Sale/Purchase Order

hbto:
More information here:
https://docs.google.com/document/d/1e9q3LvNmyLx2AYks2rYADCqWP8VdAzzcpy436XO3jmQ/edit#


hbto:
@moylop260 & @nhomar 
I'd like to move the code in invoice.py that is not directly related to account.invoice 
to somewhere else like TransientModel, I have leave in the meantime in account.invoice 
in order to avoid duplicate the code for both purchase.order and sale.order.

Any suggestion?

Regards.

PR #1225, opened 2017-08-31
[REF] partner_credit_limit: Create function to return move_lines domain

ysantiago:


partner_credit_limit/model/sale.py
```diff
@@ -15,7 +15,7 @@ def check_limit(self):
                 return True
             allowed_sale = self.env['res.partner'].with_context(
                 {'new_amount': so.amount_total,
-                 'new_currency': so.company_id.currency_id.id}).browse(
+                 'new_currency': so.currency_id}).browse(
```
JulioSerna: not necessary browse you can use

``` python
allowed_sale = po.partner_id.with_context(
 {'new_amount': so.amount_total,
'new_currency': so.currency_id}).allowed_sale
```

could you check if there are another cases?

partner_credit_limit/model/partner.py
```diff
@@ -27,52 +27,53 @@ class ResPartner(models.Model):
...
-            company = res_company.browse(company_id)
-            new_amount = context.get('new_amount', 0.0)
-            new_currency = context.get('new_currency', False)
-            if new_currency:
-                from_currency = currency_obj.browse(new_currency)
-            else:
-                from_currency = company.currency_id
-            new_amount_currency = from_currency.compute(
-                new_amount, company.currency_id)
+            company = self.env.user.company_id
```
JulioSerna: move out of the `for`

partner_credit_limit/model/partner.py
```diff
@@ -27,52 +27,53 @@ class ResPartner(models.Model):
...
+                # credit and debit maturity sums all aml
+                # with late payments
+                debit_maturity += line.debit
+            credit_maturity += line.credit
+        return debit_maturity, credit_maturity
+
     @api.multi
     def _get_overdue_credit(self):
         for partner in self:
             moveline_obj = self.env['account.move.line']
```
JulioSerna: move out of the `for`

PR #1228, opened 2017-09-23
[saas-18][FIX] Modified deprecated libraries

josemoralesp:
Added few changes to keep this branch as a dependency of a branch in saas-18


luisg123v:
saas-18 is not used anymore, and inteco is not a customer anymore. However, this is harmless.

PR #1232, opened 2017-09-30
[FIX] Time consuming Journal Item creation process can be avoided

hbto:
Time consuming Journal Item creation process can be avoided at Invoice validation if a context['novalidate'] = True is passed down to Journal Item creation method.


hbto:
@nhomar @moylop260 @JulioSerna @josemoralesp

moylop260:
@JulioSerna help us to point us what we should test it?

PR #1233, opened 2017-10-02
[FIX] supplier_invoice_number_unique: fixed the trace back triggered by the unique constraint at the invoice validation event task#9288

hugho-ad:
Fixed by setting a deferred constraint.

Idea taken from the issue https://github.com/odoo/odoo/issues/15395

https://www.postgresql.org/docs/9.1/static/sql-set-constraints.html

DUMMY PR https://github.com/Vauxoo/wohlert/pull/93

TEST INSTANCE
http://dev.lodigroup.com:32855

### DIFF
--- a/supplier_invoice_number_unique/model/account_invoice.py
+++ b/supplier_invoice_number_unique/model/account_invoice.py
@@ -20,8 +20,10 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # ##########################################################################
 import re
+from psycopg2 import IntegrityError
 
-from openerp import api, fields, models
+from openerp import api, fields, models, _
+from openerp import exceptions
 
 
 class AccountInvoice(models.Model):
@@ -39,6 +41,16 @@ class AccountInvoice(models.Model):
                 inv.supplier_invoice_number):
             invoice.supplier_invoice_number_strip = re.sub(
                 r'\W+|\_', '', invoice.supplier_invoice_number.lower())
+
+    @api.multi
+    def invoice_validate(self):
+        try:
+            res = super(AccountInvoice, self).invoice_validate()
+        except IntegrityError:
+            raise exceptions.Warning(
+                _('Error you can not validate the invoice with '
+                  'supplier invoice number duplicated.'))
+        return res
 
     _sql_constraints = [
         ('unique_supplier_invoice_number_strip', 'UNIQUE('
--- a/supplier_invoice_number_unique/i18n/es.po
+++ b/supplier_invoice_number_unique/i18n/es.po
@@ -42,3 +42,10 @@ msgstr "¡Acción no válida!"
 #: model:ir.model,name:supplier_invoice_number_unique.model_account_invoice
 msgid "Invoice"
 msgstr "Factura"
+
+#. module: supplier_invoice_number_unique
+#: sql_constraint:account.invoice:0
+#: code:addons/supplier_invoice_number_unique/model/account_invoice.py:51
+#, python-format
+msgid "Error you can not validate the invoice with supplier invoice number duplicated."
+msgstr "Error no puedes validar facturas con número de factura de proveedor duplicado"
--- a/supplier_invoice_number_unique/tests/test_supplier_invoice_reference.py
+++ b/supplier_invoice_number_unique/tests/test_supplier_invoice_reference.py
@@ -7,8 +7,10 @@
 #    info Vauxoo (info@vauxoo.com)
 #    Coded by: Luis Torres (luis_t@vauxoo.com)
 # ##########################################################################
-from openerp.tests.common import TransactionCase
+
 from openerp import workflow
+from openerp import exceptions
+from openerp.tests.common import TransactionCase
 
 
 class TestReferenceSupplierInvoiceUnique(TransactionCase):
@@ -42,8 +44,8 @@ class TestReferenceSupplierInvoiceUnique(TransactionCase):
         self.assertEquals(
             invoice.state, 'open', 'This invoice has not state in open')
         with self.assertRaisesRegexp(
-                ValueError,
-                'account_invoice_unique_supplier_invoice_number_strip'):
+                exceptions.Warning, 'Error you can not validate the invoice '
+                'with supplier invoice number duplicated.'):
             invoice2.signal_workflow('invoice_open')
 
     def test_12_validate_invoice_with_ref_unique_hyphen(self):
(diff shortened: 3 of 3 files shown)


hugho-ad:
@dsabrinarg this is ready to review

Test instance 
http://dev.lodigroup.com:32855
DUMMY
https://github.com/Vauxoo/wohlert/pull/93
LINT FIXES
https://github.com/Vauxoo/wohlert/pull/94

Regards

sabrinagnu:
@hugho-ad error message translation is missing

sabrinagnu:
@josemoralesp may you please check this technically?

hugho-ad:
@dsabrinarg 

> @hugho-ad error message translation is missing

Which translation ?, I do not get it

sabrinagnu:
This @hugho-ad 
facturas de proveedor odoo

hugho-ad:
thanks!

hugho-ad:
@dsabrinarg 

The warning message can not be translatedm it is not the only one, also other warinig messages are not tranlated
image

Also I check the translation statements, and the warning meesage has its proper translation statement
image

This is a odoo issue, look at this related issue
https://github.com/odoo/odoo/issues/3634

If you agree, I would like to solve this issue in a separate pr

sabrinagnu:
@JulioSerna may you please gave us your approval here?, @hugho-ad as soon as Julio confirm then we will move forward with this, understood the issue with the translation.

supplier_invoice_number_unique/model/account_invoice.py
```diff
@@ -40,8 +40,24 @@ def _compute_supplier_invoice_number_strip(self):
...
+        query = """
+        SET CONSTRAINTS
+            account_invoice_unique_supplier_invoice_number_strip DEFERRED
+        """
+        if constraint != 'DEFERRED':
+            query = """
+            SET CONSTRAINTS
+                account_invoice_unique_supplier_invoice_number_strip IMMEDIATE
+            """
+        self.env.cr.execute(query)
```
josemoralesp: i don't understand why you are changing the constraint type at the moment  to validate the invoice

hugho-ad: because the constraint are not raised with the odoo warning, is raised like a traceback

image

So, in order to get the raised odoo warning I change the mode of sql constraint

JulioSerna: @hugho-ad 
commonly the sql constraint return warning
do you know why in this case return traceback, maybe because is trigger by function fields?

hugho-ad: I think that the most probable cause is  because it is trigger by the work flow, instead of a direct write over the record, (save button on the form view or simillar)

hugho-ad: @josemoralesp @JulioSerna 
can you give your feedback here , what dou you think ?

Should I debug more deeper?

Regards!

sabrinagnu: @hugho-ad let's wait a little bit for the review of @JulioSerna he need to make a test but is really busy at this moment, he will let us know asap.

JulioSerna: @hugho-ad 

I make some test and I think is by the function fields in the test I changed the field by char with onchange

```
         @api.onchange('supplier_invoice_number', 'state', 'type', 'fiscal_position')
         def onchange_number(self):
             self.supplier_invoice_number_strip = re.sub(
                 r'\W+|\_', '', self.supplier_invoice_number.lower())
```

and the warning is returned

Could you check if this is happening in version11.0 to report the issue?

about your pr what about make the constraint IMMEDIATE in order to do not change in invoice_validation method, but I do not the impact of this change as in the issue comment

>Doing this appears to fully work round the issue - but in more complex scenarios I have a feeling this would blow up, as it's a pretty sketchy workaround

what do you think @moylop260 ?

hugho-ad: @JulioSerna 
> I make some test and I think is by the function fields in the test I changed the field by char with onchange and the warning is returned

Instead a computeted field, do you change the field in order to made it  a normal field with a onchange function?

Also The field must  be in the view in order to make works the sql_constraint, I also made the test, this can work just keeping the field invisible, what do you think @dsabrinarg, this works only when you are edditing the record and saving it or triggering the workflow while the form view is in edit mode

The difference is that witht the onchange function, the field `supplier_invoice_number_strip` gonna be calculated in each editing record from the form view

> about your pr what about make the constraint IMMEDIATE in order to do not change in invoice_validation method, 

I already try it, making the constraint IMMEDIATE, the warinig error is NOT displayed and instead of it, appears the traceback error

> but I do not the impact of this change as in the issue comment

in the  issue,  PeterAlabaster  comments that because he is making a `cr.commit` in the writte method.

> Could you check if this is happening in version11.0 to report the issue?

Sure, I test it already, and the warning error is triggered properly does not matter if it trigger from a computed field, but I was not able to tets it from a workflow

I test it with the following code:
```python
test2 = fields.Char(compute="_compute_test_field", store=True)                  
test = fields.Char()                                                            
                                                                                      
      @api.depends('test')                                                                                                                                                                                                                     
      def _compute_test_field(self):                                                  
          for record in self:                                                         
              if record.test == 'test':                                               
                  record.test2 = record.test                                          
                                                                                      
      _sql_constraints = [                                                            
          ('test_uniq', 'unique(partner_id, test2)', 'TEST!'),                        
      ]                                                
```

josemoralesp: @dsabrinarg @JulioSerna @hugho-ad 

I don't like this solution.

If the problem is [this](https://github.com/Vauxoo/addons-vauxoo/pull/1233/files#r142979576). I rather use something like this(While we reporting the error to odoo)

```python
def invoice_validate(self):
    try:
        res = super(AccountInvoice, self).invoice_validate()
    except ValueError, msg:
        raise BEAUTIFUL_HUGO_MSG
    return res
```

What do you think?

Best Regards.

josemoralesp (changes requested):

hugho-ad:
@josemoralesp 

That was my first thought, and I have already try it without success results, the odoo warning does not be triggered, since the transaction was being from the workflow

Regards

josemoralesp:
@hugho-ad I'll check it out

hugho-ad:
@josemoralesp can you review please

josemoralesp (approved):

hugho-ad:
@dsabrinarg can you merge please

sabrinagnu:
@hugho-ad I can't, this need also be approved by Julio because he was giving also feedback, he will check asap.

JulioSerna (approved):

PR #1235, opened 2017-10-03
[IMP] Rewriting name to improve performance

hbto:
[IMP] Caching _name_get method to improve performance in stock.location related searches.

### DIFF
--- a/stock_location_code/model/stock.py
+++ b/stock_location_code/model/stock.py
@@ -77,7 +77,7 @@ class StockLocation(models.Model):
             return res
         wh_obj = self.env['stock.warehouse']
         wh_dict = {}.fromkeys(self._ids, False)
-        query = self._cr.mogrify('''
+        self._cr.execute('''
             SELECT
                 sl.id AS location_id,
                 ARRAY_AGG(sw.id) AS warehouse_id
@@ -89,7 +89,6 @@ class StockLocation(models.Model):
                 AND sl.id IN %(ids)s
             GROUP BY sl.id
         ''', {'ids': tuple(self._ids)})
-        self._cr.execute(query)
         wh_dict.update(
             dict((l, wh_obj.browse(w[0]))
                  for l, w in self._cr.fetchall()))
--- a/stock_location_code/model/stock.py
+++ b/stock_location_code/model/stock.py
@@ -73,6 +73,8 @@ class StockLocation(models.Model):
         Visually, it's better to know which owns the warehouse location.
         """
         res = []
+        if not self._ids:
+            return res
         wh_obj = self.env['stock.warehouse']
         wh_dict = {}.fromkeys(self._ids, False)
         query = self._cr.mogrify('''
--- a/stock_location_code/model/stock.py
+++ b/stock_location_code/model/stock.py
@@ -34,7 +34,7 @@ e.g. [A012345] Stock(El Dorado)
 
 import re
 
-from openerp import api, models, tools
+from openerp import api, models
 
 
 class StockLocation(models.Model):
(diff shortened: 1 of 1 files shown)


hbto:
@nhomar @moylop260 this is the meanwhile performance improve we were talking about
in order to later applied a more thorough improve performance refactorization.

This cache reduced the calls from 1300 to 250 unique calls.
Regards

stock_location_code/model/stock.py
```diff
@@ -83,3 +85,12 @@ def _name_get(self, location):
...
         return new_name
+
+    @api.model
+    def _name_get(self, location):
+        """Implements the stock location code in a new feauture, where if the
+        location has a reference to a warehouse, the name in a m2o search
+        concatenates the code and warehouse to a location if they exist.
+        Visually, it's better to know which owns the warehouse location.
+        """
+        return self.__name_get(location.id)
```
moylop260: You need clear caches after write, create or unlink a record.

stock_location_code/model/stock.py
```diff
@@ -65,21 +65,44 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
-            if warehouse else ''
-        items = [barcode.strip(), location.name.strip(), wh_name.strip()]
-        new_name = ' '.join(item for item in items if item)
-        return new_name
+        res = []
+        if not self._ids:
+            return res
+        wh_obj = self.env['stock.warehouse']
+        wh_dict = {}.fromkeys(self._ids, False)
+        query = self._cr.mogrify('''
```
moylop260: What about use directly the execute (without mogrify) because supports this case?

stock_location_code/model/stock.py
```diff
@@ -65,21 +65,44 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
+                AND sl.id IN %(ids)s
+            GROUP BY sl.id
+        ''', {'ids': tuple(self._ids)})
+        self._cr.execute(query)
+        wh_dict.update(
+            dict((l, wh_obj.browse(w[0]))
+                 for l, w in self._cr.fetchall()))
+        for location in self:
+            barcode = "[%(barcode)s]" % {'barcode': location.loc_barcode} \
+                if location.loc_barcode else ''
```
moylop260: if there is not barcode you will have the %
Is not?

stock_location_code/model/stock.py
```diff
@@ -65,21 +65,44 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
+            # TODO: Add fields.function to get warehouse. get_warehouse is too
+            # slow
+            # TODO: Add a parameters for a location path name or a wh_name
+            warehouse = wh_dict[location.id]
+            wh_name = "(%(warehouse)s)" % {'warehouse': warehouse.name} \
```
moylop260: same case here

stock_location_code/model/stock.py
```diff
@@ -65,21 +65,44 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
-        return new_name
+        res = []
+        if not self._ids:
+            return res
+        wh_obj = self.env['stock.warehouse']
+        wh_dict = {}.fromkeys(self._ids, False)
+        query = self._cr.mogrify('''
+            SELECT
+                sl.id AS location_id,
+                ARRAY_AGG(sw.id) AS warehouse_id
```
moylop260: here you have a cumulative of ids
But in the code you are using just one id
Maybe you can use a limit...
oh wait
I dont know how to limit the agregator

forget it

moylop260 (commented):
could add a print screen (using private url from google drive incorder to avoid show private info) to see the locations with a production database?

stock_location_code/model/stock.py
```diff
@@ -65,21 +65,44 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
+            FROM stock_location sl, stock_location sl_wh
+            INNER JOIN stock_warehouse sw ON sw.view_location_id = sl_wh.id
+            WHERE
+                sl_wh.parent_left <= sl.parent_left
+                AND sl_wh.parent_right <= sl.parent_left
+                AND sl.id IN %(ids)s
+            GROUP BY sl.id
+        ''', {'ids': tuple(self._ids)})
+        self._cr.execute(query)
+        wh_dict.update(
```
moylop260: What about use this logic for get_warehouse method?
in order to compute faster the other pieces of code where this method is involed

stock_location_code/model/stock.py
hbto: I just wanted to have the query to be executed prior to execution only for debugging purposes, though having the mogrify in there is not wrong either.

stock_location_code/model/stock.py
hbto: You just realized while writing the comment

stock_location_code/model/stock.py
hbto: You meant to prepare a PR to odoo/odoo @ saas-18 or any other version?

hbto:
> could add a print screen (using private url from google drive incorder to avoid show private info) to see the locations with a production database?

(Link share as private Vauxoo People only)
https://drive.google.com/a/vauxoo.com/file/d/0B1w9gokcJL9hU28teFBZVGVqTHc/view?usp=sharing

hbto:
> What about use this logic for get_warehouse method?
> in order to compute faster the other pieces of code where this method is involed

If an improvement is done on  https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_location.py#L114-L119

that would mean that the improvement is done over one single record rather than several records 
as done here:
```
        query = self._cr.mogrify('''
            SELECT
                sl.id AS location_id,
                ARRAY_AGG(sw.id) AS warehouse_id
            FROM stock_location sl, stock_location sl_wh
            INNER JOIN stock_warehouse sw ON sw.view_location_id = sl_wh.id
            WHERE
                sl_wh.parent_left <= sl.parent_left
                AND sl_wh.parent_right <= sl.parent_left
                AND sl.id IN %(ids)s
            GROUP BY sl.id
        ''', {'ids': tuple(self._ids)})
```
so the improvement that can be proposed to 
https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_location.py#L114-L119
would be:
```
        query = self._cr.mogrify('''
            SELECT
                sw.id AS warehouse_id
            FROM stock_location sl, stock_location sl_wh
            INNER JOIN stock_warehouse sw ON sw.view_location_id = sl_wh.id
            WHERE
                sl_wh.parent_left <= sl.parent_left
                AND sl_wh.parent_right <= sl.parent_left
                AND sl.id IN %(ids)s
            LIMIT 1
        ''', {'ids': tuple(self._ids)})
```

Please advice if any other idea arises, or if I got wrong your idea

stock_location_code/model/stock.py
moylop260: you can have the query using print cr.query avoiding use mogrify

stock_location_code/model/stock.py
moylop260: Forget it
it is a native module for a deprecated version

hbto:
@moylop260 mogrify has gone

PR #1238, opened 2017-10-22
[MIG] Preparing v11.0 cleanup (context lodi).

nhomar:
Every Commit will say why the change once finish I update this description:

cc @JulioSerna @hbto @dsabrinarg @josemoralesp


account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.multi
+    def _is_policy_ok(self, vals):
+        account = ('account_id' in vals) and self.env[
+            'account.account'].browse(vals['account_id'])
```
hbto: `account = self.env['account.account'].browse(vals['account_id']) if ('account_id' in vals) else self.account_id`

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+        analytic = ('analytic_account_id' in vals) and self.env[
+            'account.account'].browse(vals['analytic_account_id'])
+        if not account:
+            # When write I will not have the account on vals if the account
+            # is actually not modified
+            account = self.account_id
+        policy = account and account.analytic_policy
+        if not policy or not account:
```
hbto: `or not account`
Is it a valid assertion?

You will not reach this state as account_id is a required field on Journal Items.
https://github.com/odoo/odoo/blob/11.0/addons/account/models/account_move.py#L418

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+    @api.multi
+    def _is_policy_ok(self, vals):
+        account = ('account_id' in vals) and self.env[
+            'account.account'].browse(vals['account_id'])
+        analytic = ('analytic_account_id' in vals) and self.env[
+            'account.account'].browse(vals['analytic_account_id'])
+        if not account:
+            # When write I will not have the account on vals if the account
+            # is actually not modified
+            account = self.account_id
```
hbto: It will not be needed as above it is properly grabbed.

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+        policy = account and account.analytic_policy
```
hbto: `policy = account.analytic_policy`

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+        if not policy or not account:
+            return True
```
hbto: `if not policy or policy == 'optional':`
Return early `by_nhomar`

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.multi
+    def _is_policy_ok(self, vals):
+        account = ('account_id' in vals) and self.env[
+            'account.account'].browse(vals['account_id'])
+        analytic = ('analytic_account_id' in vals) and self.env[
+            'account.account'].browse(vals['analytic_account_id'])
```
hbto: `analytic = self.env['account.analytic.account'].browse(vals['analytic_account_id']) if ('analytic_account_id' in vals) else self. analytic_account_id`

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+            # When write I will not have the account on vals if the account
+            # is actually not modified
+            account = self.account_id
+        policy = account and account.analytic_policy
+        if not policy or not account:
+            return True
+        if not analytic and not ('analytic_account_id' in vals):
+            # When write I will not have the account on vals if the account
+            # is actually not modified
+            analytic = self.analytic_account_id
```
hbto: It will not be needed as above it is properly grabbed.

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+             '"Never": the accountant will get an error message if an '
+             'analytic account is present.'
+    )
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.multi
+    def _is_policy_ok(self, vals):
```
hbto: I suggest the name of the method to be changed to `_is_analytic_policy_ok`

In the future there could be other policies to be implemented on Journal Items then the name could clash with this one.

In the order hand, as your are the first one to name the method and the first one to implement the policy you have the right to name as you wish, other just have to adapt to use other names.

http://www.funtrivia.com/askft/Question122991.html

account_analytic_policy/README.rst
```diff
@@ -0,0 +1,10 @@
+Account Analytic Required
+=========================
+
+This module adds an option "analytic policy" on Accounts.
+
+You have the choice between 3 policies : always, never and optional. For
+example, if you want to have an analytic account on all your expenses, set
+the policy to "always" for the account that represent the "expense" ; then, if
+you try to save an account move line with an account of type "expense" without
+analytic account, you will get an proper error message.
```
hbto, line 10: Please. Do not create a new file if there is a previous file existing.
Move the previous one to the new file name. `git move`

Preserve history while migrating modules, by doing this you are killing this module history track.

hbto (changes requested):

account_analytic_policy/model/account_move_line.py
```diff
@@ -0,0 +1,72 @@
...
+        if not analytic and not ('analytic_account_id' in vals):
+            # When write I will not have the account on vals if the account
+            # is actually not modified
+            analytic = self.analytic_account_id
+        if policy == 'never' and analytic:
+            raise UserError(_('You cannot use analytic accounts on a move line'
+                              ' with Analytic Policy "Never" contact the '
+                              'account manager or fix the account '
+                              'configuration or pick another account.\n\n') +
+                            '%s' % account.name_get())
```
hbto: _mig__preparing_v11_0_cleanup__context_lodi___by_nhomar_ _pull_request__1238_ _vauxoo_addons-vauxoo

account_analytic_policy/README.rst
nhomar: I did a git mv, but as the chancge of the readme was in too many lines it is recognized as a new one. (it is a frequent situation).

account_analytic_policy/model/account_move_line.py
nhomar: I prefer with tern I do not like inside if (matter of PoV),  do you have a better reason not opinionated for this change?

account_analytic_policy/model/account_move_line.py
nhomar: I need to check first the key is in the dict if not the vals['analytic_account_id'] will brake.

account_analytic_policy/model/account_move_line.py
nhomar: I do not get it.

account_analytic_policy/model/account_move_line.py
nhomar: It is checked because this is a create before the check on the required part, odoo does on that way in account.

account_analytic_policy/model/account_move_line.py
nhomar: account can be False

hbto:
+1

PR #1239, opened 2017-10-23
[FIX] Fixing conditions & aggregations in SQL Query

hbto:
[FIX] There is no need for aggregation. Parent right ought to be greater
than parent left on queried location. Dictionary is for unique locations.


sabrinagnu:
👍

PR #1253, opened 2017-12-24
[8.0][FIX] Added new tolist() method to use Int64Index object returned by he new version of pandas

josemoralesp:


josemoralesp:
@hbto Do you agree with these changes?

hbto:
@josemoralesp I agree on the changes, +1

PR #1258, opened 2018-01-25
[ADD] mrp_version: Add module for post inventory from work order and change routing in production order..

egrivero:


egrivero:
@nhomar This PR is ready for review

luisg123v:
This is for an old version, has conflicts and there's no activity from years ago, so I will close it.

PR #1264, opened 2018-02-20
[FIX] account_anglo_saxon_stock_move_purchase: added api model decora… task#23901

keylor2906:
…tor to cron method

The error caused when not using the decorator:
image


keylor2906:
@josemoralesp could you review/merge this pr please?

PR #1267, opened 2018-03-28
[ADD] email_show_only_changes: this module shows only the values chan…

keylor2906:
…ged of an object on email, instead of all tracked values, even if it has track_visibility set as always

Before:
image

After:
image


keylor2906:

josemoralesp (approved):

PR #1273, opened 2018-05-10
[FIX](account_currency_tools) Wrong Journal Entries for Payables / Receivable

hbto:
[FIX] Journal Entries for Payables / Receivable were computed summing up
previous values in account where Profit/Loss was recorded.


josemoralesp:
:+1:

PR #1277, opened 2018-06-01
[FIX] account_move_report:  Print time corrected according to time zone in report. vx#25102

deivislaya:
When you check any book entry or click on the seat

The Print option appears and shows you the option:

* Account Move Report

- When selecting any of the report, the system generates a PDF where the print time shown is not correct

**TODO**
- [x] Modified Account Move Report


deivislaya:
Hi @JulioSerna

This PR is ready for review

PR #1279, opened 2018-06-15
[MIG] account_asset_date: Migration to V11.0

luiseevaquer:
dummy LODI: https://git.vauxoo.com/vauxoo/lodi/merge_requests/16

Enable dependency: https://git.vauxoo.com/vauxoo/lodi/merge_requests/25


dsabrinarg:
DUMMY [#15](https://git.vauxoo.com/vauxoo/lodi/merge_requests/15)

luiseevaquer:
@josemoralesp 

Please, Can you do technical review?

account_asset_date/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# coding: utf-8
+{
+    "name": "Account Asset Date",
+    "version": "11.0.0.1.6",
```
luistorresm: Why do you not restart to 11.0.1.0.0?

account_asset_date/i18n/es_MX.po
```diff
@@ -0,0 +1,37 @@
+# Translation of OpenERP Server.
```
luistorresm: This file is not required.

Only the es.po and es.pot

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,166 @@
...
+
+    different_dates = fields.Boolean('Set Start Depreciation Date',
+                                     default=False)
+
+
+class AccountAssetAsset(models.Model):
+    _inherit = 'account.asset.asset'
+
+    start_depreciation = fields.Date('Start Depreciation Date',
+                                     help=_('Start Depreciation Date'))
```
luistorresm: The help is redundant with the field name.

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,166 @@
...
+                vals['date'], vals['prorata']),
+        })
+        asset = super(AccountAssetAsset, self).create(vals)
+        return asset
+
+    def _start_depreciation_date(self, deprec_date, prorata=False):
+        if prorata:
+            return (datetime.strptime(deprec_date, DF) +
+                    relativedelta(days=1)).strftime("%Y-%m-%d")
+        else:
```
luistorresm: If you have a return inside of an if, you do not need the `else`sentence.

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,165 @@
...
+            lambda x: x.move_check).sorted(key=lambda l: l.depreciation_date)
+        unposted_depreciation_line_ids = self.depreciation_line_ids.filtered(
+            lambda x: not x.move_check)
+
+        # Remove old unposted depreciation lines. We cannot use unlink()
+        # with One2many field
+        commands = [(2, line_id.id, False) for line_id in
+                    unposted_depreciation_line_ids]
+
+        if self.value_residual != 0.0:
```
josemoralesp: You can use an early return and then move the code :

```python
if self.value_residual == 0.0:
    self.write({'depreciation_line_ids': commands})
    return True

```

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,165 @@
...
+            else:
+                # depreciation_date = 1st of January of purchase year
+                # if annual valuation, 1st of purchase month in other cases
+                if self.method_period >= 12:
+                    asset_date = datetime.strptime(
+                        self.start_depreciation[:4] + '-01-01', DF).date()
+                else:
+                    asset_date = datetime.strptime(
+                        self._start_depreciation_date(
+                            self.start_depreciation), DF).date()
```
josemoralesp:
```python
date = self.start_depreciation[:4] + '-01-01' if self.method_period >= 12 else self.start_depreciation
asset_date = datetime.strptime(self._start_depreciation_date(date), DF).date()

```

sabrinagnu:
Functional review 👍 @josemoralesp please review and if LGTY merge.

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,165 @@
...
+            deprec_date[5:7]) + 1)
+
+    @api.multi
+    def compute_depreciation_board(self):
+        self.ensure_one()
+        if not self.different_dates:
+            return super(AccountAssetAsset, self).compute_depreciation_board()
+
+        posted_depreciation_line_ids = self.depreciation_line_ids.filtered(
+            lambda x: x.move_check).sorted(key=lambda l: l.depreciation_date)
```
josemoralesp: I think a search could be better for this case. 
```python
posted_depreciation_line_ids = selfn.env['model'].search([('move_check', '=', True), ('m20', '=', self.id)], order='depreciation_date')
```

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,165 @@
...
+        commands = [(2, line_id.id, False) for line_id in
+                    unposted_depreciation_line_ids]
+
+        if self.value_residual != 0.0:
+            amount_to_depr = residual_amount = self.value_residual
+            if self.prorata:
+                # if we already have some previous validated entries, starting
+                # date is last entry + method perio
+                if (posted_depreciation_line_ids and
+                        posted_depreciation_line_ids[-1].depreciation_date):
```
josemoralesp: This condition is used twice. It could be better if we create a new variable(depreciation_date) to use it once.

```python
depreciation_date = posted_depreciation_line_ids and posted_depreciation_line_ids[-1].depreciation_date
```

josemoralesp (changes requested):
@lescobarvx 

I tried to remove as many duplicated code and 'else' statement I could.

If you can remove more, please don't hesitate.

Remember.

This:
```python
a = b if c else d
```
is faster than:
```python
if c:
    a = b
else:
    a = d
```

account_asset_date/models/account_asset.py
```diff
@@ -0,0 +1,152 @@
...
+        return True
+
+    @api.multi
+    def _get_last_depreciation_date(self):
+        if not self.different_dates:
+            return super(AccountAssetAsset, self)._get_last_depreciation_date()
+
+        self.env.cr.execute("""
+            SELECT a.id as id,
+              COALESCE(MAX(m.date),a.start_depreciation) AS date
```
josemoralesp, line 134: Sorry, @lescobarvx  I didn't see this before.
The date returned by this query is in UTC. Is there no problem?

cc. @dsabrinarg

luiseevaquer: @josemoralesp 

I copied from original method to overwrite it and they use that query

https://github.com/odoo/odoo/blob/11.0/addons/account_asset/models/account_asset.py#L123-L129

josemoralesp (approved):

PR #1286, opened 2018-06-27
[FIX] Fixed wrong field name in demo record

luiseevaquer:


luiseevaquer:
LODI Dummy https://git.vauxoo.com/vauxoo/lodi/merge_requests/27

PR #1287, opened 2018-07-04
[REF] product_cost_usd: standard_price_usd as compute field

suniagajose:
Summary
--------------

The field Cost in USD, now, is computed getting the price in USD recordered in vendor pricelist. In this way, is not necessary have constraint to ensure that at least a vendor pricelist in USD currency is loaded, if it isn't registered in the vendor pricelist, then, will be 0.0.

Preview
-----------

d3694770 rear inner cover duplex unit mp2550sp - odoo

d3694770 rear inner cover duplex unit mp2550sp - odoo 1

### DIFF
--- a/product_cost_usd/i18n/es.po
+++ b/product_cost_usd/i18n/es.po
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 10.0+e\n"
+"Project-Id-Version: Odoo Server 11.0+e\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-02-06 18:52+0000\n"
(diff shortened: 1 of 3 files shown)


luisg123v:
Hi @suniagajose,

Is this still outstanding?

suniagajose:
@luisg123v 

is a improvement but need to be approved by customers that use the module on this version...

maybe the right way is merge in v13.0 or a future odoo version where this module is not yet implemented by a customer

luisg123v:
Hi @suniagajose,

I don't think this will be used, given that it's been more than two years from its creation now. Hence, I'm going to close this PR.

I'll take this into account if this module gets ever migrated to v12 or above, though.

Of course, feel free to reopen it if you need it in the future, or if you consider it should remain opened.

PR #1295, opened 2018-10-16
[ADD] Added new module to compute the segmentation cost. This module …

keylor2906:
…adds a new model to get the history of the move to know the real cost of a move

[ADD] Added new module to compute segmentation cost recursively from BOM

[ADD] Added new module to a needed field to compute the segmentation cost

[ADD] Added new module to compute the segmentations at the moment to validate a landed record

[IMP] Added new cronjob to compute the cost every day at night


josemoralesp:
@hbto Could you please bless this?

hbto (approved):
👍  @hbto has approved these changes

PR #1299, opened 2018-10-18
 [MIG] email_show_only_changes: Migration to 12.0

hugho-ad:


hugho-ad:
@moylop260 

could you review this please?

email_show_only_changes/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# coding: utf-8
```
moylop260: remove coding

email_show_only_changes/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+{
+    "name": "Email Show Only Changes",
+    "version": "12.0.0.0.1",
```
moylop260: `12.0.1.0.0`

email_show_only_changes/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "author": "Vauxoo",
+    "category": "",
```
moylop260: Use a category

email_show_only_changes/models/mail_thread.py
```diff
@@ -0,0 +1,34 @@
+from odoo import models, api
+
+
+class MailThread(models.AbstractModel):
+    _inherit = 'mail.thread'
+
+    @api.multi
+    def _message_track(self, tracked_fields, initial):
+        """ For a given record, fields to check (column name, column info)
```
moylop260: Strip first space

moylop260 (commented):
- Add a README
- Check if there is a OCA project that we can use to added it there.

hugho-ad:
@moylop260 Done

email_show_only_changes/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Email Show Only Changes",
+    "version": "12.0.0.0.0",
```
moylop260: 12.0.1.0.0

email_show_only_changes/models/mail_thread.py
```diff
@@ -0,0 +1,34 @@
...
+    @api.multi
+    def _message_track(self, tracked_fields, initial):
+        """For a given record, fields to check (column name, column info)
+        and initial values, return a structure that is a tuple containing :
+
+        - a set of updated column names
+        - a list of changes (old value, new value, column name, column info)
+        """
+        self.ensure_one()
+        changes = super(MailThread, self)._message_track(
```
moylop260, line 15: Is possible a IndexError here?

email_show_only_changes/models/mail_thread.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @api.multi
+    def _message_track(self, tracked_fields, initial):
+        """For a given record, fields to check (column name, column info)
+        and initial values, return a structure that is a tuple containing :
+
+        - a set of updated column names
+        - a list of changes (old value, new value, column name, column info)
+        """
+        self.ensure_one()
```
moylop260: The super has the ensure_one then you dont need twice times
is not?

moylop260 (commented):
You are missing looking for a OCA project to create PR

email_show_only_changes/models/mail_thread.py
hugho-ad: No, it not possible because _messages_track always return a tuple

email_show_only_changes/models/mail_thread.py
hugho-ad: It's added just to be consistent across the inheritance.

hugho-ad:
@moylop260 

About the pr to an oca project, I was thinking in server-tools, what do u think?

Also, Why this module can't be only at addons-vauxoo?

Regards!

email_show_only_changes/models/mail_thread.py
moylop260: Ok I got it
Thank you.

Just for the record,
Completing the answer
The method is here:
https://github.com/odoo/odoo/blob/b81c2bce846c94b5695d680ebbc30f1cfe51e614/addons/mail/models/mail_thread.py#L596-L620

email_show_only_changes/models/mail_thread.py
moylop260: If you use
```python
def inherited_method(self):
    self.ensure_one()
    self.something_before_super()
    ....
```
You have a point to use ensure_one

But if you use:
```python
def inherited_method(self):
    super().inherited_method()  # First line of this method is ensure_one
    ....
```
So you don't need add twice times the ensure_one

moylop260:
What about https://github.com/OCA/server-ux?

Let's to create the PR to server-ux and wait feedback about.

Let me know the PR.

> Also, Why this module can't be only at addons-vauxoo?

Because exists a community that we are supporting and they are supporting us.

moylop260:
Another reason is that addons-vauxoo is not executing tests, then if we find a home in OCA for this module we will have a community supporting it, we will have a coverage report, runbot, travis tests, lints... and so on.

hugho-ad:
@moylop260 
Cool! thanks for the feedback, I'll make the pr to server-ux

Regards!

hugho-ad:
PR https://github.com/Vauxoo/server-ux/pull/3

moylop260:
Please, uses https://github.com/OCA/server-ux project instead of Vauxoo one.
Like my last comment:
https://github.com/Vauxoo/addons-vauxoo/pull/1299#issuecomment-431886623

hugho-ad:
https://github.com/OCA/server-ux/pull/40

PR #1300, opened 2018-10-23
[ADD] invoice_split: Add module to split customer invoices

luisg123v:
This module allows to split a customer invoice, so the customer may pay
a percentage of the invoice amount and, optionally,  create a new
invoice to pay the remaining amount later.

For more info, see the README file.

Dummy MR: https://git.vauxoo.com/absa/absa/merge_requests/75

### DIFF
--- a/invoice_split/README.md
+++ b/invoice_split/README.md
@@ -1,0 +1,47 @@
+Split Invoices
+==============
+
+This module allows to split a customer invoice, so the customer may pay a
+percentage of the invoice amount and, optionally,  create a new invoice
+to pay the remaining amount later.
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+1. Go to any invoice in draft mode
+2. Click on the menu Actions → "Split Invoice"
+3. Fill the percent by which the invoice will be split, and check/uncheck if a
+   new invoice will be created for the  remaining amount
+4. Click on the "Split" button. When a new invoice is created, you'll be
+   redirected to a list view where both invoices are displayed
+
+
+Considerations
+==============
+
+When invoices are split, their lines are copied and product quantities are
+divided which may produce non-integer quantities. This process is done without
+taking into account unit of measures. That means, if a line contains, for
+instance, three shirts, and the invoice is split to 50%; then the invoice will
+be for 1.5 shirts after the invoice is split, even though that is not
+possible in practice.
+
+
+Credits
+=======
+
+**Contributors**
+
+* Luis González <lgonzalez@vauxoo.com> (Developer)
+* Yanina Aular <yanina.aular@vauxoo.com> (Planner/Auditor)
+
+
+Maintainer
+==========
+
+.. image:: https://www.vauxoo.com/logo.png
+   :alt: Vauxoo
+   :target: https://vauxoo.com
--- a/invoice_split/wizards/invoice_split.py
+++ b/invoice_split/wizards/invoice_split.py
@@ -1,0 +1,80 @@
+# Copyright 2018 Vauxoo
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class AccountInvoiceSplit(models.TransientModel):
+    _name = "account.invoice.split"
+    _description = "Invoice Splitter"
+
+    invoice_id = fields.Many2one('account.invoice', String="Invoice to Split")
+    new_invoice_id = fields.Many2one('account.invoice')
+    percent = fields.Float(
+        string="Percent to Pay",
+        help="Percent of the amount that will be paid on the current invoice")
+    create_invoice = fields.Boolean(
+        string="Create New Invoice", default=True,
+        help="Specifies whether a new invoice will be created in draft mode "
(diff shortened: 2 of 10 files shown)


JulioSerna:
@luisg123v @yaniaular 

Is there an UH related to this PR?

luisg123v:
Hi @JulioSerna  
This is related to the [task#27639](https://www.vauxoo.com/web#id=27639&view_type=form&model=project.task&action=464&active_id=9&menu_id=317)

luisg123v:
@yaniaular  
This is ready for review.

invoice_split/README.md
```diff
@@ -0,0 +1,36 @@
...
+percentage of the invoice amount and, optionally,  create a new invoice
+to pay the remaining amount later.
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+1. Go to ani invoice
```
yaniaular: @luisg123v ani -> any
```suggestion
1. Go to any invoice
```

invoice_split/wizards/invoice_split.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class AccountInvoiceSplit(models.TransientModel):
+    _name = "account.invoice.split"
+    _description = "Invoice Splitting"
+
+    invoice_id = fields.Many2one('account.invoice', String="Invoice to Split")
+    new_invoice_id = fields.Many2one('account.invoice')
+    percent = fields.Float(
+        string="Percent to Pay", default=50.0,
```
yaniaular: @luisg123v 
```suggestion
        string="Percent to Pay",
```

invoice_split/wizards/invoice_split.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record id="split_wizard" model="ir.ui.view">
+        <field name="name">split.invoice.wizard</field>
+        <field name="model">account.invoice.split</field>
+        <field name="arch" type="xml">
+            <form string="Split Invoice">
+                <group>
+                    <field name="percent"/>
```
yaniaular, line 10: @luisg123v 
```suggestion
                    <field name="percent" required="1"/>
```

yaniaular:
@luisg123v Add translations please

invoice_split/wizards/invoice_split.py
```diff
@@ -0,0 +1,46 @@
...
+        help="Percent of the amount that will be paid on the current invoice")
+    create_invoice = fields.Boolean(
+        string="Create New Invoice", default=True,
+        help="Specifies whether a new invoice will be created in draft mode "
+        "with the remaining uncovered amount")
+
```
yaniaular, line 21: @luisg123v Check if the invoice is in draft state, something like this:
```suggestion
    @api.model
    def default_get(self, fields):
        rec = super(AccountInvoiceSplit, self).default_get(fields)
        active_ids = self.env.context.get('active_ids')
        active_model = self.env.context.get('active_model')

        # Check for selected invoices ids
        if not active_ids or active_model != 'account.invoice':
            return rec

        invoices = self.env['account.invoice'].browse(active_ids)

        # Check invoices are draft
        if any(invoice.state != 'draft' for invoice in invoices):
            raise UserError(_("You can only split draft invoices"))
```

luisg123v:
Done @yaniaular, translations were included.

invoice_split/wizards/invoice_split.xml
luisg123v: This is a numeric field, `required` has no effect over it, because it always has a value. If you leave it empty, it will be set to zero.

invoice_split/wizards/invoice_split.py
luisg123v: Are you sure? If default value is not given, it takes one anyway, but it's zero.

luisg123v: We could remove the default and raise a warning if the number if out of range, e.g. 0.

invoice_split/wizards/invoice_split.xml
luisg123v: I added a validation to prevent this from being zero.

luisg123v:
@yaniaular  
All comments were applied or answered.

Note: 
For now, this is red because of [this issue](https://github.com/Vauxoo/addons-vauxoo/pull/1304#discussion_r227633903)

luisg123v:
@yaniaular  
This is not red anymore.

invoice_split/wizards/invoice_split.xml
yaniaular: Good. It's better

yaniaular:
@luisg123v The taxes in the splited invoices are not recalculating, could you check if [this method](https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L732-L739) is called correctly in your method?  please

luisg123v:
@yaniaular   done.

I also improved the README and docstrings to explain te case of invoicing three units and splitting by 50%, where 1.5 of any product  will be  invoiced, even if it's not possible in practice.

luisg123v:
We found this module to not be generic enough to be here (at least for now), hence I'm closing this.

PR #1302, opened 2018-10-23
[REF] .travis.yml: Use Python 3.6 for Odoo 12.0

luisg123v:
Odoo 12.0 should use Python 3.6, instead of Python 3.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,7 @@ cache:
     - $HOME/.pip-cache/
 
 python:
-  - "3.4"
-
-virtualenv:
-  system_site_packages: true
+  - "3.6"
 
 env:
   global:


luisg123v:
@moylop260  
Could you merge this, please?

.travis.yml
```diff
@@ -7,7 +7,7 @@ cache:
     - $HOME/.pip-cache/
 
 python:
-  - "3.4"
+  - "3.6"
 
 virtualenv:
   system_site_packages: true
```
moylop260: The CI is failing because travis is not supporting this system_site_packages for 3.6

luisg123v: You're right, thanks!. Done.

Now it's green.

PR #1304, opened 2018-10-23
[REF] *: General cleanup

### LINKED ISSUE #1301: .travis.yml needs some clean up
The file `.travis.yml` has some code that, AFAIK, is unnecessary. For instance, [this code to determine changed modules](https://github.com/vauxoo/addons-vauxoo/blob/12.0/.travis.yml#L41). That is natively done by MQT.

CC @moylop260

luisg123v:
This performs the following changes:
- Update code of the .travis.yml and remove deprecated/unnecessary code
- Fix small typos on the file CONTRIBUTING.md and remove not-useful
  section
- Remove file pseudo_oca_dependencies and replace it by an empty
  oca_dependencies.txt

Closes #1301


luisg123v:
@moylop260  
What do you think?

.travis.yml
```diff
@@ -4,46 +4,42 @@ sudo: false
...
   apt: true
   directories:
-    - $HOME/.pip-cache/
+    - $HOME/.cache/pip
+
+addons:
+  postgresql: "9.6"
+  apt:
+    packages:
+      - expect-dev
```
moylop260: All these packages are not used anymore since that we are using pip wheels with backup of cache directory

luisg123v: @moylop260  
Done. I left apt dependencies as they are, without changes.

luisg123v: @moylop260 
You removed all packages, but I think those were necessary. Travis is failing now (see #1300's log [here](https://travis-ci.org/Vauxoo/addons-vauxoo/jobs/445457431#L958)).

It's failing to run unbuffer because the package  `expect-dev` is not installed.

moylop260: You are right.
Let’s add that package again

PR #1305, opened 2018-10-23
[REF] product_extended_segmentation: change xpath to view because it … task#27478

keylor2906:
…was raising error when using a different language than english

image


keylor2906:
@josemoralesp could you review this, please?

PR #1307, opened 2018-10-26
[REF] product_extended_segmentation: When using the wizard to change …

keylor2906:
…cost, change all segmented costs of product and its components

### DIFF
--- a/product_extended_segmentation/__manifest__.py
+++ b/product_extended_segmentation/__manifest__.py
@@ -16,6 +16,7 @@
     'data': [
         'data/ir_cron.xml',
         'views/res_config_settings_views.xml',
+        'views/stock_change_standard_price.xml',
     ],
     'demo': [
         'demo/demo.xml',
--- a/product_extended_segmentation/models/product.py
+++ b/product_extended_segmentation/models/product.py
@@ -3,7 +3,7 @@
 
 import logging
 from odoo.tools.float_utils import float_is_zero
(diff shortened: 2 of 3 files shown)


keylor2906:
@josemoralesp this PR is for recursivelly set the new costs for a product and all it's components using the wizard, also this fix the case that using the wizard was only setting the total cost and not all segments. The creation of the wizard during the calculation of segmented cost was split on a separate method because it will be inherited by lodi, because they will use a different account instead of the one that is already set, that is the one that odoo use by default.

josemoralesp:
Hello @keylor2906 

When we talked about this in the morning I was thinking about something like this:

This [here](https://github.com/Vauxoo/addons-vauxoo/tree/11.0/product_extended_segmentation/views).

```xml
 <xpath expr="//button[@name='compute_price']" position="attributes">
       <attribute name="context">{'from_wizard': False}</attribute>
</xpath>   
```

Then [here](https://github.com/Vauxoo/addons-vauxoo/blob/11.0/product_extended_segmentation/models/product.py#L165)

```python
 prod_costs_dict = self._calc_price_seg(child_bom_id, self.env.context.get('from_wizard', True))
```

Then you should return sgmnt_dict [here](https://github.com/Vauxoo/addons-vauxoo/blob/11.0/product_extended_segmentation/models/product.py#L245), with the key price.

What do you think?

What did you see that I couldn't see?

Best Regards.

PR #1308, opened 2018-10-29
[REF] purchase_third_validation: Remove "Third approval" state from b…

keylor2906:
…eing always visible on the widget, to only be visible if the purchase order is on that state. This because it was confusing. And this way it behaves like the state "To approve", that is the second validation state.

Without change:
image

With change:
image
image


keylor2906:

PR #1309, opened 2018-10-30
[FIX] stock_landed_segmentation: Fixed case when changing the invoice…

keylor2906:
…s on the landing cost, the cost lines where repeated as many times as lines were on the invoice.

[REF] stock_landed_segmentation: Using split method of the product on the creation of the cost line for a landed cost, instead of just using a default value.


keylor2906:

PR #1311, opened 2018-11-09
[FIX] stock_cost_segmentation: Ordered domain to get candidates with … task#29355

keylor2906:
…fifo logistic in move, this caused error division by 0 in other operation.


keylor2906:
@josemoralesp if this change is needed here is the pr

PR #1313, opened 2018-11-14
[REF] product_extended_segmentation: change the change price wizard t… task#27478

keylor2906:
…o save the segment costs before the standard price, this because if the standard price is changed before, the segemnts are never changed


keylor2906:
@josemoralesp

PR #1315, opened 2018-11-21
[REF] product_extended_segmentation: Change the calculation of the ma… task#6882

keylor2906:
…terial cost, instead of using the unit price of the stock move based on fifo, use the standard price of the product.


keylor2906:

PR #1318, opened 2018-12-06
[ADD] account_index_based_currency: Hold the rates of the currency involved in an invoice at validation & agreement time.

josemoralesp:


josemoralesp:
Hello @hbto 

I've added some changes.

1) I changed the way the _get_conversion_rate method is called because it works differently in this version.
2) The method _convert does not exist in this version. I've used the compute method instead.
3) I am using the context to get the rate using a specific date.

Do you agree with these changes?

hbto:
@josemoralesp I do agree

PR #1320, opened 2018-12-12
[FIX] Transaction Amount is field computed from the agreement rate

hbto:
[FIX] Transaction Amount is field computed from the agreement rate and amount.

Agreement amount is no longer a computed amount.

### DIFF
--- a/account_index_based_currency/models/account_invoice.py
+++ b/account_index_based_currency/models/account_invoice.py
@@ -27,23 +27,13 @@ class AccountInvoice(models.Model):
                     inv.amount_total_signed, inv.index_based_currency_id,
                     inv.company_id, inv.date_invoice or fields.Date.today())
             inv.index_based_currency_amount = index_based_currency_amount
-            inv.agreement_currency_amount = (
-                inv.index_based_currency_amount /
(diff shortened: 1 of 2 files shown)


yaniaular:
@moylop260 Can you review this please? I was tested yesterday and today with @hbto 

The plan is this.
image

If you want to test: <s>private info</s>
The credentials by hangout.

yaniaular (approved):

account_index_based_currency/models/account_invoice.py
```diff
@@ -93,6 +88,7 @@ def default_get(self, default_fields):
         copy=False, digits=(12, 6))
     agreement_currency_rate = fields.Float(
         help="Currency rate this Document was agreed",
+        default=lambda *args: 1,
```
moylop260: If this is not needed recomputed since that is a static value then you can use just `default=1`

hbto: Done

moylop260:
I just merged using squash button since that there are 2 commits and I used a custom message based on standard:
`[TYPE] module: msg`

`[FIX] account_index_based_currency: Transaction Amount is ...`

PR #1321, opened 2018-12-13
[REF] mrp_workcenter_account_move: Post journal entries that are crea…

keylor2906:
…ted at the end of the manufacturing production.


keylor2906:
@JulioSerna could you review this, please?

This is for the case of the journal entries of the cost production that were not being posted after being created. The post method is used after the move is created and the lines are assigned.

JulioSerna (approved):

sabrinagnu:
@josemoralesp merge please!

PR #1323, opened 2018-12-16
[REF] Better trust on currency_amount instead of rates

hbto:
[FIX] Fixing view for supplier invoices to bear the Transaction Amount
and making aware Transaction Amount of changes in Agreement Amount
Using agreement_currency_amount instead of rates.

### DIFF
--- a/account_index_based_currency/models/account_invoice.py
+++ b/account_index_based_currency/models/account_invoice.py
@@ -18,7 +18,7 @@ class AccountInvoice(models.Model):
                 inv.company_id, inv.date_invoice or fields.Date.today())
 
     @api.depends('amount_total_signed', 'agreement_currency_rate',
-                 'date_invoice')
+                 'agreement_currency_amount', 'date_invoice')
     def _compute_currency_amount(self):
         for inv in self:
             index_based_currency_amount = inv.amount_total_signed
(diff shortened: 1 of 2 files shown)


hbto:
@moylop260 could please merge.

Regards.

hbto:
@yaniaular be aware

account_index_based_currency/models/account_invoice.py
```diff
@@ -18,7 +18,7 @@ def _compute_currency_rate(self):
                 inv.company_id, inv.date_invoice or fields.Date.today())
 
     @api.depends('amount_total_signed', 'agreement_currency_rate',
-                 'date_invoice')
+                 'agreement_currency_amount', 'date_invoice')
```
moylop260, line 23: Shouldn't be both?
Since that you are using both to compute the values

hbto: Both are used, agreement_currency_amount & agreement_currency_rate,

`agreement_currency_rate` is in the line above. It seems you have missed.

Regards

moylop260:
FYI I just saw the `account_id` and `currency_rate` fields used from the onchange method.
Shouldn't be added to depends too?

PR #1327, opened 2018-12-27
[FIX] stock_cost_segmentation: Fix case of recursivity error when get…

keylor2906:
…ting the custom numbers of landing costs when there are moves with returns on returns. When returning a return move causes both to be origin and destiny for eachother. With the changes applied, the search of landed wont search on an origin move that is the return move of itself.


keylor2906:
@JulioSerna Could you test and review this, please?

The dev instance is http://test.typrefrigeracion.com:30010

JulioSerna (approved):

PR #1329, opened 2019-01-16
[REF] account_refund_early_payment: Change widget of create refunds t… task#29987

keylor2906:
…o not hide the filter of refund type when the invoice is partially paid, because without it it's not possible to create more than one refund of early payment.


keylor2906:
@dsabrinarg could you functionally review this, please?

This is the dev instance http://dev.lodigroup.com:30025

PR #1330, opened 2019-01-16
[REF] product_extended_segmentation: Improve error message at the upd… task#30607

keylor2906:
…ate of the standard price to show the error and not only a generic message.


keylor2906:

PR #1333, opened 2019-02-15
[FIX] stock_by_warehouse: Use normal fields without formating the amounts in order to allow js made correct comparations t31436

hugho-ad:
Before:
"1,896.00" > 0 === false
"896.00" >0 === true

After:
1896.00 > 0 === true
896.00 >0 === true

https://youtu.be/Im-cGVj3110


hugho-ad:
@JulioSerna 
could you merge please

PR #1334, opened 2019-02-20
[IMP] message_post_model: enable tracking using xml. proved in 10.0

caber:
The set_tracking function enables tracking using xml files:
<function name="set_tracking" model="ir.model" eval="[[],'account.invoice']"/>
There's also some minor fixes.


luisg123v:

PR #1338, opened 2019-03-05
[REF] product_extended_segmentation: Added the segmented cost calcula… task#31736

keylor2906:
…tion with sub products done in the same bom (byproducts)


keylor2906:
Dummy: https://git.vauxoo.com/vauxoo/lodigroup/merge_requests/866

PR #1345, opened 2019-03-29
[FIX] Consumable and service type products are excluded in qty2produce

frahikLV:
Consumable and service type products are removed from the bill of materials when are analyzed in the get_qty_availble_to_produce method.

This bug was discovered in the task#31810 of the ticket#5182.

The client tries to produce a product with bills of material that include service, so the function doesn't consider the service because this doesn't appear in the order, and the function always returns 0.

### DIFF
--- a/mrp_partial_production/models/mrp.py
+++ b/mrp_partial_production/models/mrp.py
@@ -49,10 +49,10 @@ class MrpProduction(models.Model):
                                           res.quantity_done)
 
         for line, line_data in lines:
+            if line.product_id.type in ('service','consu'):
+                continue
             lines_dict[line.product_id.id] += line_data['qty'] / quantity
-
-        qty = [(result[key] / val) for key, val in lines_dict.items()
-               if val]
+        qty = [(result[key] / val) for key, val in lines_dict.items() if val]
         return (float_round(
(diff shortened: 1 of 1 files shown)


frahikLV:
@keylor2906 Could review this?

mrp_partial_production/models/mrp.py
```diff
@@ -49,10 +49,11 @@ def get_qty_available_to_produce(self):
                                           res.quantity_done)
 
         for line, line_data in lines:
-            lines_dict[line.product_id.id] += line_data['qty'] / quantity
-
-        qty = [(result[key] / val) for key, val in lines_dict.items()
-               if val]
+            if line.product_id.type in ('service','consu'):
+                lines_dict.pop(line.product_id.id, None)
+            else:
```
keylor2906: @frahikLV please remove the else statement, instead add a continue inside the if

keylor2906 (changes requested):

mrp_partial_production/models/mrp.py
frahikLV: Fixed 👍

frahikLV:
@keylor2906 Done :+1:

PR #1349, opened 2019-05-17
[FIX] The wizard to compute historical values was not working as expected.

sabrinagnu:
[DUMMY](https://git.vauxoo.com/vauxoo/lodigroup/merge_requests/886)


sabrinagnu:
@keylor2906 may you please give me your review here?.

keylor2906 (approved):

PR #1350, opened 2019-05-17
[REF] account_asset_date: When the start date for the asset is differ…

keylor2906:
…ent and is marked the prorata option dont calculate using the prorata rules and calculate like normal


keylor2906:
@dsabrinarg Could you review this, please? Dummy MR https://git.vauxoo.com/vauxoo/quebec/merge_requests/38 Test instance http://dev.lodigroup.com:30005

sabrinagnu (approved):

PR #1352, opened 2019-06-03
[IMP] account_move_report: Improves in the module

luistorresm:
1. If the move currency is the same that the company, the field
   amount_currency is empty, then not must be showed the column.
2. If the move currency is the same that the company, the currency is
   empty, then the column not must be added because is empty the value.

### DIFF
--- a/account_move_report/views/account_move_report.xml
+++ b/account_move_report/views/account_move_report.xml
@@ -42,7 +42,9 @@
                                         <th class="text-right">Debit</th>
                                         <th class="text-right">Credit</th>
                                         <th>Analytic account</th>
-                                        <th class="text-right">Amount Currency</th>
+                                        <t t-if="o.line_ids.filtered('amount_currency')">
+                                            <th class="text-right">Amount Currency</th>
+                                        </t>
                                         <th>Currency</th>
                                         <th>Tax Account</th>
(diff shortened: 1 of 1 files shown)


luistorresm:
@moylop260 

Could you merge this PR, please?

Regards

account_move_report/views/account_move_report.xml
```diff
@@ -63,10 +65,12 @@
                                             <span t-field="l.credit"/>
                                         </td>
                                         <td><span t-field="l.analytic_account_id"/></td>
-                                        <td class="text-right">
-                                            <span t-field="l.amount_currency"/>
-                                        </td>
-                                        <td><span t-field="l.currency_id.name"/></td>
+                                        <t t-if="l.amount_currency">
```
luisg123v, line 68: @luistorresm    
What if `amount_currency` is zero?

luistorresm: Could you send me an example, please?

Regards

luisg123v: @luistorresm   
Yes. Moves corresponding to exchange difference entries could have 0 in `amount_currency`, but could have set `currency_id`.

AFAIK, the right way to know if an amount is on a different currency, is to check if `currency_id is set`

PR #1354, opened 2019-06-19
[FIX] account_anglo_saxon_stock_move: Fix error when canceling an inv…

keylor2906:
…oice because the name of a field is incorrect.


keylor2906:
@dsabrinarg Could you review/merge please?

dev instance http://dev.lodigroup.com:30013

PR #1355, opened 2019-07-08
[FIX] account_move_report: Show amount currency when it's 0

luisg123v:
In some cases, the amount of a journal entry could be in a different
currency from the company, but still be zero, like in exchange diference
entries. Currently, in such cases, the column is hidden.

To solve the above, the column is shown when the currency is set for the
journal entry.


luisg123v:
@luistorresm  
Could you approve, please?

luistorresm (approved):
👍

luisg123v:
@moylop260  
Could you merge, please?

CC @edy1192

PR #1356, opened 2019-07-18
[REF] product_extended_segmentation: The cron that revaluates costs o… task#33968

keylor2906:
…nly considers products that have bom and that have cost_method standard, so fifo and average products are not considered.


keylor2906:
@josemoralesp Could you review/merge this, please?

PR #1357, opened 2019-07-18
[FIX] product_extended_segmentation: The cost_method field is a compu…

keylor2906:
…ted field and cant be used on search method, but can be used on filtered.


keylor2906:
@josemoralesp Forgive me for being dumb, the cost_method field is a computed field and cant be used on the search method, so using search without it and then using filtered.

PR #1358, opened 2019-07-20
[IMP] stock_by_warehouse_sale: Fix big performance overhead

moylop260:
[Try me on Runbot](https://runbot.vauxoo.com/runbot/85/671)

# Current behavior before PR:

We need avoid using compute fields as possible because of the following issue:
 - https://github.com/odoo/odoo/issues/30578
warehouses_stock is a computed field used just one time but re-computed too many times:
 - Opening sale order (one time by each line)
 - Saving sale order
 - Choosing a product_id
And it computation is more and more slow when the database grows.
It is making a big overhead saving sale.order.

Before of this commit editing just one product:
 - Screen Shot 2019-07-21 at 10 09 00 PM

production pyflame profiling the black rectangles are of `stock_by_warehouse...` modules:
 - Screen Shot 2019-07-21 at 11 45 23 PM

# Desired behavior after PR is merged:

With this change we have the same behaviour using it just for onchange
product_id or forcing from a seudo-button (field boolean)
A real button is not used since that it requires save the record to work

Using the technical of fields with store=False and onchange methods
Dummy PR:
 - https://git.vauxoo.com/vauxoo/typ/merge_requests/671

Check the following video testing it:
 - [IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=3rmx0B2GQrQ)

# Considerations:
We have changed the computed to field computed on-demand or product_id onchange then we have a little change of functionality.
1) For sale and purchase form view the widget data is filled from onchange but If you open an exists sale.order.line and you like to fill the widget data (without onchange) then you need use `EDIT` button and press the `new button` to fill the widget data.
 - Screen Shot 2019-07-21 at 11 27 52 PM
2) For product form view the widget is not auto-computed, then you will need use the `new button` to fill widget data.
  - Screen Shot 2019-07-21 at 11 26 53 PM

### DIFF
--- a/stock_by_warehouse/models/product.py
+++ b/stock_by_warehouse/models/product.py
@@ -18,8 +18,17 @@ class ProductTemplate(models.Model):
     )
 
     warehouses_stock = fields.Text(
-        compute='_compute_get_quantity_warehouses_json'
+        store=False, readonly=True,
     )
+    warehouses_stock_recompute = fields.Boolean(store=False, readonly=False)
+
+    @api.onchange('warehouses_stock_recompute')
+    def _warehouses_stock_recompute_onchange(self):
+        if not self.warehouses_stock_recompute:
+            self.warehouses_stock_recompute = True
+            return
+        self.warehouses_stock = self._compute_get_quantity_warehouses_json()
+        self.warehouses_stock_recompute = True
 
     @api.multi
     @api.depends('product_variant_ids.qty_available_not_res')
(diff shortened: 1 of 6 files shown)


moylop260:
@JulioSerna @hugho-ad @suniagajose @nhomar 
Could you check it, please?

JulioSerna:
👍 LGTM

the only change of behavior in the original way is that the button(toggle_button) must be pressed when the sale order is already created if you want to see the current inventory in that sale order line

hugho-ad:
Grate Job Moy
LGTM :+1:

luisg123v:
@moylop260  
I understand why this solution.. However, I don't fully agree with, you have more context for sure, tough.

The problem:  
Since you're relying on an onchange to compute the widget value, it only works when the record is on edit mode. That's because onchanges are not triggered otherwise.

Proposed solution:  
AFAIK, the boolean field is not even required. A button could trigger the action  to recompute, without needing any field:

```
<button name="compute_warehouses_stock"
    type="object"
    icon="fa-circle"
    groups="stock.group_stock_multi_warehouses"/>
```

Did you consider that approach?

CC @yaniaular

moylop260:
If you are talking about product form view, I agree!
In fact, it were a pending task.

But for [sale|purchase]_order is better the onchange

Could you create the PR for 11.0, please?

moylop260:
@luisg123v 
Could you check if is possible get the computed value without `edit` permission for product form view, please?

PR #1360, opened 2019-07-24
[REF] mrp_workcenter_segmentation: Fix runbot duplicated journal

keylor2906:


keylor2906:
@josemoralesp 

Don't know why the previous PR fixed the error on the build but now has error on the runbot, and the error don't even looks logical, because the error says that the key (code, name, company_id)=(vxpro, Mrp Routing Journal (Demo), 1) already exists but on the data it has a different code vxprod1. That's why I'm chaning the code and name

image

PR #1361, opened 2019-07-26
[FIX] product_extended_segmentation: On the calculation of material c…

keylor2906:
…osts make the calculation on base of the line qty instead of qty on bom.


keylor2906:
@josemoralesp Could you review/merge this, please?

PR #1362, opened 2019-07-29
[FIX] stock_by_warehouse[_sale|_purchase]: Get quantity even if there is a warning

moylop260:
DUMMY: https://git.vauxoo.com/vauxoo/typ/merge_requests/684


JulioSerna:
👍

PR #1364, opened 2019-08-01
[FIX] stock_by_warehouse: Using self._origin from top of methods and fix duplicated quantity from quants

moylop260:
DUMMY:
 - https://git.vauxoo.com/vauxoo/typ/merge_requests/688


JulioSerna:
👍

PR #1365, opened 2019-08-07
[REF] product_extended_segmentation: When a bom line has a bom and st…

keylor2906:
…andard cost, correctly call the method to calculate and change cost based on that line product and not using self.


keylor2906:
@josemoralesp @dsabrinarg Could you review/merge this, please?

PR #1368, opened 2019-08-30
Improve the advance_payment with Default Account and Domain

fanzalika:
1. The Default of Advance Account is improved by referring the default payable and
receivable account when Advance Account is not setup.
2. Add the domain filter in the field of advance_account_id so that we could limit
the account thru account_journal.


luisg123v:

PR #1370, opened 2019-10-16
[FIX] account_currency_tools: Fix case of using non multi currency ac…

keylor2906:
…count.


sabrinagnu:
@hbto may you please review/merge this?.

PR #1371, opened 2019-10-18
[IMP] partner_credit_limit: Improve the module for multicompany

edy1192:
- For instances of multiple companies, the credit limit value or days of grace
does not differentiate between company, so if the customer has a credit limit
`100` with `company A`, that same value will be in the `company B`.

- In some cases these values changes according to the company.

- With this change the `account.move.line` are searched according to the
company and additionally the credit limit fields and grace days now
depend on the company.

- Migration script to set the old values for the `credit_limit` and
`grace_payment_days` fields.

Videos in the comments: 

- https://youtu.be/d9p8Hs4JkiQ
- https://youtu.be/7Qyi-SSNMt8 
- https://youtu.be/z6quTSznYpA
- https://youtu.be/LeavM1fmDJ8

### DIFF
--- a/partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
+++ b/partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
@@ -1,0 +1,33 @@
+from odoo import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    set_credit_limit_values_to_partner(cr)
+
+
+def set_credit_limit_values_to_partner(cr):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env.cr.execute(
+        """
+        SELECT id, credit_limit, grace_payment_days FROM res_partner
+        WHERE credit_limit != 0 OR grace_payment_days != 0""")
+    data = env.cr.fetchall()
+    for company in env['res.company'].search([]):
+        for partner_id, credit_limit, grace_payment_days in data:
+            env['res.partner'].browse(partner_id).with_context(
+                force_company=company.id).write({
+                    'credit_limit': credit_limit,
+                    'grace_payment_days': grace_payment_days})
+    env['ir.logging'].create({
+        'name': 'partner_credit_limit migration',
+        'level': 'info',
+        'type': 'server',
+        'dbname': env.cr.dbname,
+        'message': data,
+        'path': 'migration script',
+        'line': 'migration script',
+        'func': 'migration script'
+    })
+    env.cr.execute("""ALTER TABLE res_partner DROP COLUMN credit_limit""")
+    env.cr.execute("""ALTER TABLE res_partner
+                   DROP COLUMN grace_payment_days""")
--- a/partner_credit_limit/model/partner.py
+++ b/partner_credit_limit/model/partner.py
@@ -8,9 +8,10 @@ from odoo import models, fields, api
 class ResPartner(models.Model):
     _inherit = 'res.partner'
 
+    credit_limit = fields.Float(company_dependent=True)
     grace_payment_days = fields.Float(
         'Days grace payment',
-        help='Days grace payment')
+        help='Days grace payment', company_dependent=True)
 
     credit_overloaded = fields.Boolean(
         compute='_get_credit_overloaded',
@@ -40,7 +41,10 @@ class ResPartner(models.Model):
 
     @api.model
     def movelines_domain(self, partner):
+        """Return the domain for search the
+        account.move.line for the user's company."""
         domain = [('partner_id', '=', partner.id),
+                  ('company_id', '=', self.env.user.company_id.id),
                   ('account_id.internal_type', '=', 'receivable'),
                   ('move_id.state', '!=', 'draft'),
                   ('reconciled', '=', False)]
--- a/partner_credit_limit/__manifest__.py
+++ b/partner_credit_limit/__manifest__.py
@@ -2,7 +2,7 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 {
     "name": "Partner Credit Limit",
-    "version": "11.0.0.0.1",
+    "version": "11.0.0.0.2",
     "author": "Vauxoo",
     "category": "",
     "website": "http://www.vauxoo.com/",
--- a/partner_credit_limit/i18n/es.po
+++ b/partner_credit_limit/i18n/es.po
@@ -16,12 +16,13 @@ msgstr ""
 "Plural-Forms: \n"
 
 #. module: partner_credit_limit
-#: field:res.partner,allowed_sale:0
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_partner_allowed_sale
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_users_allowed_sale
 msgid "Allowed Sales"
 msgstr "Venta a crédito permitida"
 
 #. module: partner_credit_limit
-#: code:addons/partner_credit_limit/model/sale.py:30
+#: code:addons/partner_credit_limit/model/sale.py:28
 #, python-format
 msgid "Can not confirm the Sale Order because Partner has late payments or has exceeded the credit limit.\n"
 "Please cover the late payment or check credit limit\n"
@@ -31,7 +32,7 @@ msgstr "No se puede confirmar la orden de venta por que el cliente tiene pagos a
 "Límite de Crédito : %s"
 
 #. module: partner_credit_limit
-#: code:addons/partner_credit_limit/model/accout_invoice.py:31
+#: code:addons/partner_credit_limit/model/account_invoice.py:30
 #, python-format
 msgid "Can not validate the Invoice because Partner has late payments or has exceeded the credit limit.\n"
 "Please cover the late payment or check credit limit\n"
@@ -41,28 +42,39 @@ msgstr "No se puede validar la Factura porque el cliente tiene pagos atrasados o
 "Límite de Crédito : %s"
 
 #. module: partner_credit_limit
-#: field:res.partner,credit_overloaded:0
+#: model:ir.model,name:partner_credit_limit.model_res_partner
+msgid "Contact"
+msgstr "Contacto"
+
+#. module: partner_credit_limit
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_partner_credit_overloaded
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_users_credit_overloaded
 msgid "Credit Overloaded"
 msgstr "Crédito Excedido"
 
 #. module: partner_credit_limit
-#: field:res.partner,grace_payment_days:0
-#: help:res.partner,grace_payment_days:0
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_partner_grace_payment_days
+#: model:ir.model.fields,field_description:partner_credit_limit.field_res_users_grace_payment_days
+#: model:ir.model.fields,help:partner_credit_limit.field_res_partner_grace_payment_days
+#: model:ir.model.fields,help:partner_credit_limit.field_res_users_grace_payment_days
 msgid "Days grace payment"
 msgstr "Días de gracia para pago"
 
 #. module: partner_credit_limit
-#: help:res.partner,allowed_sale:0
+#: model:ir.model.fields,help:partner_credit_limit.field_res_partner_allowed_sale
+#: model:ir.model.fields,help:partner_credit_limit.field_res_users_allowed_sale
 msgid "If the Partner has credit overloaded or late payments, he can't validate invoices and sale orders."
 msgstr "Si el Cliente tiene crédito excedido o pagos atrasados, no puede validar facturas y ordenes de venta."
 
 #. module: partner_credit_limit
-#: help:res.partner,credit_overloaded:0
+#: model:ir.model.fields,help:partner_credit_limit.field_res_partner_credit_overloaded
+#: model:ir.model.fields,help:partner_credit_limit.field_res_users_credit_overloaded
 msgid "Indicates when the customer has credit overloaded"
 msgstr "Indica cuando el cliente tiene crédito excedido"
 
 #. module: partner_credit_limit
-#: help:res.partner,overdue_credit:0
+#: model:ir.model.fields,help:partner_credit_limit.field_res_partner_overdue_credit
+#: model:ir.model.fields,help:partner_credit_limit.field_res_users_overdue_credit
 msgid "Indicates when the customer has late payments"
 msgstr "Indica cuando el cliente tiene pagos atrasados"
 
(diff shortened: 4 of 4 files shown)


edy1192:
@luistorresm Could you check this, please?

Regards.

luistorresm:
@JulioSerna @hbto 

LGTM this module improve, what do you think?

Regards

hugho-ad:
An extra LGTM :+1:

hugho-ad (approved):

luistorresm (approved):

partner_credit_limit/model/partner.py
```diff
@@ -40,7 +41,10 @@ def _get_credit_overloaded(self):
 
     @api.model
     def movelines_domain(self, partner):
+        """Return the domain for search the
+        account.move.line for the user's company."""
         domain = [('partner_id', '=', partner.id),
+                  ('company_id', '=', self.env.user.company_id.id),
```
JulioSerna, line 47: I think this change is not necessary because there is a ir.rule already that validate the company

did you test without this line?

edy1192:
@hbto Could you check this, please? 

Regards.

hbto:
:+1:

hbto (approved):
LGTM

luistorresm:
@moylop260 

Could you merge this PR, please?

Regards

partner_credit_limit/model/partner.py
edy1192: Hi @JulioSerna 

Yes, a customer and I tested it without this line. Actually, the client requested this change and @luistorresm advised me to do it in this repo and thus have it available for other clients.
Here is a small video with my test https://youtu.be/d9p8Hs4JkiQ

Thanks and Regards

cc @hbto

hugho-ad: @edy1192 

Did you test it with another customer DB?

I'm wonder if will be needed a migration script, in order to set old values on new properties

edy1192: Hi @hugho-ad I tested it here https://fertico-11-0-partner-credit-limit-edy1192-634184.dev.odoo.com/web this is a customer db.  Yesterday I sent a video https://youtu.be/7Qyi-SSNMt8 to the customer with my tests. 

With this video you can know if the script is necessary? I ask because I don't understand what you mean.

I await your comments

Thanks in advance.

hugho-ad: @edy1192 

Since fields now will become on property fields, I don't know id the value was set on it

I mean:
e.g.
Before:
partner.credit_limit  return 5 from res.partner table

After
partner.credit_limit returns ??  not sure it will return the original value 5 
or False since this is now get it from ir.properties

Regards!

luistorresm: Hello @hugho-ad 

These fields are computed with store=False, why you think that the value not will be updated?

Regards

edy1192: Hi @hugho-ad 

I tested what you told me, https://youtu.be/z6quTSznYpA 

Like like you said, the credit limit value is lost after the change.

The migration script you propose what should It do?

Thanks in advance
cc @luistorresm

hugho-ad: @luistorresm tests made by @edy1192  proved what I said

@edy1192 
the migration script should recover all that values and set them again in the ir.properties for each partner

values must exist after the upgrade, so you will create a post-migration script that will read the columns from res.partner table and then write the new value on the ir.property

something like following:

```sql 
select id, credit_limit, from res_partner where credit_limit is not null;
```

```python
# then you read the result and make id browseables records
data = res.dictfetchall()
partners = env['res.parnter'].browse(res.get('ids') or [])
for partner in partners:
    # not sure, if is needed here pass the company_id
    partner.write({'credit_limit': data[parner.id]})  # This will automatically set the ir.property
```

Note: The above is pseudo code, just to show you the main idea about the job to be done

Regards!

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+
+def migrate(cr, version):
+    set_credit_limit_values_to_partner(cr)
+
+
+def set_credit_limit_values_to_partner(cr):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env.cr.execute(
+        """
+        SELECT id, credit_limit FROM res_partner
```
moylop260: What is the reason to use a query instead of odoo api?

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+        WHERE credit_limit IS NOT NULL""")
+    data = env.cr.dictfetchall()
+    credit_limit_vals = {p['id']: p['credit_limit'] for p in data}
+    partner_ids = [p.get('id') for p in data]
+    partners = env['res.partner'].browse(partner_ids or [])
+    for partner in partners:
+        partner.write({'credit_limit': credit_limit_vals.get(partner.id, 0)})
```
moylop260: You should force the write for all companies
I mean, a
for company in companies:
  partner.with_context(force_company=company).write(...

I don't remember the real flag

Since that old version the same value applies for all companies

moylop260: You are moving to company_dependent 2 fields
But you are updating just one

What about “grace_payment_days”?

hugho-ad: grace_payments does not appear on DB yet, @edy1192 still debugging

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
hugho-ad: because this is a post-migration script, so after the field change values returned by the api like 
`partner.credit_limit` is False,
that why the migration script in order to recover that values from db

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
edy1192: Hi @moylop260 @hugho-ad 

I think what happens is that when the module is installed with the change of `company_dependent`, these fields become computed and are not found in the bd. But if the module was installed before this change, it already exists in bd. It is the theory that I have, but I have to do other tests to verify this.

But right now I am with other issues.

Thanks for your comments.

Regards.

edy1192: > You should force the write for all companies
> I mean, a
> for company in companies:
> partner.with_context(force_company=company).write(...
> 
> I don't remember the real flag
> 
> Since that old version the same value applies for all companies

Got it, I will try that. Thanks 😄

hugho-ad: @edy1192 

grace_payments only will found on instances with partner_credit_limit module previously installed only because these fields become computed ones.

So don't worry, the migration script only will be executed on those instances

try it and let us know

edy1192: Thanks @hugho-ad  you're right! 😄 

@moylop260 You were right too! Without the force company it only established the values in one company and not in all of them. Thanks for the advice 😄 

Regards.

edy1192:
@hugho-ad @moylop260 I applied the changes that you suggested! Could you check it please?

Regards.

hugho-ad:
LGTM :+1:

edy1192:
Hi @moylop260 Could you review / merge this, please?

cc @hugho-ad 

Regards

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def migrate(cr, version):
+    set_credit_limit_values_to_partner(cr)
+
+
+def set_credit_limit_values_to_partner(cr):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env.cr.execute(
+        """
```
moylop260, line 11: My last review was about using api instead of query here

edy1192: @moylop260 For that @hugho-ad replied this:

> because this is a post-migration script, so after the field change values returned by the api like
> `partner.credit_limit` is False,
> that why the migration script in order to recover that values from db

And he's right when I use api this return False for the values but the query returns the correct values.

Regards.

partner_credit_limit/model/partner.py
JulioSerna: @edy1192 

you tested with admin, admin pass the ir.rule

edy1192: Good point @JulioSerna Let me try with normal users. I'll let you know the results.

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
moylop260: You need using domain with None or empty string

partner_credit_limit/model/partner.py
edy1192: @JulioSerna @hugho-ad I tried it with demo user and here is the result https://youtu.be/LeavM1fmDJ8

Could you let me know your comments?

Regards.

JulioSerna: @edy1192 

LGTM 👍

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
edy1192: Hi @moylop260 I was trying to test your advice. 

This is my tests:
1. Install the module before the `company_dependent` change.
2. Set values to fields `credit_limit` and `grace_payment_days` for 6 partners.
3. Update the module with the `company_dependent` change, without the migration.
4. Set values to fields `credit_limit` and `grace_payment_days` for 2 more partners.
5. Update the module with the migration and this is the result:
     When I execute the sql query, the result is 6 partners.
     When I execute the search the result is 2 partners.
     When I execute the search with None or ' ' I get an error.

I left this image with a pdb in the migration script.

Screen Shot 2019-10-25 at 12 10 42 PM


Please let me know your comments. 

Thanks  in advance.

cc @hugho-ad

partner_credit_limit/model/partner.py
hugho-ad: @moylop260 here is the approval

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+        SELECT id, credit_limit, grace_payment_days FROM res_partner
+        WHERE credit_limit != 0 OR grace_payment_days != 0""")
+    data = env.cr.dictfetchall()
+    credit_limit_vals = {
+        p['id']: {
+            'credit_limit': p['credit_limit'],
+            'grace_payment_days': p['grace_payment_days']} for p in data}
+    partner_ids = [p.get('id') for p in data]
```
moylop260: You don't need transform too much times the same records from "list-of-tuples" to "dict-of-dict" and get a list of ids to  get a browse list.

You can using directly the fetched result with `fetchall` and a `for` over the result
```python
    env = api.Environment(cr, SUPERUSER_ID, {})
    env.cr.execute("""
        SELECT id, credit_limit, grace_payment_days FROM res_partner
        WHERE credit_limit != 0 OR grace_payment_days != 0""")
    data = env.cr.fetchall()
    for company in env['res.company'].search([]):
        for partner_id, credit_limit, grace_payment_days in data:
            env['res.partner'].browse(partner_id).with_context(force_company=company.id).write({
                'credit_limit': credit_limit
                'grace_payment_days': grace_payment_days})
```

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+    partners = env['res.partner'].browse(partner_ids or [])
+    companies = env['res.company'].search([])
+    for company in companies:
+        for partner in partners:
+            partner.with_context(force_company=company.id).write({
+                'credit_limit': credit_limit_vals.get(partner.id, {}).get(
+                    'credit_limit', 0),
+                'grace_payment_days': credit_limit_vals.get(
+                    partner.id, {}).get('grace_payment_days', 0)})
```
moylop260: `ALTER TABLE res_partner DROP COLUMN credit_limit;`
`ALTER TABLE res_partner DROP COLUMN grace_payment_days;`

hugho-ad: Why remove it?
I think we should keep it in order to have that values for the record.

Regards!

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
moylop260: Cool!
I understand that after our meeting Edi.

The matter here is that the api is adapted for new definition of the field using ir.property (multi-company field).
So, it is not possible get the old values since that the search will be for ir.property table instead of res.partner like old way.

So, let's using query directly.

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
hugho-ad: I like the browse out the for loop, should be faster ?

this is just a migration script, that will be executed only one time

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
edy1192: Yes! Maybe my explanations was not totally clear! Thanks for your time! @moylop260  😄

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
moylop260: The browse-list is faster to fetch data, since that read all data at once time.
But to write data it is the same, since that write record-by-record
I know this will be executed just one time
But all PM must understand the full change
And the current code is so heavy to understand using too much transformations

partner_credit_limit/migrations/11.0.0.0.2/post-migration.py
moylop260: Better using [ir_model._drop_column method](https://github.com/odoo/odoo/blob/9d7147a3c189d9e2edcdeb0e5cf28b75cf7f366b/odoo/addons/base/ir/ir_model.py#L524)

Because it shouldn't used anymore.
If you have a script using this value directly from database so it will use a deprecated field and outdated value silently.

We need a raise or real fail in other to fix it.

But if you need storing for the record (good point here), you can use another way (Storing a `ir_log` with `data` variable result)

hugho-ad: Got it

edy1192:
@moylop260 I applied the comments that you gave me in the meeting. 

I will ask the team, customers with this module to make dummy MRs and test these changes.

Thanks for your advices. 

@hugho-ad @luistorresm Thanks you too 

😄 

Regards.

moylop260 (approved):
👍

PR #1372, opened 2019-10-18
[IMP] addons_vauxoo: New `variables.sh` file configuration

edy1192:
- According to the new deployV configuration, it is necessary to add the
variables.sh file and its additional configurations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ virtualenv:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


edy1192:
@luistorresm Could you check this, please?

Regards.

luisg123v:
@edy1192 
Why are you performing such a change here? I have to be missing the point :confused:

PR #1377, opened 2019-11-06
[FIX] mrp_workcenter_account_move: Avoid traceback when creating a valuation adjustment line in account.move

edy1192:
If, when closing a manufacturing order, a valuation adjustment line is created, taking the `stock_valuation` account, the following traceback is obtained

image

image

This happens because it try to create the line with the object `account.account` and not with the account id.

image

This PR fix that case. 

image

image


edy1192:
@hbto @moylop260 @keylor2906  Could you check this, please??

This traceback is in lodi. 

Regards. 

cc @dsabrinarg

PR #1381, opened 2019-11-29
[FIX] account_anglo_saxon_stock_move: Use correct field to search for… issue#7855

keylor2906:
… the amls that should be reconciled.


keylor2906:
@hbto This is the change of using the correct field. 

cc @dsabrinarg

hbto:
:+1:

I agree on this change 

Go ahead and merge

PR #1385, opened 2019-12-06
[FIX] account_refund_early_payment: Fix account dependency

deivislaya:
Travis is failing because the `account_refund_early_payment` module
    depends on `account_accountant` instead of `account`. The module
    `account_accounting`, from the v12 is available on enterprise.

### DIFF
--- a/account_refund_early_payment/__manifest__.py
+++ b/account_refund_early_payment/__manifest__.py
@@ -9,7 +9,7 @@
     'license': 'LGPL-3',
     'category': '',
     'depends': [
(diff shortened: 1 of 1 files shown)


luisg123v (changes requested):
`account_accountant` does exist in v12, but it's on enterprise.

Could you reword the commit message?

luisg123v (approved):

luisg123v:
@moylop260,

This is the first step to make travis green again (use Trump's voice to read)

PR #1387, opened 2019-12-13
[FIX] account_analytic_policy: When create an account move line, the validation of the account and analytic account fails because self does not exist

ysantiago:
When a stock_move_line is created, validation does not work because self is empty, therefore, the creation (called super) must be first, and then, validation works correctly.


ysantiago:
Hello @keylor2906 can you please review this change please. Is related with task https://www.vauxoo.com/web#id=8255&model=helpdesk.ticket&view_type=form&menu_id=

PR #1388, opened 2019-12-28
[MIG] stock_by_warehouse: Migrate to v12.0

luisg123v:
The following is performed:
- Remove useless test that actually did nothing
- Remove data tags from views
- Move assets to a separated file
- Fix JS
- Remove encoding headers from Python files
- Split product template and product variant code into two files
- Fix typos on the README


luisg123v:
@moylop260  

CC @yaniaular

stock_by_warehouse/tests/test_stock_available_unreserved.py
```diff
@@ -0,0 +1,124 @@
...
+        picking_in_b = self.create_picking(self.ref('stock.picking_type_in'),
+                                           self.supplier_location,
+                                           self.stock_location,
+                                           product_b, 3)
+
+        def compare_qty_available_not_res(product, value):
+            # Refresh, because the function field is not recalculated between
+            # transactions
+            product.refresh()
+            self.assertEqual(product.qty_available_not_res, value)
```
nhomar, line 78: why not simply put this linear in the method? this complicates the test readability!.

In other hand the test should have a clear returned message if it fails, now it s returning nothing.

luisg123v: I kind of agree. I didn't pay too much attention to the tests, because I didn't understand them very well and they passed, so I left them alone.

PR #1389, opened 2019-12-30
[FIX] stock_by_warehouse: File bad named.

JVegaB:
## Problem

A file in /stock_by_warehouse/models was named without the extension, so
cannot be used/imported.

## Fix

The extension was just added.

## Pics

Screenshot from 2019-12-30 10-44-44
Screenshot from 2019-12-30 10-44-35


luisg123v (approved):
@moylop260  
Please merge, my bad.

CC @yaniaular

PR #1390, opened 2019-12-30
[MIG] stock_by_warehouse_(sale|purchase): Migrate to v12.0

luisg123v:
The following is performed:
- Remove useless test that actually did nothing
- Remove data tags from views
- Remove encoding headers from Python files
- Remove superfluous readonly attribute from related fields
- Remove useless keys from the manifest
- Fix typos on the README


luisg123v:
@moylop260  

CC @yaniaular

PR #1392, opened 2020-01-22
[I18N] stock_by_warehouse_sale: Fix term names

luisg123v:
Some term names are handled differently on v12.0, but those changes were
not taken into account when migrating this module from v11.0.


luisg123v:
@moylop260   
Could you merge, please?

CC @yaniaular

PR #1394, opened 2020-02-06
[FIX] account_anglo_saxon_stock_move: Handle exception properly

luisg123v:
When the accrual stock is reconciled through the cron and an exception
occurs, the error is logged. however, it's not being handled properly,
which causes another exception.

The exception is checked for its name, value and message. However, Odoo
exceptions don't normally have a message attribute. A safer way is to
check its arguments, which is what this commit does.


luisg123v:
@keylor2906 could you check this, please?

I'm getting this error on the databases I'm migrating.

CC @hbto

hbto:
👍

PR #1395, opened 2020-02-11
[FIX] partner_credit_limit: Add hook to initialize credit limit

luisg123v:
Starting from 1e18a7457, the partner's credit limit field is
company-dependent, so their values need to be set manually, because they
are properties instead of a standard field.

A migration script was provided on the aforementioned commit to set that
value when the module is already installed. However, since that field is
a native one, values need to be set also when the module is installed.

This commit fixes the above by implementing a post-init hook to set
required values.

### DIFF
--- a/partner_credit_limit/hooks.py
+++ b/partner_credit_limit/hooks.py
@@ -1,0 +1,42 @@
+# Copyright 2020 Vauxoo (https://www.vauxoo.com) <info@vauxoo.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, SUPERUSER_ID
+from odoo.tools import column_exists
+
+
+def post_init_hook(cr, registry):
+    _set_credit_limit(cr, registry)
+
+
+def _set_credit_limit(cr, registry):
+    """Set the credit limit field to partners
+
+    Since this module enables company-specific credit limits, their values need
+    to be set manually for each company.
+    """
+    if not column_exists(cr, 'res_partner', 'credit_limit'):
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    cr.execute("""
+        SELECT id, credit_limit
+        FROM res_partner
+        WHERE credit_limit != 0;
+    """)
+    # Create a dict {partner_id: credit limit, ...}
+    credit_limit_values = dict(env.cr.fetchall())
+
+    # Set values for each company, creating corresponding properties
+    companies = env['res.company'].search([])
+    property_model = env['ir.property']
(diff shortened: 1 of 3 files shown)


luisg123v:
@keylor2906 Could you review, please?

This fix the issue regarding credit limit values not being migrated.

CC @edy1192

partner_credit_limit/hooks.py
```diff
@@ -0,0 +1,39 @@
...
+        WHERE credit_limit != 0;
+    """)
+    res = env.cr.fetchall()
+    companies = env['res.company'].search([])
+    partner_model = env['res.partner']
+    for company in companies:
+        for partner_id, credit_limit in res:
+            partner = partner_model.browse(partner_id)
+            partner.with_context(force_company=company.id).write({
+                'credit_limit': credit_limit,
```
nhomar: I do not get it, are you wiring the same credit limit in all partners?

Who did ask this for?

nhomar: Why aren't the PM simply exporting and importing after install? which problem is being solving this?

luisg123v: Hi @nhomar,

> Why aren't the PM simply exporting and importing after install? which problem is being solving this?

Short answer:
if those values are exported, all will be zero, so that won't work.

Long answer:
This module makes the credit limit field to be company-dependent. That causes current values to be lost (rater ignored, as they're actually on DB).

Because of the above, this feature requires a manual procedure to read the values from DB and re-write them for all companies, which will create the correct properties.

This is already performed on a migration script (see https://github.com/Vauxoo/addons-vauxoo/pull/1371), but that didnt consider when this module is being installed.

luisg123v:
@moylop260 or @keylor2906 

partner_credit_limit/hooks.py
moylop260: @luisg123v 
It was considered from https://github.com/Vauxoo/addons-vauxoo/pull/1371
In fact, I requested the migration script:


https://github.com/vauxoo/addons-vauxoo/blob/2e5633e46780c4c2d3b0e531753a28f40f99ff7f/partner_credit_limit/migrations/11.0.0.0.2/post-migration.py#L10-L20


So, this post_init_hook is not required since that if you are installing the module from scratch (not version previous) so migration is not required.

moylop260: in fact, we are considering both fields `credit_limit` and `grace_payment_days`

moylop260:
The main issue here is because of the field `partner_credit_limit` is defined by odoo itself.
So, it is required for installation from scratch.

PR #1396, opened 2020-02-13
[FIX] stock_cost_segmentation: Access error when validating an invoice

luisg123v:
If an user that is not a warehouse manager tries to validate an invoice,
an access error is triggered.

This is a port of odoo/enterprise#3649, which applies to the module
`l10n_mx_edi_landing`, which this code was based on.

Dummy MR: https://git.vauxoo.com/vauxoo/quebec/merge_requests/64


luisg123v:
@keylor2906 or @luistorresm 
Could you review, please?

CC @edy1192

luistorresm (approved):
LGTM

luisg123v:
@moylop260 could you merge, please?

PR #1400, opened 2020-03-04
[FIX] stock_cost_segmentation: Decrease logistic qty on multi-step outputs

luisg123v:
Currently, when a product delivery is processed  and that product has no
associated lot/serial, the logistic remaining quantity is decreased from
the source location of the stock move that is processing the output.
However, that beavior is not correct on multi-step outputs, because the
source location will be the previous step, instead of the actual stock
location.

To solve the above, when the product is not found on the stock move's
source location, it is looked for on the warehouse's stock location,
along with its childs locations.


luisg123v:
@josemoralesp @keylor2906   

CC @belisac

PR #1401, opened 2020-03-11
[FIX] partner_credit_limit: Add hook to initialize credit limit

luisg123v:
Starting from 1e18a7457, the partner's credit limit field is
company-dependent, so their values need to be set manually, because they
are properties instead of a standard field.

A migration script was provided on the aforementioned commit to set that
value when the module is already installed. However, since that field is
a native one, values need to be set also when the module is installed.

This commit fixes the above by implementing a post-init hook to set
required values.

### DIFF
--- a/partner_credit_limit/__init__.py
+++ b/partner_credit_limit/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from .hooks import post_init_hook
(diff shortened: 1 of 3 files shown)


luisg123v:
@moylop260 could you merge, please?

This is a cherry pick of 9f3863c9ec5ed834c1172f81fa61e86b2a4cc874

moylop260:
Foward-port from https://github.com/Vauxoo/addons-vauxoo/pull/1395

PR #1407, opened 2020-04-06
[FIX] partner_credit_limit: fix error `OverflowError: date value out of  range`

edy1192:
1. [FIX] partner_credit_limit: fix error `OverflowError: date value out of  range`
- Since the grace_payment_days field is a float, in some cases the user can set really high values (> 999999) which causes the traceback `OverflowError: date value out of range`.
- Since that field is used for date operations in the `debit_credit_maturity` method of the` res.partner` model.

2. [FIX] partner_credit_limit: Add = to the validation condition

- There is data demo with the value of the grace_payment_days field set to 999999, that value must be included.

- Improve message that is displayed to the user when the constraint is executed, to show the value that is trying to save.

### DIFF
--- a/partner_credit_limit/__manifest__.py
+++ b/partner_credit_limit/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Partner Credit Limit",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
     "author": "Vauxoo",
     "category": "",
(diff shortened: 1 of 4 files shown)


edy1192:
Hi @luisg123v @moylop260 

Cherry pick is done.

Regards.

luisg123v (approved):

moylop260:
Thank you!

PR #1411, opened 2020-05-13
[FIX] default_warehouse_from_sale_team: decorator on write method

frahikLV:

### DIFF
--- a/default_warehouse_from_sale_team/models/account_journal.py
+++ b/default_warehouse_from_sale_team/models/account_journal.py
@@ -18,7 +18,7 @@ class AccountJournal(models.Model):
             self.clear_caches()
         return super(AccountJournal, self).write(vals)
 
-    @api.multi
+    @api.model
     def create(self, values):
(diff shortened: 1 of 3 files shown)


hugho-ad:
LGTM :+1:

JulioSerna:
👍

hugho-ad:
@frahikLV 
please check if v12.0 does not need this, in order to forward port it

luisg123v:
@frahikLV,

Did you checked if this needs forward-port?

CC @CarmenMiranda

frahikLV:
@luisg123v  No longer necessary in V12 as the create method is not inherit/called in the "new" version

luisg123v:
@frahikLV,

You're right, in v12 there's only one create method and it already has the appropriate decorator.

Thanks for checking.

hugho-ad:
AFAIK there are not any new version, the create methods was inherited because of the clear caches, a problem that should be persisting on v12.0

PR #1416, opened 2020-05-30
[REF] default_warehouse_from_sale_team: Backport fixes from 12.0

luisg123v:
The following fixes are back-ported from 12.0:
- Default journal at invoice is get from the sale team (c6866c8f6121)
- [FIX] Lints (617a8c4f7048)
- [REF] i18n: Update translations (04ab724b4ca0)


luisg123v:
@hugho-ad @moylop260,

This are clean cherry-picks, I only modified a little the translations-related commit so it doesn't introduce new-v12 syntax.

CC @belisac

hugho-ad:
@luisg123v there are dummy mr?

could you get the functional approval  please?

luisg123v:
@hugho-ad,

[Here](https://git.vauxoo.com/vauxoo/lodi/-/merge_requests/185) is a dummy MR. It has no deployv instance created, but I think its Runbot will do. the trick.

belisac:
Hi @luisg123v 

I did the functional test using runbot as you suggest and everything was ok (Quote and SO have the journal specify in Sales Channel) Please continue in order to generate the dev instance so the user could check also. Thanks in advance.

cc: @hugho-ad

PR #1418, opened 2020-06-07
[ADD] stock_custom_exchange_rate_date: Module to set custom rate date

luisg123v:
This module allows to set a customized date on transfers to be used when
computing product unit prices, instead of the date when the transfer  is
validated, in case exchange rates are involved.

The use for the above is that created journal entries take the newly
computed unit price, which makes the inventory to be valuated using the
specified rate.

### DIFF
--- a/stock_custom_exchange_rate_date/__manifest__.py
+++ b/stock_custom_exchange_rate_date/__manifest__.py
@@ -1,0 +1,18 @@
+{
+    "name": "Customized Rate Date in Stock",
+    "version": "12.0.1.0.0",
+    "author": "Vauxoo",
+    "category": "stock",
+    "website": "http://www.vauxoo.com/",
+    "license": "LGPL-3",
+    "depends": [
+        "purchase_stock",
+    ],
+    "data": [
+        "views/stock_picking_views.xml",
+    ],
+    "demo": [
+    ],
+    "installable": True,
+    "auto_install": False,
+}
--- a/stock_custom_exchange_rate_date/models/stock_move.py
+++ b/stock_custom_exchange_rate_date/models/stock_move.py
@@ -1,0 +1,26 @@
+from odoo import api, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    @api.multi
+    def _get_price_unit(self):
+        """Take into account custom rate date, if provided"""
+        line = self.purchase_line_id
+        if (not self.picking_id.exchange_rate_date or not line or self.product_id != line.product_id
+                or line.currency_id == line.company_id.currency_id):
+            return super()._get_price_unit()
+        price_unit = line.price_unit
+        if line.taxes_id:
+            price_unit = line.taxes_id.with_context(round=False).compute_all(
+                price_unit, currency=line.currency_id, quantity=1.0)['total_excluded']
+        if line.product_uom.id != line.product_id.uom_id.id:
+            price_unit *= line.product_uom.factor / line.product_id.uom_id.factor
+        price_unit = line.currency_id._convert(
+            price_unit,
+            line.company_id.currency_id,
+            line.company_id,
+            self.picking_id.exchange_rate_date,
+            round=False)
+        return price_unit
--- a/stock_custom_exchange_rate_date/models/stock_picking.py
+++ b/stock_custom_exchange_rate_date/models/stock_picking.py
@@ -1,0 +1,12 @@
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    exchange_rate_date = fields.Date(
+        states={'done': [('readonly', True)], 'cancel': [('readonly', True)]},
+        copy=False,
+        help="If set, specifies a customized date that will be used when "
+        "computing product unit prices, instead of the date when the transfer "
+        "is validated, in case exchange rates are involved.")
--- a/stock_custom_exchange_rate_date/views/stock_picking_views.xml
+++ b/stock_custom_exchange_rate_date/views/stock_picking_views.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form.inherit.custom.rate.date</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='scheduled_date']" position="before">
+                <field name="purchase_id" invisible="1"/>
+                <field name="exchange_rate_date" attrs="{'invisible': ['|', ('purchase_id', '=', False), ('picking_type_code', '!=', 'incoming')]}"/>
+            </xpath>
+        </field>
+    </record>
+
+</odoo>
--- a/stock_custom_exchange_rate_date/README.rst
+++ b/stock_custom_exchange_rate_date/README.rst
@@ -1,0 +1,32 @@
+Customized Rate Date in Stock
+=============================
+
+This module allows to set a customized date on transfers to be used when
+computing product unit prices, instead of the date when the transfer  is
+validated, in case exchange rates are involved.
+
+The use for the above is that created journal entries take the newly
+computed unit price, which makes the inventory to be valuated using the
+specified rate.
+
+Contributors
+------------
+
+* Luis González <lgonzalez@vauxoo.com>
+
+
+Maintainer
+==========
+
+.. image:: https://www.vauxoo.com/logo.png
+   :alt: Vauxoo
+   :target: https://vauxoo.com
+
+This module is maintained by Vauxoo.
+
+a latinamerican company that provides training, coaching,
+development and implementation of enterprise management
+systems and bases its entire operation strategy in the use
+of Open Source Software and its main product is Odoo.
+
+To contribute to this module, please visit http://www.vauxoo.com.
--- a/stock_custom_exchange_rate_date/__init__.py
+++ b/stock_custom_exchange_rate_date/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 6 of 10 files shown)


luisg123v:
@nhomar,

Could you review this, please?

CC @rsilvam21 @hugho-ad

luisg123v:
FYI @dsabrinarg,

This is the module for the DTT's requirement regarding date rate on transfers.

stock_custom_exchange_rate_date/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Customized Rate Date in Stock",
+    "version": "12.0.1.0.0",
+    "author": "Vauxoo",
+    "category": "stock",
+    "website": "http://www.vauxoo.com/",
+    "license": "LGPL-3",
+    "depends": [
+        "stock",
```
JulioSerna: @luisg123v 

what about only purchase_stock module?

stock_custom_exchange_rate_date/views/stock_picking_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form.inherit.custom.rate.date</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='scheduled_date']" position="before">
+                <field name="exchange_rate_date"/>
```
JulioSerna: this field only make sense when it is a reception only should be in picking_type = supplier

JulioSerna:
@luisg123v 

I think this field could be omitted by the user that make the reception and I think this field should be required when the partner o purchase related it is in secondary currency

what do you think @nhomar ?

stock_custom_exchange_rate_date/__manifest__.py
luisg123v: I was trying to avoid depending on that module because it's an autoinstall one, but I can change it if you want.

stock_custom_exchange_rate_date/views/stock_picking_views.xml
luisg123v: Oh really? I wasn't aware of that, I though this also made sense on outgoing transfers.

nhomar:
>> I think this field could be omitted by the user that make the reception and I think this field should be required when the partner o purchase related it is in secondary currency

@JulioSerna I think it is ok like this with good dcumentation, Maybe we must to look for an automatic filling following some criteria and more smarter wizard to ask explicitaly for this.

For now it is ok the required but I think it will bring functional problems (@luisg123v please execute a required with this criteria if possible).

stock_custom_exchange_rate_date/__manifest__.py
nhomar: the correct dependency is purchase_stock (it is the one with the behavior we are overwritting here)

stock_custom_exchange_rate_date/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+        "purchase",
```
nhomar: purchase_stock only is the one needed to depend from.

stock_custom_exchange_rate_date/models/stock_move.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @api.multi
+    def _get_price_unit(self):
+        """Take into account custom rate date, if provided"""
+        line = self.purchase_line_id
+        if (not self.picking_id.exchange_rate_date
+                or not line
+                or self.product_id != line.product_id
+                or line.currency_id == line.company_id.currency_id):
+            return super()._get_price_unit()
```
nhomar, line 13: try 119 length of line please :-(

stock_custom_exchange_rate_date/models/stock_move.py
```diff
@@ -0,0 +1,28 @@
...
+        if line.taxes_id:
+            price_unit = line.taxes_id.with_context(round=False).compute_all(
+                price_unit, currency=line.currency_id, quantity=1.0)['total_excluded']
+        if line.product_uom.id != line.product_id.uom_id.id:
+            price_unit *= line.product_uom.factor / line.product_id.uom_id.factor
+        price_unit = line.currency_id._convert(
+            price_unit,
+            line.company_id.currency_id,
+            line.company_id,
+            self.picking_id.exchange_rate_date,
```
nhomar, line 24: `self.picking_id.exchange_rate_date and self.picking_id.exchange_rate_date or fields.Date.context_today(self)`

stock_custom_exchange_rate_date/models/stock_move.py
luisg123v: You mean the if or the super?

stock_custom_exchange_rate_date/models/stock_move.py
luisg123v: `exchange_rate_date` will always be set.. If not, I'm falling back to default behavior.

stock_custom_exchange_rate_date/__manifest__.py
luisg123v: Done

stock_custom_exchange_rate_date/views/stock_picking_views.xml
luisg123v: Done, readonly domain added

stock_custom_exchange_rate_date/__manifest__.py
luisg123v: Done

stock_custom_exchange_rate_date/models/stock_move.py
luisg123v: Done. I personally liked a little bit more the previous form, though ;).

stock_custom_exchange_rate_date/models/stock_move.py
luisg123v: It's the first condition evaluated to run the super

luisg123v:
> For now it is ok the required but I think it will bring functional problems (@luisg123v please execute a required with this criteria if possible).

AFAIK, this is not possible without introducing several new fields, because we would need to compare the company's currency against the linked purchase order's currency, neither of which are available on the picking natively.

nhomar (approved):
i APPROVE AS VERSION 1.0 @JulioSerna THE FUNCTIONAL validations can be done in a separated commit (we need to pass the tests in different instances bringing new list of topics to be done here as well but This MVP should not be stop for that).

nhomar:
@hbto Your accounting eagle eye is required here please before merge.

hbto:
As a MVP is ok for me,

But we have to take into account the suggestions made by @JulioSerna 
- this only to be applied on purchase-related pickings.

The if statemente at stock.move says: `I will skip the new computation unless a exchange_rate_date` is not set.

Next version should care about that.

Regards.

hbto (approved):
As an MVP is ok for me,

But we have to take into account the suggestions made by @JulioSerna

this only to be applied on purchase-related pickings.
The if statemente at stock.move says: I will skip the new computation unless a exchange_rate_date is not set.

The next version should care about that.

Regards.

luisg123v:
@hbto,

The if statement also skips the new behavior if the picking doesn't come from a purchase order.

luisg123v:
@hbto,

I managed to hide the new field when the picking doesn't come from a purchase order, thanks to the field `purchase_id` available on the picking.

FYI I also included tests

sabrinagnu:
Successful functional tests 👍

stock_custom_exchange_rate_date/models/stock_picking.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    exchange_rate_date = fields.Date(
+        states={'done': [('readonly', True)], 'cancel': [('readonly', True)]},
```
JulioSerna, line 8: @luisg123v 

The fields should not be copy when there is a backorder, maybe only adding copy=False it is enough to backorder

luisg123v: Done, thanks! Yes, that should be enough, as backorder is generated copying the original picking. That field should be copy=False anyway.

rsilvam21:
+1

I tested in this dummy:

Dummy B&F:
https://git.vauxoo.com/benandfrank/benandfrank/-/merge_requests/990

@nhomar @moylop260 

Could you merge this, please

JulioSerna (approved):

luisg123v:
This has been approved by @nhomar, @hbto, @JulioSerna  and @dsabrinarg, so I'll merge.

stock_custom_exchange_rate_date/models/stock_move.py
```diff
@@ -0,0 +1,26 @@
+from odoo import api, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    @api.multi
+    def _get_price_unit(self):
```
hbto, line 8: BTW

@luisg123v @JulioSerna This method will only run when the Picking is validated? Won't it?

In case that the Pedimento is set as a date after the Picking is validated, what will be the approch?

Will the Journal Be modified, Is there a provision for it, or the picking will remain the same?


Regards

luisg123v: Once the picking is validated, this field is readonly. Not sure if that answer your questiong, though, or you're asking for a new feature.

hbto: @luisg123v You have nailed it.

So if the picking is validated prior to receiving the pedimento then odds are that the padimento date is different that the one in the picking.

How will we deal with this? @JulioSerna @rsilvam21 @nhomar 

Thanks

PR #1420, opened 2020-06-15
[FIX] stock_allow_past_date: Add stock move valuation logic that is o… task#39361

keylor2906:
…riginally added on _action_done() method for model stock_move inherited on module stock_account.


keylor2906:
@moylop260 Could you review this, please?

moylop260:
@luisg123v 
Could you check it, please?

luisg123v (approved):
Even though it's duplicated code, it seems unavoidable, and the whole method is already duplicated anyway.

PR #1421, opened 2020-06-15
[FIX] partner_credit_limit: New Raise Limit Credit Message

CarmenMiranda:
### [T#39965](https://www.vauxoo.com/web#id=39965&view_type=form&model=project.task)

Adding new messages to allow the user know the real reason why its invoice or sale order can't be validated/confirmed. The reasons can be if the partner has late payments, has exceeded the credit limit or both.

Task 39965 - Raise New Message

### DIFF
--- a/partner_credit_limit/i18n/es.po
+++ b/partner_credit_limit/i18n/es.po
@@ -22,26 +22,6 @@ msgid "Allowed Sales"
 msgstr "Venta a crédito permitida"
 
 #. module: partner_credit_limit
-#: code:addons/partner_credit_limit/model/sale.py:28
-#, python-format
-msgid "Can not confirm the Sale Order because Partner has late payments or has exceeded the credit limit.\n"
-"Please cover the late payment or check credit limit\n"
-"Credit Limit : %s"
-msgstr "No se puede confirmar la orden de venta por que el cliente tiene pagos atrasados o tiene un límite de crédito excedido.\n"
-"Por favor cubrir los pagos atrasados o revisar el límite de crédito.\n"
-"Límite de Crédito : %s"
-
-#. module: partner_credit_limit
-#: code:addons/partner_credit_limit/model/account_invoice.py:30
-#, python-format
-msgid "Can not validate the Invoice because Partner has late payments or has exceeded the credit limit.\n"
-"Please cover the late payment or check credit limit\n"
-"Credit Limit : %s"
-msgstr "No se puede validar la Factura porque el cliente tiene pagos atrasados o tiene un límite de crédito excedido.\n"
-"Por favor cubrir los pagos atrasados o revisar el límite de crédito.\n"
-"Límite de Crédito : %s"
-
-#. module: partner_credit_limit
 #: model:ir.model,name:partner_credit_limit.model_res_partner
 msgid "Contact"
 msgstr "Contacto"
@@ -79,6 +59,7 @@ msgid "Indicates when the customer has late payments"
 msgstr "Indica cuando el cliente tiene pagos atrasados"
 
 #. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/account_invoice.py:23
 #: model:ir.model,name:partner_credit_limit.model_account_invoice
 msgid "Invoice"
 msgstr "Factura"
@@ -100,3 +81,55 @@ msgstr "Valor incorrecto %s para los días de gracia de pago: el valor debe esta
 #: model:ir.model,name:partner_credit_limit.model_sale_order
 msgid "Quotation"
 msgstr "Presupuesto"
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:107
+#, python-format
+msgid "Please: \n"
+""
+msgstr "Por favor: \n"
+""
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/sale.py:21
+#, python-format
+msgid "Sale Order"
+msgstr "Pedido de venta"
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:107
+#, python-format
+msgid "record"
+msgstr "registro"
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:114
+#, python-format
+msgid "• Check credit limit.\n"
+""
+msgstr "• Revisar el límite de crédito.\n"
+""
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:111
+#, python-format
+msgid "• Cover the late payments.\n"
+""
+msgstr "• Cubrir los pagos atrasados.\n"
+""
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:113
+#, python-format
+msgid "• Has exceeded the credit limit.\n"
+""
+msgstr "• Tiene un límite de crédito excedido.\n"
+""
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:110
+#, python-format
+msgid "• Has late payments.\n"
+""
+msgstr "• Tiene pagos atrasados.\n"
+""
--- a/partner_credit_limit/models/account_invoice.py
+++ b/partner_credit_limit/models/account_invoice.py
@@ -1,5 +1,5 @@
 from odoo import models, api, _
-from odoo import exceptions
+from odoo.exceptions import ValidationError
 
 
 class AccontInvoice(models.Model):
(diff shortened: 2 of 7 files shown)


partner_credit_limit/models/account_invoice.py
```diff
@@ -23,13 +23,11 @@ def check_limit_credit(self):
...
-                        'has late payments or has exceeded the credit limit.'
-                        '\nPlease cover the late payment or check credit limit'
-                        '\nCredit'
-                        ' Limit : %s') % (invoice.partner_id.credit_limit)
-                raise exceptions.Warning(msg)
+            msg = invoice.partner_id.with_context(
+                {'new_amount': invoice.amount_total,
+                    'new_currency': invoice.currency_id,
+                    'model_name': _('Invoice')})._check_limit_credit()
+            raise exceptions.Warning(msg)
```
luisg123v: `exceptions.Warning` is deprecated. Could you use `UserError` instead?

partner_credit_limit/models/account_invoice.py
```diff
@@ -23,13 +23,11 @@ def check_limit_credit(self):
...
                 return True
-            else:
-                msg = _('Can not validate the Invoice because Partner '
-                        'has late payments or has exceeded the credit limit.'
-                        '\nPlease cover the late payment or check credit limit'
-                        '\nCredit'
-                        ' Limit : %s') % (invoice.partner_id.credit_limit)
-                raise exceptions.Warning(msg)
+            msg = invoice.partner_id.with_context(
+                {'new_amount': invoice.amount_total,
```
luisg123v: Please avoid passing context as a dictionary, this causes current context to be lost. A lint is coming for this on https://github.com/OCA/pylint-odoo/pull/256

partner_credit_limit/models/partner.py
```diff
@@ -92,3 +95,23 @@ def get_allowed_sale(self):
...
+
+    def _check_limit_credit(self):
+        """Method that will return a message when the partner has late payments and/or has exceeded the credit limit.
+        """
+        self.ensure_one()
+        if self.allowed_sale:
+            return False
+        credit_overloaded = self.credit_overloaded
+        overdue_credit = self.overdue_credit
+        msg = _('Can not confirm the %s because the Partner:\n') % (self.env.context.get('model_name', _('record')))
```
luisg123v: instead of passing these parameters by context, why not simply passing them as standard method parameters?

partner_credit_limit/models/partner.py
```diff
@@ -92,3 +95,23 @@ def get_allowed_sale(self):
         for partner in self:
             partner.allowed_sale = not partner.credit_overloaded and \
                 not partner.overdue_credit
+
+    def _check_limit_credit(self):
```
luisg123v: `_check_credit_limit`

Limit credit is semantically incorrect. Yes, I know, it was that way.

luisg123v (changes requested):
Could you increase version, please? So modules using this in Odoo.sh update translations

partner_credit_limit/models/partner.py
```diff
@@ -92,3 +95,23 @@ def get_allowed_sale(self):
...
+        msg = _('Can not confirm the %s because the Partner:\n') % (self.env.context.get('model_name', _('record')))
+        do_msg = _('Please: \n')
+
+        if overdue_credit:
+            msg += TAB_SPACE + _('• Has late payments.\n')
+            do_msg += TAB_SPACE + _('• Cover the late payments.\n')
+        if credit_overloaded:
+            msg += TAB_SPACE + _('• Has exceeded the credit limit.\n')
+            do_msg += TAB_SPACE + _('• Check credit limit.\n')
+            do_msg += TAB_SPACE*2 + _('Credit limit: %s') % self.credit_limit
```
luisg123v, line 116: This is not so clear IMHO.

What do you think about using a method like the following?

```
def indent_meesage(msg, indent_level=1):
```

partner_credit_limit/models/account_invoice.py
CarmenMiranda: Done, I used ValidationError

partner_credit_limit/models/account_invoice.py
CarmenMiranda: Done

partner_credit_limit/models/partner.py
CarmenMiranda: Done

hugho-ad:
LGTM :+1:

partner_credit_limit/models/partner.py
CarmenMiranda: Done

partner_credit_limit/i18n/es.po
```diff
@@ -100,3 +81,56 @@ msgstr "Valor incorrecto %s para los días de gracia de pago: el valor debe esta
 #: model:ir.model,name:partner_credit_limit.model_sale_order
 msgid "Quotation"
 msgstr "Presupuesto"
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:107
+#, python-format
+msgid "Please: \n"
+""
```
JulioSerna, line 89: I think this is not correct

partner_credit_limit/i18n/es.po
```diff
@@ -100,3 +81,56 @@ msgstr "Valor incorrecto %s para los días de gracia de pago: el valor debe esta
...
+#: code:addons/partner_credit_limit/models/partner.py:107
+#, python-format
+msgid "record"
+msgstr "registro"
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/partner.py:114
+#, python-format
+msgid "• Check credit limit.\n"
+""
```
JulioSerna, line 109: same here

partner_credit_limit/i18n/es.po
```diff
@@ -100,3 +81,56 @@ msgstr "Valor incorrecto %s para los días de gracia de pago: el valor debe esta
...
+msgstr "• Revisar el límite de crédito.\n"
+""
```
JulioSerna, line 111: same here

JulioSerna:
@hugho-ad @CarmenMiranda 

Please could you check the translation file is working as expected?

partner_credit_limit/i18n/es.po
hugho-ad: Why?

partner_credit_limit/i18n/es.po
```diff
@@ -100,3 +81,56 @@ msgstr "Valor incorrecto %s para los días de gracia de pago: el valor debe esta
...
+#: code:addons/partner_credit_limit/models/partner.py:107
+#, python-format
+msgid "Please: \n"
+""
+msgstr "Por favor: \n"
+""
+
+#. module: partner_credit_limit
+#: code:addons/partner_credit_limit/models/sale.py:26
+#: model:ir.model,name:partner_credit_limit.model_sale_order
```
JulioSerna: @hugho-ad @CarmenMiranda @luisg123v 

I think this is the translate in the issue, this line is twice here and line 81

CarmenMiranda: Yes, just minutes ago I saw this, I'm doing a DB to check it with the overwrite included

CarmenMiranda: Yes, that was the issue, I already delete this line :smile:

hugho-ad (approved):

partner_credit_limit/models/account_invoice.py
```diff
@@ -19,17 +19,14 @@ def check_limit_credit(self):
             return True
         for invoice in self:
             allowed_sale = invoice.partner_id.with_context(
```
luisg123v: This logic is duplicated in the method `_check_credit_limit`, which is called just below. I think this should be removed and used instead the returning value of that method

hugho-ad: @luisg123v 

maybe just getting ride off  the 
```python
if self.allowed_sale:
    return False
```
at _check_credit_limit.

luisg123v: That's also possible. However what I really don't like of that (and the current code) is the unnecessary use of `with_context` twice with the same parameters

CarmenMiranda: Those parameters are being used on the `compute` method of the fields `credit_overloaded` and `overdue_credit`

luisg123v: Yes, that's why I think only a call to `check_credit_limit` is required, no need to compute it again here.

CarmenMiranda: I get ride of
```python
if self.allowed_sale:
    return False
```
on both methods (`sale` and `account_invoice`)

luisg123v (changes requested):
Well done!

Before your change, this module was a mess, had lints, etc. Now it's much more better!

luisg123v:
@CarmenMiranda,

If you have a time, could you cherry-pick this to v11, please?

CC @hugho-ad

Issue #1422, opened 2020-06-16
Branch 12.0 is red


luisg123v:
Both lint and tests are broken

luisg123v:
First step solved on https://github.com/Vauxoo/addons-vauxoo/pull/1385

luisg123v:
Lints solved on https://github.com/Vauxoo/addons-vauxoo/pull/1423, only tests are left

PR #1423, opened 2020-06-16
[FIX] *: Lint and flake8 issues

luisg123v:
The following modules are fixed, one per commit:
- account_anglo_saxon_stock_move
- account_anglo_saxon_stock_move_purchase
- account_anglo_saxon_stock_move_sale
- account_asset_move_check
- account_currency_tools
- account_invoice_line_asset_category_required
- account_move_line_production
- mrp_workcenter_account_move
- product_extended_segmentation
- product_extended_segmentation_byproduct
- purchase_third_validation
- stock_cost_segmentation
- stock_move_entries

Partially solves https://github.com/Vauxoo/addons-vauxoo/issues/1422


luisg123v:
@moylop260,

what do you think about this change?

I know it's kind of big, but:
- AFAIK none of these modules is used, they were migrated for the customer perfection, which won't be implemented anyway
- Some of them currently don't work anyway.

CC @hugho-ad

luisg123v:
@moylop260 

- And it's one commit per module, which will make easy to revert one of them if required

moylop260:
@hugho-ad @lescobarvx
Could you check it, please?

account_anglo_saxon_stock_move/model/invoice.py
```diff
@@ -235,10 +232,10 @@ def reconcile_stock_accrual(self, rec_ids, query_col, do_commit=False):
         query = self._get_accrual_query(query_col, 'query4', query_params)
         self._cr.execute(query)
 
-        if query_col == 'sale_id':
-            obj = self.env['sale.order']
-        elif query_col == 'purchase_id':
-            obj = self.env['purchase.order']
+        obj = self.env[
```
hugho-ad: single row

account_asset_move_check/wizard/wizard_asset_depreciation.py
```diff
@@ -38,10 +12,10 @@ class WizardAssetDepreciation(models.TransientModel):
         help='Select date stop to depreciation lines that '
         'will write that the lines are historical', required=True)
 
-    def write_historical_true(self, asset):
+    def write_historical_true(self):
         date_start = self.date_start
         date_stop = self.date_stop
-        for asset in asset['active_ids']:
+        for asset in self._context['active_ids']:
```
hugho-ad: self.env.context

account_currency_tools/demo/account_move.xml
```diff
@@ -1,329 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<odoo>
-        <record id="res_partner_23" model="res.partner">
```
hugho-ad, line 3: why we delete this file?
it's not used?

account_anglo_saxon_stock_move/model/invoice.py
luisg123v: What do you mean?

hugho-ad:
LGTM

account_anglo_saxon_stock_move/model/invoice.py
hugho-ad: all the statement in in single line instead of

luisg123v: Done. Also moved to a separate method, because this same code was duplicated several times.

account_asset_move_check/wizard/wizard_asset_depreciation.py
luisg123v: They're the same, but done.

account_currency_tools/demo/account_move.xml
luisg123v: Yes, it's not used from a long time ago, it was even already commented [when migrated](https://github.com/vauxoo/addons-vauxoo/commit/5fc63ace4842427e3a4b200d2c066cc479846840#diff-19edcbfc7004f63abf1e9b9c1d857562L34)

luisg123v:
@hugho-ad,

Do you have any other comment? Or can we merge this already?

hugho-ad:
:+1:

hugho-ad (approved):

PR #1424, opened 2020-07-03
[FIX] stock_by_warehouse: Quantity Available

CarmenMiranda:
Available quantities were being retrieved not only for the exact location but also for their child locations. That caused some product
to be double-counted when computing quantities.

This commit fixes the above.

### DIFF
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -43,38 +43,20 @@ class ProductProduct(models.Model):
         return domain_quant
 
     @api.multi
-    def _compute_product_available_not_res_dict(self):
+    def _compute_qty_available_not_reserved(self):
+        """Method to set the quantity available that is not reserved of a serie of products.
+        """
+        domain_quant = self._prepare_domain_available_not_reserved()
+        quants = self.env['stock.quant'].with_context(lang=False).read_group(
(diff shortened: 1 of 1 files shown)


stock_by_warehouse/models/product_product.py
```diff
@@ -43,38 +43,20 @@ def _prepare_domain_available_not_reserved(self):
...
 
-            vals['qty_available_not_res'] = quantity
-            res[prod.id] = vals
-        self._product_available_not_res_hook(quants)
-
-        return res
+        res = {}
+        for quant in quants:
+            quantity = quant.get('quantity') - quant.get('reserved_quantity')
+            res[quant.get('product_id')[0]] = quantity
```
luisg123v: Could you add the quantity only if it's positive

CarmenMiranda: `res[quant.get('product_id')[0]] = quantity > 0 and quantity or 0.0`

or

```
if quantity > 0:
    res[quant.get('product_id')[0]] = quantity
```

luisg123v: The first one but shorter:

`res[quant.get('product_id')[0]] = max(quantity, 0.0)`

CarmenMiranda: Done :smile:

luisg123v (approved):

PR #1425, opened 2020-07-07
[FIX] product_lifecycle: some msg imp, and check virtual available instead of qty available

GavyMG:
[FIX] product_lifecycle: some msg imp, and check virtual available instead of qty available

### DIFF
--- a/product_lifecycle/views/product_view.xml
+++ b/product_lifecycle/views/product_view.xml
@@ -19,7 +19,16 @@
             </xpath>
         </field>
     </record>
-
+    <record id="product_product_tree_view" model="ir.ui.view">
+        <field name="name">product.lifecycle.product.tree.view</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_product_tree_view"/>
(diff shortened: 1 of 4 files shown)


product_lifecycle/views/product_view.xml
```diff
@@ -19,7 +19,16 @@
             </xpath>
         </field>
     </record>
-
+    <record id="product_product_tree_view" model="ir.ui.view">
+        <field name="name">absa.product.product.tree.view</field>
```
luisg123v: absa?

product_lifecycle/views/product_view.xml
```diff
@@ -19,7 +19,16 @@
...
+            <field name="model">product.product</field>
```
luisg123v: indentation

luisg123v (changes requested):

product_lifecycle/views/product_view.xml
GavyMG: jajaja tnks

luisg123v (approved):

PR #1426, opened 2020-07-13
[FIX] stock_by_warehouse: Set Most Quantity Location

CarmenMiranda:
When there is no available location for a warehouse the set of the most quantity location was giving an error, now if there is no location available a zero will be set indicating it.

**ERROR**
Error Stock by Warehouse


CarmenMiranda:
@luisg123v Could you review this? Please

luisg123v (approved):

PR #1339, opened 2019-03-07
[FIX] default_warehouse_from_sale_team: When deleting a user from a

hugho-ad:
sales team, their sales_team id got empty, making the user failing on
the constraint since it's empty then is not on the list of ids

PR #1429, opened 2020-07-17
[FIX] default_warehouse_from_sale_team: Default Sales Team Constraint

CarmenMiranda:
### [T#40491](https://www.vauxoo.com/web#id=40491&view_type=form&model=project.task)

When a user is created without a default sales team or if a user is being deleted the field `sale_team_id` will be set to False, the constraint will fail since the value False is not on the `sale_team_ids` list.


CarmenMiranda:
Hi @luisg123v Could you review this? Please :smile:

luisg123v (approved):
CC @hugho-ad, this is a forward-port of https://github.com/Vauxoo/addons-vauxoo/pull/1339, plus a little improvement on the docstring

PR #1431, opened 2020-08-01
[FIX] *: Failure in tests

### LINKED ISSUE #1422: Branch 12.0 is red

luisg123v:
The following changes are performed:
- Don't test modules that depend on other ones from enterprise
- The module `account_currency_tools` depends on Python package pandas.
  It's now installed manually (not through requirements.txt) to don't make
  it to be installed on every project depending on this repo
- Don't use MXN in tests
- Adapt tests for v12, fixing code and using the `Form` class

Closes #1422


luisg123v:
@moylop260 Finally green! What do you think?

Note:
Several commits are intentional.

moylop260 (approved):

PR #1434, opened 2020-08-03
[FIX] partner_credit_limit: New Raise Limit Credit Message (#1421)

CarmenMiranda:
### Cherry-pick from v12: #1421 

Adding new messages to allow the user know the real reason why its invoice or sale order can't be validates/confirmed. The reasons can be if the partner has late payments, has exceeded the credit limit or both.


CarmenMiranda:
Hi @luisg123v This is the MR with the cherry-pick from the v12.0 :smile:

luisg123v (approved):

PR #1439, opened 2020-08-30
[ADD] Added new module to allow multi-company users see which company…

nhomar:
… their accounting records belong. (#1288)


moylop260 (approved):
:+1:

PR #1440, opened 2020-08-30
[ADD] Added new module to allow multi-company users see which company…

nhomar:
… their accounting records belong. (#1288)


luisg123v:
I won't include this module in the new branch's cleaned version, because:
- There's already a module in the OCA that does the same, called [res_company_code](https://github.com/OCA/multi-company/tree/12.0/res_company_code), which BTW it's more complete (it considers the company code on the company switcher) and efficient (it uses an SQL constraints instead of a o(n^2) search)
- It's not actually fully migrated, not even version was dumped.

If you still want me to migrate it despite the above, please let me know

CC @moylop260 @hugho-ad

PR #1444, opened 2020-09-05
[FIX] *: Remove unnecessary headers

luisg123v:
This removes:
- Hencoding headers, fixing lint `utf8-coding-comment`
- Headers like `# Part of Odoo`
- And other ones


luisg123v:
@moylop260 @hugho-ad 

What do you think?

Note:  
Separated commits are intentional, to be able to forward-port if necessary.

moylop260:
LGTM

PR #1445, opened 2020-09-10
[REM] `purchase_order_type` module task#37211

CarmenMiranda:
### [T#37211](https://www.vauxoo.com/web#id=37211&view_type=form&model=project.task)

No migrated module to version 11 and this module already exists on the repo `purchase-workflow` on Vauxoo.

[Repo](https://github.com/Vauxoo/purchase-workflow/commit/6771e3f98c2103d6f23b40f4d2cb7de6e6fd5b40)

PR #1446, opened 2020-09-10
[REM] *: Remove uninstallable modules

luisg123v:
Most of the modules are not installable, they come from previous
versions and were never migrated.

This removes all those, 307 in total.

The complete list of removed modules may be checked [here](https://github.com/Vauxoo/addons-vauxoo/files/5202926/deleted_modules.txt)


luisg123v:
@hugho-ad @moylop260 @hbto 

Could you review, please?

CC @JulioSerna @dsabrinarg @CarmenMiranda

moylop260:
This is a old matter
Please, confirm with @nhomar 
IMHO we shouldn't make these kind of changes in branches <=11.0 since that it will be deprecated soon.

nhomar:
@luisg123v 

This version will not be mainatined anymore in 1 month

hbto:
@moylop260 & @nhomar Yet this version won't die overnight and it will be still be used by some customers thus the cleaning is still neeeded.

Regards.

luisg123v:
@moylop260 and @nhomar 

In addition, this already forced us to rollback a deployment into production due to conflicts (see https://github.com/Vauxoo/addons-vauxoo/pull/1445), so we don't want this to happen again.

moylop260:
@luisg123v 
I got it

The main description of this PR is not justify the reason of #1445 

Could you explain it better since that you have a good point?

luisg123v:
I've meet with @nhomar, @hugho-ad and @carlosrve regarding this.

We accorded that all changes performed to branches from 11.0 will be reverted, to keep it's stability and to follow a guideline (unknown to most of us so far) that only the two latest version will be actively maintained, and the oldest one only until january of the year Odoo dropps support for it.

An email will be sent informing the above.

PR #1449, opened 2020-09-29
[FIX] mrp_partial_production: Missing quant reservation when producing t#11323

hugho-ad:
[FIX] mrp_partial_production: Missing quant reservation when producing
    less than the qty reserved
    
    Before this commit when you recorded production less than the qty you
    reserved on raw materials, the stock quants did not get reserved with
    the remaining qty from previous reservation but inventory lines.
    
    This causes a conflict when trying to cancel/unreserve the mrp
    production getting following error:
    
    It is not possible to unreserve more products of %s than you have in stock.
    
    After this commit, now quants reservation it's consistent with inventory
    reservation.

### DIFF
--- a/mrp_partial_production/models/stock.py
+++ b/mrp_partial_production/models/stock.py
@@ -15,10 +15,11 @@ class StockMove(models.Model):
             lambda aml: aml.qty_done > 0 and
             aml.product_uom_qty > aml.qty_done)
         aml_vals = []
+        available_quantity = 0
         for aml in amls:
+            product_uom_qty = aml.product_uom_qty - aml.qty_done
             vals = {
                 'product_id': aml.product_id.id,
-                'product_uom_qty': aml.product_uom_qty - aml.qty_done,
                 'product_uom_id': aml.product_uom_id.id,
                 'workorder_id': aml.workorder_id.id,
                 'location_id': aml.location_id.id,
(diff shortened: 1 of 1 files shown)


mrp_partial_production/models/stock.py
```diff
@@ -28,10 +29,14 @@ def _split(self, qty, restrict_partner_id=False):
...
+            available_quantity += product_uom_qty
             aml_vals.append(vals)
         new_move_id = super(StockMove, self)._split(qty, restrict_partner_id)
         for vals in aml_vals:
             vals['move_id'] = new_move_id
             self.env['stock.move.line'].create(vals)
-        self.browse(new_move_id)._recompute_state()
+        new_move = self.browse(new_move_id)
+        new_move._update_reserved_quantity(
+            new_move.product_uom_qty, available_quantity, new_move.location_id, strict=False)
```
luisg123v: Are you sure it should be strict=False?

hugho-ad: yes, we got it from after debugging to following line
https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_move.py#L950

luisg123v (approved):
LGTM, just provide a commit description, please.

hugho-ad:
@luisg123v 

as we talk, now we ensure reserving inventory materials on strict mode since we should keep the same inventory when producing less than the reserved and avoid re-reserve on another child locations

hugho-ad:
@keylor2906 @luisg123v  @JulioSerna 
please review

luisg123v (approved):
:+1:

CC @belisac

keylor2906:
@hugho-ad Missing line at EOF, but LGTM

keylor2906 (approved):

hugho-ad:
@keylor2906 Done!
thanks!

hugho-ad:
@JulioSerna 

Your comments are welcome

luisg123v:
This has been approved by the customer, so merging.

I'll squash because the 2nd commit is part of the reservation and it was a required change we noticed after the fix was implemented.

hugho-ad:
:+1:

PR #1451, opened 2020-10-02
[REF] stock_by_warehouse: Stock per locations Compute

CarmenMiranda:
### [T#39295](https://www.vauxoo.com/web#id=39295&view_type=form&model=project.task)

Separating the obtain of the locations and their available quantities by warehouse, so the compute `_compute_get_stock_location` is less complicated, also, this information can be used from other methods.

### DIFF
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -96,6 +96,42 @@ class ProductProduct(models.Model):
             })
         return json.dumps(info)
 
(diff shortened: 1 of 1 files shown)


CarmenMiranda:
Hi, @luisg123v Could you review this?
I did it because I need the information about the locations and quantity in another widget (Private repo), and to not repeat that code I want to use this that we already have :see_no_evil:

PR #1452, opened 2020-10-02
[FIX] stock_by_warehouse: Making the get of qty per location private

CarmenMiranda:
Because we are using `sudo()` we need to make the method called `get_qty_per_location` private.


CarmenMiranda:
Hi @luisg123v Sorry about the error, this is the fix to make the method private

luisg123v (approved):

luisg123v:
@CarmenMiranda It's OK, I didn't notice it neither.

CC @nhomar

PR #1453, opened 2020-10-03
[FIX] default_warehouse_from_sale_team: Field name in rules

luisg123v:
When migrated, field was renamed from:
`default_warehouse` -> `default_warehouse_id`
but record rules were not updated.


luisg123v:
@suniagajose could you review, please?

suniagajose (approved):

PR #1447, opened 2020-09-27
[MIG] default_warehouse_from_sale_team: Migrate to v13.0

suniagajose:

PR #1454, opened 2020-10-04
[MIG] default_warehouse_from_sale_team: Migrate to v13

luisg123v:
The following is performed:
- Replace account invoice model by account move
- import openerp -> import odoo
- Fix class names
- Fix indentation in data files
- Adapt tests


luisg123v:
This is pretty much the same done in https://github.com/Vauxoo/addons-vauxoo/pull/1447 and https://github.com/Vauxoo/addons-vauxoo/pull/1453, so merging.

CC @suniagajose

PR #1458, opened 2020-10-06
[MIG] account_multicompany_code: Migrate to v14

luisg123v:
The following is performed:
- Adapt tests
- Update translation terms


luisg123v:
Even though it's red because coverage decreased, it's because coverage is not 100% anymore, but 97, wich is more than acceptable.

In addition, the non-covered part is an error message that will be migrated to an SQL constraint anyway.

PR #1460, opened 2020-10-23
[MIG] stock_by_warehouse: Migration to 13.0

deivislaya:
The following is performed:
- Remove decimal precision in float fields
- Remove all the decorators @api.multi
- Added @api.depends_context in non-stored computer methods, which
  use context in calls and its super method also uses it


stock_by_warehouse/models/product_template.py
```diff
@@ -0,0 +1,73 @@
...
+    @api.onchange('warehouses_stock_recompute')
+    def _warehouses_stock_recompute_onchange(self):
+        if not self.warehouses_stock_recompute:
+            self.warehouses_stock_recompute = True
+            return
+        self.warehouses_stock = self._compute_get_quantity_warehouses_json()
+        self.warehouses_stock_recompute = True
+
+    @api.depends('product_variant_ids.qty_available_not_res')
+    @api.depends_context('warehouse', 'force_company')
```
luisg123v, line 30: Why this?

luisg123v (changes requested):
Translations also need migration

stock_by_warehouse/models/product_template.py
deivislaya: Because, now the computed field for the product quantities in [product.template](https://github.com/odoo/odoo/blob/13.0/addons/stock/models/product.py#L617) and [product.product](https://github.com/odoo/odoo/blob/13.0/addons/stock/models/product.py#L96), uses `@api.depends_context`, when using this decoration, it only recognizes the keys mentioned in it, for the calls that use context.

Also, it is necessary to filter by warehouses and companies, otherwise, it takes the general quantity for all warehouses

luisg123v: @CarmenMiranda could you check @deivislaya's comment regarding "it is necessary to filter by warehouses and companies, otherwise, it takes the general quantity for all warehouses", please?

deivislaya:
@JulioSerna 

Please review this MR

Regards!

cc @nhomar @hugho-ad

hugho-ad (approved):

hugho-ad:
@JulioSerna @nhomar 

could you please review


Shall we add more tests?

stock_by_warehouse/models/product_template.py
hugho-ad: @CarmenMiranda could you please review

CarmenMiranda: I'm not fully aware of the `@api.depends_contex` but as I see it on the @deivislaya examples is need it, but, also I saw that there was a [commit](https://github.com/odoo/odoo/commit/de2725c173cfe028c3c0f411e613ba732be14772) that removed the `warehouse` key from the `@api.depends_contex` 
 of the method `_compute_quantities`. So I'm not sure if that key really needs to be here or if it can be a problem with it.

deivislaya: Yes, they eliminated it because, from the current code of odoo, they no longer need to obtain the search for warehouses in a single attempt, in the field

JulioSerna:
👍 

tested with @hugho-ad and @deivislaya 

I just tested that the quantity it is displayed into product by company and depends on the companies available by the user

@ljpachec

luisg123v (approved):
Even though coverage decreased, it's expected because there currently are very few modules and those have a high coverage. Besides, this module does include tests.

PR #1461, opened 2020-10-24
    [MIG] sale_margin_percentage: Migration to 12.0

angelJuaGar:
The following is performed:
- Added group for managing sales percentages.
- Updated translations and test modifications.
- Modifications made for test execution.
- Added the 'for' attribute inside the labels.


sale_margin_percentage/i18n/es.po
```diff
@@ -0,0 +1,78 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* sale_margin_percentage
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 11.0+e\n"
```
luisg123v: Translations also needs migration. Take into account syntax changed for model terms from 11.0 -> 12.0.

sale_margin_percentage/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,26 @@
...
+    <record id="view_res_sale_margin_config_settings_inherit" model="ir.ui.view">
+        <field name="name">res.sale.margin.percentage.settings</field>
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="sale.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[@data-string='Sales']//div[hasclass('o_settings_container')][2]" position="inside">
+                <div class="col-xs-12 col-md-6 o_setting_box" attrs="{'invisible': [('module_sale_margin' ,'=', False)]}" groups="base.group_no_one">
+                    <div class="o_setting_left_pane"/>
+                    <div class="o_setting_right_pane">
+                        <label for="label_margin_threshold" string="Margin Threshold"/>
```
luisg123v: The `for` attribute should point to the involved field:

```suggestion
                        <label for="margin_threshold" string="Margin Threshold"/>
```

luisg123v (changes requested):

luisg123v:
Hi @angelJuaGar

CI is red. Could you fix it, please?

sale_margin_percentage/i18n/es.po
angelJuaGar: Change done.

sale_margin_percentage/views/res_config_settings_views.xml
angelJuaGar: Change done.

angelJuaGar:
Hi @luisg123v 
Fixed up. Could you review please

luisg123v (changes requested):
Hi @angelJuaGar,

I think translations are not migrated yet. As I commented before, term syntax changed from 11.0 -> 12.0. For instance, the line:  
`#: model:ir.model.fields,field_description:sale_margin_percentage.field_res_company_margin_threshold`

Now would be something like:  
`#: model:ir.model.fields,field_description:sale_margin_percentage.field_res_company__margin_threshold`

The easiest way to solve this IMHO is to re-export translations and paste the result.

Could you check, please?

sale_margin_percentage/tests/test_sale.py
```diff
@@ -0,0 +1,164 @@
...
+                })],
+            })
+            sale_line = sale.order_line
+            self.assertEqual(sale_line.margin, 48)
+            self.assertEqual(sale_line.margin_percentage, 100.0)
+
+    def test_02_margin_sale_price_0(self):
+        """ Margin and Margin Percentage when sale price is 0.0
+        """
+        with self.env.do_in_onchange():
```
luisg123v: Use the Form class, please, [here](https://github.com/Vauxoo/addons-vauxoo/blob/8c9a1597b4e3bd3904ceca75ecd8cef24c7862ec/purchase_third_validation/tests/test_purchase_third_validation.py#L30) an example.

This comment is applicable for the below tests

angelJuaGar:
Hi @luisg123v 
Changes have been made. You can check?

sale_margin_percentage/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+            if not line.product_uom_qty:
+                line.margin_percentage = 0.0
+                continue
+
+            if not line.price_unit or line.price_subtotal == 0.0:
+                line.margin_percentage = -100.0
+                continue
+
+            purchase_price = \
+                line.purchase_price or line.product_id.standard_price
```
hugho-ad: doit in a single row  please

hugho-ad:
LGTM :+1: 

@JulioSerna  could you review

hugho-ad:
@angelJuaGar 

please squash your commits in a single one

angelJuaGar:
Done, already combine all commits into one.

angelJuaGar:
Hi @luisg123v
You can review it?

sale_margin_percentage/i18n/es_CR.po
```diff
@@ -0,0 +1,107 @@
+# Translation of Odoo Server.
```
luisg123v: `es.po`, please.

Country-specific languages should only be included when their terms differ from the base ones. In addition, you deleted `es.po`

sale_margin_percentage/models/res_config_settings.py
```diff
@@ -0,0 +1,11 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class ResConfiguration(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    margin_threshold = fields.Float(
+        related='company_id.margin_threshold', help="Margin Threshold for product in order lines", readonly=False)
```
luisg123v: Remove help, it's already taken from the related field

sale_margin_percentage/security/sale_margin_percentage_security.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
luisg123v, line 1: Shouldn't this be noupdate?

sale_margin_percentage/tests/test_sale.py
```diff
@@ -0,0 +1,177 @@
+import time
+from odoo.tests.common import TransactionCase
+from odoo.tests import tagged
```
luisg123v, line 3: `from odoo.tests import tagged, TransactionCase`

luisg123v (changes requested):
Hi @angelJuaGar,

This [comment](https://github.com/Vauxoo/addons-vauxoo/pull/1461/#pullrequestreview-520168704) is not applied yet.

sale_margin_percentage/security/sale_margin_percentage_security.xml
angelJuaGar: Why should it be 'noupdate=true'?

luisg123v: Data is usually noupdate so it may be edited functionally, otherwise changes would be lost next time the module is updated.

luisg123v:
BTW @angelJuaGar you squashed correctly commits, but you didn't update commit message.

sale_margin_percentage/security/sale_margin_percentage_security.xml
angelJuaGar: ok, already apply the change.

angelJuaGar:
Hi @luisg123v 
The latest changes have been applied.
Can you check it please?

angelJuaGar:
Hi @deivislaya, @suniagajose 
Could you review, please?

It's the migration of the module from v11 to v12.

hugho-ad:
@deivislaya 


you can test it at https://git.vauxoo.com/vauxoo/alvarado/-/merge_requests/2

deivislaya:
@angelJuaGar 

Test functional  :+1: 

Perform the test with different cases of purchase_price and price_unit per sales line and the sum. The minimum margin percentage allowed to confirm the sale order was also verified.

@hugho-ad @suniagajose

hugho-ad:
@luisg123v 

could you review please

luisg123v (approved):
@angelJuaGar LGTM, just a little comment regarding the commit message. It's now clear and complete, however, you still left messages from old commits, just remove these lines

```
[MIG] sale_margin_percentage: Changed applied. Added group for managing sales percentages.

[MIG] sale_margin_percentage: Modifications made for test execution.

[MIG] sale_margin_percentage: Updated translations and test modifications.
```

On the other hand, as an advice, I'd recommend you to get used to always rebase before pushing anything. Here is not as mandatory, because we may merge PRs even if they're not fully rebased (as long as there are no conflicts). But on GitLab, your branch needs to be up-to-date in order for it to be mergeable.

Good job here.

sale_margin_percentage/tests/test_sale.py
luisg123v: @hugho-ad AFAIK you helped @angelJuaGar to use the Form class, but it's not in the tests. Did you decided to don't use it?

hugho-ad: @angelJuaGar 

which problem did you get using form?

I remember you told me that onchange function didn't be called right?

angelJuaGar: It is right. 
Using the form class, it does not detect the onchange function, applying a different process in the calculation of the percentage and causing different results than those expected in the test.

luisg123v: It should be quite the opposite, the Form class is used in order to get onchanges triggered automatically, instead of calling them manually. Actually, using the Form class, the result should be closer to the functional tests.

angelJuaGar: I understood that when it was recommended to me to use the form class but when applying it in the tests it was not reflected that the onchange was activated.
The test was replicated in a functional way to be able to check the results and they were incorrect.

angelJuaGar: Hi @luisg123v
Is it okay to handle the test this way?

luisg123v: @angelJuaGar not ideal, but it was that way, it's working as it is and I understand make it to work using the Form class could be a little bit complex, so It's OK for me.

PR #1463, opened 2020-10-30
[FIX] default_warehouse_from_sale_team: avoid sequence re-call

suniagajose:
### Resume
The sequence is calling two times, one from original model and another from 'default.warehouse' inherit, this fix avoid the
re-call to next_sequence number if it was called previously, this is just by checking if the name has been already set.

Before:
- Current Purchase Order P00039, next P000041.
Now:
- Current Purchase Order P00041, next P00042.

Solicitudes de presupuesto - Odoo

### Debugging:
Captura de pantalla de 2020-10-30 00-07-08
Captura de pantalla de 2020-10-29 23-46-47

### DIFF
--- a/default_warehouse_from_sale_team/models/sales_team.py
+++ b/default_warehouse_from_sale_team/models/sales_team.py
@@ -64,7 +64,7 @@ class WarehouseDefault(models.Model):
     def create(self, vals):
         sequence_obj = self.env['ir.sequence']
         pick_type_obj = self.env['stock.picking.type']
-        if vals.get('warehouse_id', 'picking_type_id'):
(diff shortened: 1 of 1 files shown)


luisg123v (approved):

hugho-ad (approved):

hugho-ad:
@JulioSerna 

just missing you review

suniagajose:
Hi @JulioSerna 

Could you help me with you review... 

Regards,

cc @hugho-ad 
cc @luisg123v

luisg123v:
I think this change is safe enough, so I'll merge.

suniagajose:
Thanks @luisg123v

PR #1465, opened 2020-11-08
[MIG] web_widget_json_graph: migration to 12.0

luisg123v:


luisg123v:
@moylop260 @hugho-ad @frahikLV 

could you review, please?

I'm migrating this module from the web repo, because that module was actually never in the OCA, so ours is diverging without needing to.

hugho-ad:
why not post it at oca/web?

luisg123v:
@hugho-ad because it has never belonged there.

It could be added, but it would be a new module.

hugho-ad:
@luisg123v 

>>It could be added, but it would be a new module.
New module name?

I think this we should move it there, at oca/web

luisg123v:
@hugho-ad same module name, but new module for the OCA

moylop260:
I think It is not a bad idea to move this module to OCA

luisg123v:
@moylop260 @hugho-ad PR to the OCA created: https://github.com/OCA/web/pull/1733. Authorship has been taken care of.

I'm closing this one.

PR #1469, opened 2020-11-25
[ADD] stock_by_warehouse: add context for multi-warehouse

GavyMG:
For example: when we want to show only the sale warehouse and its resupply warehouses on the sale order line

https://git.vauxoo.com/absa/absa/-/merge_requests/1326

### DIFF
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -77,11 +77,20 @@ class ProductProduct(models.Model):
         # itself, we enable this context management
         warehouse_id = self._context.get('warehouse_id')
 
-        for warehouse in self.env['stock.warehouse'].sudo().search([]):
+        # Just in case it's asked from only a group of warehouses
+        warehouse_ids = self._context.get('warehouse_ids')
+        some_wh = False
+        if warehouse_ids:
+            info['warehouse'] = 0.0
+            some_wh = not warehouse_id
+
+        for warehouse in warehouse_ids or self.env['stock.warehouse'].sudo().search([]):
             product = self_origin.sudo().with_context(
                 warehouse=warehouse.id, location=False)
             if warehouse_id and warehouse_id.id == warehouse.id:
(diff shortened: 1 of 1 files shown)


stock_by_warehouse/models/product_product.py
```diff
@@ -77,11 +77,20 @@ def _compute_get_quantity_warehouses_json(self):
         # itself, we enable this context management
         warehouse_id = self._context.get('warehouse_id')
 
-        for warehouse in self.env['stock.warehouse'].sudo().search([]):
+        # Just in case it's asked from only a group of warehouses
+        warehouse_ids = self._context.get('warehouse_ids', False)
+        some_wh = False
```
luisg123v, line 82: This variable is not necessary, is it?

stock_by_warehouse/models/product_product.py
```diff
@@ -77,11 +77,20 @@ def _compute_get_quantity_warehouses_json(self):
         # itself, we enable this context management
         warehouse_id = self._context.get('warehouse_id')
 
-        for warehouse in self.env['stock.warehouse'].sudo().search([]):
+        # Just in case it's asked from only a group of warehouses
+        warehouse_ids = self._context.get('warehouse_ids', False)
```
luisg123v:
```suggestion
        warehouse_ids = self._context.get('warehouse_ids')
```

luisg123v (changes requested):

stock_by_warehouse/models/product_product.py
GavyMG: To avoid comparing if not warehouse_id and warehouse_ids in every iteration.

moylop260:
@JulioSerna 

Could you check it, please'

stock_by_warehouse/models/product_product.py
GavyMG: @luisg123v 

could we merge this?? (You can commit the suggestion)

luisg123v (approved):

PR #1471, opened 2020-12-12
[MIG] account_chatter: Migration to 13.0 vx#43062

deivislaya:
Migration standard


account_chatter/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+1. Settings > Users & Companies > Users
+
+2. Select any user you can share account chatter notifications
+
+3. Accounting > Check > Show account notifications
+
+Account:
+
+    .. image:: static/description/chatter_in_account.png
```
luisg123v: Does this display the image correctly when the module is shown functionally?

account_chatter/models/account_journal.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Vauxoo
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models, fields
+
+
+class AccountJournal(models.Model):
+
+    _inherit = "account.journal"
```
luisg123v, line 9: Why this? Does the journal include chatter by default in newer versions? AFAIK, no

account_chatter/views/account_move_views.xml
```diff
@@ -0,0 +1,26 @@
+<odoo>
+    <record id="view_move_form_inherit" model="ir.ui.view">
+        <field name="name">account.move.form.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form"/>
+        <field name="groups_id" eval="[(4, ref('account_chatter.group_show_account_chatter_notifications'))]"/>
+        <field name="arch" type="xml">
+            <xpath expr="//sheet" position="after">
+                <div class="oe_chatter">
+                    <field name="message_follower_ids"/>
```
luisg123v, line 11: I don't get this change. Could you explain, please?

luisg123v (changes requested):

account_chatter/README.rst
deivislaya: Fixed!

Thanks
 
Odoo (1)

account_chatter/models/account_journal.py
deivislaya: Yes, now in V13 the journal include [chatter](https://github.com/odoo/odoo/blob/13.0/addons/account/models/account.py#L681) by default

account_chatter/views/account_move_views.xml
deivislaya: This change is to show these fields only with the group ```account_chatter.group_show_account_chatter_notifications```. Why by default the chatter is hidden from all users

luisg123v: ok, But why the class d-none and that stuff?

Wouldn't be simpler to add the fields and add the desired group as an attribute?

deivislaya: > ok, But why the class d-none and that stuff?

Because the chatter must be hidden by default, if the user does not belong to the group

> Wouldn't be simpler to add the fields and add the desired group as an attribute?

Because the fields (chatter) are already shown by default in odoo. This module now hides them by default and they are only visible through the group. Reading in forums and testing the chatter does not respond to the attribute `` groups '' to be displayed, in recent versions

luisg123v: > Because the fields (chatter) are already shown by default in odoo. This module now hides them by default and they are only visible through the group

That's why I ask, why not simply adding the attribute group="group_can_see_chatter"? That would show the chatter only for members of that group. It doesn't need to be used on the fields themselves, it could be applied in the whole chatter dif.

deivislaya: Yes, this was my first option but it did not work with the attribute group, neither for this fields nor for the div, reading I found this [forum](https://www.odoo.com/es_ES/forum/help-1/visibility-of-oe-chatter-only-form-employee-manager-how-to-141797)

randall-vx:
@luisg123v do you have any news about this MR ?

cc: @deivislaya

randall-vx:
Hi @luisg123v do you have any news about thos PR ?

deivislaya:
@luisg123v 

Please review again

Regards!

luisg123v:
Hi @randall-vx,

I meet yesterday with @deivislaya to review this, and we accorded some changes.

@deivislaya are those changes already applied? If so, I wasn't aware

deivislaya:
@luisg123v 

yes, the new changes already applied. I missed notifying you in a timely manner

@randall-vx

luisg123v (approved):

randall-vx:
Thanks ! @deivislaya please fix the MR in 14.0 in order to @luisg123v instructions.

Regards,

PR #1474, opened 2021-01-07
[MIG] internal_transfer_with_agreed_amount: Migration to 14.0 task #43948

angelJuaGar:


internal_transfer_with_agreed_amount/i18n/es.po
```diff
@@ -0,0 +1,120 @@
...
+
+#. module: internal_transfer_with_agreed_amount
+#: model_terms:ir.ui.view,arch_db:internal_transfer_with_agreed_amount.view_internal_transfer_multicurrency
+msgid "Fill with the total amount that you will receive for this operation"
+msgstr "Complete con el monto total que recibirá por esta operación"
+
+#. module: internal_transfer_with_agreed_amount
+#: model:ir.model.fields,field_description:internal_transfer_with_agreed_amount.field_internal_transfer_multicurrency__id
+msgid "ID"
+msgstr "ID"
```
luisg123v: Did you add this manually?

When te translation is the same as the original term, it shouldn't be provided, because when exporting, it will be cleared anyway.

angelJuaGar: Ok, I already applied the correction.

hbto:
@angelJuaGar @luisg123v 

While technically this module could be properly migrated.

We must halt this module from installing as the functionality is not the same than in previous versions.


A functional design meeting must scheduled to check what will be the course of action for this module.

In the meantime, this module must be set a installable = False.

Regards.

luisg123v:
I don't like to have modules in code that are not actually installable, because it breaks the OCA procedure, only available modules are present.

I'll close this for now, I left a reference to this PR as a TODO in the Vauxoo's manifest.

nhomar:
This app will not be migrated at least for now, this message is just for the records.

PR #1473, opened 2020-12-17
[IMP] account_move_report: Add names to XML elements

luisg123v:
This is to be able to inherit the report on a deterministic manner.

PR #1475, opened 2021-01-08
[IMP] account_move_report: Add names to XML elements

luisg123v:
This is to be able to inherit the report on a deterministic manner.

Cherry-pick of https://github.com/Vauxoo/addons-vauxoo/pull/1473


luisg123v:
I'm merging because this is just a cherry-pick of https://github.com/Vauxoo/addons-vauxoo/pull/1473, which doesn't affect functionality.

PR #1478, opened 2021-02-16
[FIX] sale_margin_percentage: Avoid clear cache value computed - ticket#13578

moylop260:
The order of the methods could be executed of different way if the method doesn't depend of price_subtotal
Adding it you can be sure that price_subtotal was computed before to use this method
In order to avoid to reset the original value from database because this will be zero


hugho-ad:
DUMMY MR https://git.vauxoo.com/vauxoo/tanner/-/merge_requests/367

hugho-ad (approved):

hugho-ad:
functional review  :+1: 
image

JulioSerna:
👍

PR #1485, opened 2021-03-17
WIP: [MIG] mrp_partial_production: Migration to v14.0

luiseevaquer:


hugho-ad:
this is supported at v14.0, so no longer neede it

PR #1486, opened 2021-05-30
[ADD] stock_segmentation_cost_fraction: Module to compute segmentation cost

luisg123v:
This module implements a way to compute product costs by segments, to be
able
to break down the total cost in:
- Material cost
- Landed cost
- Production cost
- Subcontracting cost

To achieve the above, two concepts are introduced, namely fractions and
atomic
costs. For more info, please check the README file.


luisg123v:
@hbto could you review, please?

CC @belisac

PR #1487, opened 2021-05-30
[FIX] stock_landed_segmentation: Missing dependency to mrp

luisg123v:
This module adds a relationship between landed costs and unbuild orders
(with the field `unbuild_ids`), but it was missing the dependency to the
`mrp` module, where the unbuild model is defined.


luisg123v:
@hbto could you review, please?

hbto:
@luisg123v 

Looks good to me.

Regards.

PR #1493, opened 2021-07-24
[MIG] default_warehouse_from_sale_team: Migrate to 14.0

luisg123v:
Odoo now provides a native way to specify a default warehouse on the
user. So this module uses the same behavior, but giving priority if a
warehouse is set on the sales team.

Other tasks are also performed:
- Code and views are reorganized by model
- Models that were actually mixings are renamed to use the `mixing` suffix
- Test are improved to use the `Form` class
- Some Typos are fixed
- General code clean up


luisg123v:
@moylop260 what do you think?

luisg123v:
CC @desdelinux @belisac

moylop260:
👍

PR #1494, opened 2021-08-01
[FIX]  purchase_third_validation: Don't hide cancel i#15838

luisg123v:
The cancellation button was being hidden when the purchase order is in
"Third Approve" status.


luisg123v:
@hugho-ad could you review, please?

CC @belisac

belisac:
Hello @luisg123v 

Thanks for the change I tested and it has the behavior that its expect

Regards, 

cc: @hugho-ad 

PO-41832-Odoo - Updates -Commit - sin grupo

belisac (approved):

fernandahf:
+1

fernandahf (approved):

PR #1497, opened 2021-08-06
[FIX] invoice_discount: Include only invoice lines, not al move lines from the journal to compute discount totals

hugho-ad:


hugho-ad:
@moylop260 could you pls review

ebirbe:
:+1: LGTM

PR #1499, opened 2021-08-17
 [MIG] stock_by_warehouse_sale: Migration to 14.0

hugho-ad:


hugho-ad:
@deivislaya @CarmenMiranda @moylop260 

This module could be deprecated, but it show all the info about all warehouse instead of the odoo one that only shows up the stock on the warehouse on the SO

hugho-ad:
@moylop260 @deivislaya could you pls review

moylop260:
👍

deivislaya:
LGTM :+1:

CarmenMiranda:
> @deivislaya @CarmenMiranda @moylop260
> 
> This module could be deprecated, but it show all the info about all warehouse instead of the odoo one that only shows up the stock on the warehouse on the SO

Hi @hugho-ad, yes this widget shows us the specific qty in the sale's warehouse but if they click on the widget they can see the info of all the warehouses.

If you wish that the widget only show us the info of all the warehouse and omitting the warehouse, it can be done by removing [this context](https://github.com/vauxoo-dev/addons-vauxoo/blob/14.0-mig-stock_by_warehouse_sale-dev-hugho-ad/stock_by_warehouse_sale/models/sale_order.py#L14)

Also, omitting the warehouse in the [view](https://github.com/vauxoo-dev/addons-vauxoo/blob/14.0-mig-stock_by_warehouse_sale-dev-hugho-ad/stock_by_warehouse_sale/views/sale_view.xml#L11) and instead of having:
`The saleable stock in: <field name="warehouse_id"/> that can be delivered immediately is:`
We can have:
`The saleable stock that can be delivered immediately is:`

Also, the field `warehouse_id`  Odoo already has it, so it needs to be removed from [here](https://github.com/vauxoo-dev/addons-vauxoo/blob/744031a730abede2568b11276fd8571f8128306e/stock_by_warehouse_sale/models/sale_order.py#L8), and [here](https://github.com/vauxoo-dev/addons-vauxoo/blob/744031a730abede2568b11276fd8571f8128306e/stock_by_warehouse_sale/models/sale_order.py#L22) if you are going to remove it as I said in the previous paragraph (if not leave it because the related is not set until the save of the line).

Because indeed, as you said, Odoo already provides a widget for the stock of the sales' warehouse in the tree view of the line, although we aren't showing the same info, Odoo, only shows the available and not the incoming qty as we do
Review from PR 1499 - stock_by_warehouse_sale


And if is possible an update of the images in the readme would be great 🙈

hugho-ad:
@LeptonAlvarado 

pls, let do the improvements @CarmenMiranda propose in a separated mr


thanks!

stock_by_warehouse_sale/models/sale_order.py
```diff
@@ -5,11 +5,8 @@ class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
     warehouses_stock = fields.Text(store=False, readonly=True)
-    warehouse_id = fields.Many2one(
```
hugho-ad: remove it because 
https://github.com/odoo/odoo/blob/14.0/addons/sale_stock/models/sale_order.py#L275

stock_by_warehouse_sale/i18n/es.po
```diff
@@ -0,0 +1,42 @@
...
+#: model:ir.model,name:stock_by_warehouse_sale.model_sale_order_line
+msgid "Sales Order Line"
+msgstr "Línea de pedido de venta"
+
+#. module: stock_by_warehouse_sale
+#: model_terms:ir.ui.view,arch_db:stock_by_warehouse_sale.sale_order_form_view
+msgid "The saleable stock in:"
+msgstr "La existencia vendible en:"
+
+#. module: stock_by_warehouse_sale
```
CarmenMiranda, line 28: @hugho-ad  Just missing the removal of this translation

stock_by_warehouse_sale/i18n/es.po
```diff
@@ -0,0 +1,42 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* stock_by_warehouse_sale
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 11.0+e\n"
```
CarmenMiranda: @hugho-ad Maybe an update of version in here 🙈 That was missing from 12 too hahaha

stock_by_warehouse_sale/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Stock by Warehouse on Sales",
+    "version": "12.0.1.0.0",
```
CarmenMiranda: Also, in here a big bump in the version is needed @hugho-ad

stock_by_warehouse_sale/i18n/es.po
hugho-ad: thanks!

stock_by_warehouse_sale/__manifest__.py
hugho-ad: for sure

CarmenMiranda:
LGTM 👍🏼

PR #1503, opened 2021-09-17
[FIX] account_chatter: change category group

randall-vx:
* Change category group to avoid conflicts with account group

##### BEFORE
Screen Shot 2021-09-16 at 11 06 09 PM

##### AFTER
Screen Shot 2021-09-16 at 11 13 25 PM

### DIFF
--- a/account_chatter/security/account_chatter_security.xml
+++ b/account_chatter/security/account_chatter_security.xml
@@ -3,7 +3,7 @@
 
     <record id="group_show_account_chatter_notifications" model="res.groups">
         <field name="name">Show account chatter notifications</field>
-        <field name="category_id" ref="base.module_category_accounting_accounting"/>
+        <field name="category_id" ref="base.module_category_hidden"/>
         <field name="comment">Show notifications of accounting entries, journal and accounts reflecting changes and observations.</field>
     </record>
 


randall-vx:
@hugho-ad could you review this MR, please?

regards,

hugho-ad (approved):

account_chatter/security/account_chatter_security.xml
```diff
@@ -3,7 +3,7 @@
 
     <record id="group_show_account_chatter_notifications" model="res.groups">
         <field name="name">Show account chatter notifications</field>
-        <field name="category_id" ref="base.module_category_accounting_accounting"/>
+        <field name="category_id" ref="base.module_category_hidden"/>
```
nhomar, line 6: The module category Hidden is to set this groups as part of a setting, which mean if set the all users will be set directly. Is that what you meant?

randall-vx: @nhomar odoo has two categories:

* Technical Settings (base.module_category_hidden): Used for custom settings like create contacts, show SLA policies, access to export feature, etc.

* Hidden (base.module_category_dependecy_hidden): Used to set as a part of a setting.

luisg123v (approved):

account_chatter/security/account_chatter_security.xml
luisg123v: @randall-vx I though you were right, but @nhomar seems to be right.

The group to create contacts [has no category](https://github.com/odoo/odoo/blob/5820bde45017416d14245a64b7d3624a9d575fba/odoo/addons/base/security/base_groups.xml#L49), while a technical group (the one enabled for all users, i.e. to enable multi-location) [has the category "hidden"](https://github.com/odoo/odoo/blob/5820bde45017416d14245a64b7d3624a9d575fba/addons/stock/security/stock_security.xml#L12).

So, I think the proper fix is to leave the category empty. In the meantime, I will revert this change.

PR #1504, opened 2021-09-26
Revert "[FIX] account_chatter: change category group"

luisg123v:
This reverts commit a7e4b8c749f5d6a739efbe3010f9fb22c30475a1.

The "hidden" category is used for thechnical groups (the ones enabled/disabled
for all users), which is not the case for the account chatter group.


luisg123v:
Reverting PR https://github.com/Vauxoo/addons-vauxoo/pull/1503 according to https://github.com/Vauxoo/addons-vauxoo/pull/1503#discussion_r716232422

PR #1505, opened 2021-09-27
[FIX] stock_by_warehouse_sale: Edit SO line using form view i#16357

luisg123v:
In previous Odoo versions, the form view was poped up when editing a
sale order line, as long as the option "Manage Product packaging" was
enabled. This is no longer the case [1], so we re-enable it manually,
in order to be able to see current inventory when filling a line.

[1] https://github.com/odoo/odoo/commit/e813487e75f1

### DIFF
--- a/stock_by_warehouse_sale/views/sale_view.xml
+++ b/stock_by_warehouse_sale/views/sale_view.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
-    <record id="sale_order_form_view" model="ir.ui.view">
+    <record id="view_order_form" model="ir.ui.view">
         <field name="name">sale.order.stock.warehouse</field>
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form"/>
@@ -21,6 +21,9 @@
                             groups="stock.group_stock_multi_warehouses"/>
                     </p>
                 </div>
+            </xpath>
+            <xpath expr="//field[@name='order_line']/tree" position="attributes">
+                <attribute name="editable" />
             </xpath>
         </field>
     </record>


luisg123v:
@hugho-ad 

Could you check, please?

CC @desdelinux

hugho-ad (approved):

hugho-ad:
LGTM :+1:

hugho-ad:
@moylop260 

could you merge pls

JulioSerna:
@hugho-ad @luisg123v 

Why this feature it is added in this module?

stock_by_warehouse_sale/views/sale_view.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
-    <record id="sale_order_form_view" model="ir.ui.view">
+    <record id="view_order_form" model="ir.ui.view">
```
moylop260, line 4: Be careful changing `xmlid` since that it will need a migration script but there is not

hugho-ad: changing xml_id  of views without noupdate=1 will drop those views, and create them again, isn't it?

moylop260:
The commit referenced here:
 - https://github.com/odoo/odoo/commit/e813487e75f1

has the following warning:
 - The next version of the product configurator will only be compatible with inline tree view editing.

So, it could be a problem if we force a deprecated way for Odoo

I have a doubt here... they have a feature to show stock of product from sale.order.line
After https://github.com/odoo/odoo/commit/e813487e75f1 it was disabled too?

If not was disabled what about using the same way than Odoo to show it

Or better making the changes required in Odoo itself inheriting to show the extra data that we need

hugho-ad:
@JulioSerna 

> @hugho-ad @luisg123v
> 
> Why this feature it is added in this module?


Because the button on sale order lines form view can be used Only when clicking the lines and the pop form view being displayed on editable mode. without this, clicking the lines will nod open the form view of the lines.


Regards!

stock_by_warehouse_sale/views/sale_view.xml
moylop260: I'm not sure

Could you confirm that, please?

hugho-ad:
@moylop260 

> The commit referenced here:
> 
> * [odoo/odoo@e813487](https://github.com/odoo/odoo/commit/e813487e75f1)
> 
> has the following warning:
> 
> * The next version of the product configurator will only be compatible with inline tree view editing.
> 
> So, it could be a problem if we force a deprecated way for Odoo

Yes it could, until implemented or impproved this feature

> 
> I have a doubt here... they have a feature to show stock of product from sale.order.line
> After [odoo/odoo@e813487](https://github.com/odoo/odoo/commit/e813487e75f1) it was disabled too?
> 
> If not was disabled what about using the same way than Odoo to show it
> 
> Or better making the changes required in Odoo itself inheriting to show the extra data that we need

That is the correct way, I'm sharing with you the same point of view, in the meanwhile this change is made only because typ is migration to v14, and they what to keep this feature as it is, is that correct @desdelinux?

stock_by_warehouse_sale/views/sale_view.xml
luisg123v: I'm indeed sure. Records not marked as noupdate may be safely modified, including their external IDs.

As @hugho-ad mentioned, if they're renamed, old ones will be automatically removed and new ones created.

luisg123v:
> in the meanwhile this change is made only because typ is migration to v14, and they what to keep this feature as it is, is that correct @desdelinux?

Yes, because the feature can't be used on editing mode. AFAIK there's no native way to do it (that's why this module exists on the first place).

I Understand your concerns regarding the warning included inthe referenced commit's description, but, what other choice do we have? If module is installed, we need that in order for the module to be useful.

Of course, I'm open to suggestions.

moylop260:
Ok

I think there is another way to do it better since that Odoo has a similar feature but I'm no sure right now

For now,

👍

desdelinux:
There are 2 reasons, the first is related to the operation of the stock module, the second is that from the form view you can also view the optional variants (which you cannot from the list view).

As the user requires many fields to make decisions about the sale, the form is the simplest view to see it.

Right now users are making use of this view in 14, no problem, but the change was made at the form level.

In principle, we could include this fix temporarily (since the utility is critical for the user) and work on an improvement in the stock functionality so that it does not depend on that view.

cc @hugho-ad @moylop260 @hugho-ad @luisg123v @JulioSerna

PR #1508, opened 2021-10-20
[REF] default_warehouse_from_sale_team: Sequence assignation in v14.0 i#16547

luisg123v:
Automatic sequence numbering was not working in 14.0. For instance, now
purchase requisitions have different sequence codes, depending on
their types.

### DIFF
--- a/default_warehouse_from_sale_team/models/default_warehouse_mixing.py
+++ b/default_warehouse_from_sale_team/models/default_warehouse_mixing.py
@@ -5,6 +5,10 @@ class DefaultWarehouseMixing(models.AbstractModel):
     """If you inherit from this model and add a field called warehouse_id into
     the model, then the default value for such model will be the one
     set into the sales team.
+
+    Make sure to put this mixing at the top of inheritance (before the base model), e.g.
+
+        _inherit = ["default.warehouse.mixing", "sale.order"]
     """
     _name = "default.warehouse.mixing"
     _description = "Default Warehouse"
@@ -32,21 +36,19 @@ class DefaultWarehouseMixing(models.AbstractModel):
 
     @api.model
     def create(self, vals):
-        if vals.get('warehouse_id') and 'name' not in vals:
-            code = self._get_sequence_code()
-            pick_warehouse = self.env['stock.picking.type'].browse(vals.get('picking_type_id')).warehouse_id
-            warehouse_id = vals.get('warehouse_id', pick_warehouse.id)
-            section = self.env['crm.team'].search([('default_warehouse_id', '=', warehouse_id)], limit=1)
-            sequence = self.env['ir.sequence'].search(
-                [
-                    ('section_id', '=', section.id),
-                    ('code', '=', code),
-                ],
-                limit=1,
-            )
-            if sequence:
-                vals['name'] = sequence.next_by_id()
-        return super().create(vals)
+        """Pass sales team by context so it's taken into account when computing sequence name"""
+        salesteam = self._get_salesteam_from_vals(vals)
+        self_team = self.with_context(sequence_salesteam_id=salesteam.id)
+        return super(DefaultWarehouseMixing, self_team).create(vals)
 
-    def _get_sequence_code(self):
-        return self._name
+    def _get_salesteam_from_vals(self, vals):
+        """Determine sales team from creation values"""
+        if "name" in vals:
+            # Already has a name, so salesteam won't be used anyway
+            return self.env["crm.team"]
+        warehouse = (
+            self.env["stock.warehouse"].browse(vals.get("warehouse_id"))
+            or self.env["stock.picking.type"].browse(vals.get("picking_type_id")).warehouse_id
+        )
+        salesteam = warehouse.sale_team_ids[:1]
+        return salesteam
--- a/default_warehouse_from_sale_team/models/ir_sequence.py
+++ b/default_warehouse_from_sale_team/models/ir_sequence.py
@@ -1,7 +1,28 @@
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class IrSequence(models.Model):
     _inherit = "ir.sequence"
 
     section_id = fields.Many2one('crm.team', string='Sales Team')
+
+    @api.model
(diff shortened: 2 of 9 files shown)


luisg123v:
@moylop260 @hugho-ad @CarmenMiranda 

Could you review, please?

CC @desdelinux

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
```diff
@@ -5,6 +5,10 @@ class DefaultWarehouseMixing(models.AbstractModel):
     """If you inherit from this model and add a field called warehouse_id into
     the model, then the default value for such model will be the one
     set into the sales team.
+
+    Make sure to put this mixing at the top of inheritance (before the base model), e.g.
+
+        _inherit = ["default.warehouse.mixing", "sale.order"]
```
hugho-ad, line 11: why sale.order?

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
```diff
@@ -32,21 +36,19 @@ def default_get(self, fields_list):
...
+        """Pass sales team by context so it's taken into account when computing sequence name"""
+        salesteam = self._get_salesteam_from_vals(vals)
+        self_team = self.with_context(salesteam=salesteam)
+        return super(DefaultWarehouseMixing, self_team).create(vals)
 
-    def _get_sequence_code(self):
-        return self._name
+    def _get_salesteam_from_vals(self, vals):
+        """Determine sales team from creation values"""
+        if "name" in vals:
```
hugho-ad, line 46: I don't get it, why the field name?

default_warehouse_from_sale_team/models/stock_warehouse.py
```diff
@@ -0,0 +1,9 @@
+from odoo import models, fields
+
+
+class StockWarehouse(models.Model):
+    _inherit = "stock.warehouse"
+
+    sale_team_ids = fields.One2many(
```
hugho-ad, line 7: why multiple sales teams here?

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
luisg123v: This is just to demonstrate the mixing should go first than other inheritances

default_warehouse_from_sale_team/models/stock_warehouse.py
luisg123v: It's a field added by convenience, to reverse the already existing  m2o using a o2m. This saves us a search every time we need to do it.

BTW, this field is already in TyP.

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
luisg123v: Because if a name is already provided, it means the record already has one and a sequence number won't be required to be generated.

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
CarmenMiranda: Just a question, why should the mixing go first?

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
luisg123v: For more info, see https://github.com/Vauxoo/addons-vauxoo/pull/1463#

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
luisg123v: @CarmenMiranda Because if it's put at the end, the inheritance chain changes, which causes the native sale order's `create()` method to be called before this one; which in turn means when the `create()` in this module is called, the sale order would already have a  name with a generated sequence number.

default_warehouse_from_sale_team/models/ir_sequence.py
```diff
@@ -1,7 +1,28 @@
...
+        salesteam = self.env.context.get("salesteam")
+        if salesteam is None:
+            return super().next_by_code(sequence_code, sequence_date)
+        self.check_access_rights('read')
+        company_id = self.env.company.id
+        sequence = self.search(
+            [
+                ("code", "=", sequence_code),
+                ("company_id", "in", [company_id, False]),
+                ("section_id", "in", [salesteam.id, False]),
```
moylop260: salesteam is a value sent from context you should avoid using browseable in context better using integer directly

default_warehouse_from_sale_team/models/ir_sequence.py
```diff
@@ -1,7 +1,28 @@
...
 
 class IrSequence(models.Model):
     _inherit = "ir.sequence"
 
     section_id = fields.Many2one('crm.team', string='Sales Team')
+
+    @api.model
+    def next_by_code(self, sequence_code, sequence_date=None):
+        """If a sales team is provided by context, give priority to sequences having it set"""
+        salesteam = self.env.context.get("salesteam")
```
moylop260: This name of context looks like too generic

So, maybe another developer could use the same variable name

Could you define another non-generic name, please?

default_warehouse_from_sale_team/models/ir_sequence.py
luisg123v: Done, now it's an integer

default_warehouse_from_sale_team/models/ir_sequence.py
luisg123v: @moylop260 Done, now it's called `sequence_salesteam_id`, which means it's the salesteam ID to be used to compute sequence numbers. Not ambiguous anymore.

moylop260 (approved):
👍

default_warehouse_from_sale_team/models/default_warehouse_mixing.py
CarmenMiranda: Ohh, interesting, I didn't know that 🙈

CarmenMiranda (approved):
LGTM 👍🏼

PR #1509, opened 2021-10-23
[IMP] default_warehouse_from_sale_team: Rename field Default Warehouse i#16598

luisg123v:
Field name for default warehouse on sales teams was missing the proper
suffix on previous versions of this module.

This implements a migration script so the database column is renamed
when we're coming from previous versions.


luisg123v:
@moylop260 @hugho-ad 

Could you review, please?

CC @desdelinux

hugho-ad (approved):

PR #1510, opened 2021-11-23
[FIX] default_warehouse_from_Sale_team: Don't overwrite team nor warehouse i#16738

luisg123v:
When the partner is changed in a sale order, the current user's sales
team and warehouse should have priority over other ones.

Dummy MR: https://git.vauxoo.com/vauxoo/typ/-/merge_requests/885

### DIFF
--- a/default_warehouse_from_sale_team/models/crm_team.py
+++ b/default_warehouse_from_sale_team/models/crm_team.py
@@ -19,6 +19,16 @@ class CrmTeam(models.Model):
         'the warehouse of this sales team',
     )
 
+    def _get_default_team_id(self, user_id=None, domain=None):
+        """When specified by context, ensure the sales team is taken from the current user"""
+        if (
+            user_id is not None
+            and self.env.context.get("keep_current_user_salesteam")
+            and self.env.user.sale_team_id
+        ):
+            user_id = self.env.uid
+        return super()._get_default_team_id(user_id=user_id, domain=domain)
+
     def _get_default_journal_sale(self):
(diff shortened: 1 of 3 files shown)


luisg123v:
@moylop260 @CarmenMiranda 

Could you review, please?

CC @hugho-ad @desdelinux

moylop260:
@hugho-ad @CarmenMiranda 

I think this PR needs a functional review since that I'm not in context of this change

Check if it will not raise permissions errores removing sudo

CarmenMiranda:
> @moylop260 @CarmenMiranda
> 
> Could you review, please?
> 
> CC @hugho-ad @desdelinux

Technically LGTM 👍🏼 , but I agree that needs functional review 🙈

luisg123v:
Permission errors shouldn't occur because it's reading current user's sales team. All users [have permission to read users](https://github.com/odoo/odoo/blob/0d8d276ce76287f60d9e321312cd060e4c080a05/addons/sales_team/security/ir.model.access.csv#L2) and internal users [have permissions to read sales teams](https://github.com/odoo/odoo/blob/0d8d276ce76287f60d9e321312cd060e4c080a05/odoo/addons/base/security/ir.model.access.csv#L70)

If a permission error occurs, it would need to be checked separately, because you need access to read your own sales team to be able to use it.

desdelinux (approved):
Functionally LGTM

cc @luisg123v @hugho-ad

luisg123v:
Thanks you all.

PR #1516, opened 2021-12-28
[ADD] portal_multicurrency_totals: Show multicurrency totals on portal i#16797

luisg123v:
This module adds te feature to display total amounts by currency when
accessing documents through the portal:
- Quotations
- Sale orders
- Invoices

### Totals in sale orders:

totals_in_so

### Totals in invoices:

totals_in_invoices

Dummy MR: https://git.vauxoo.com/vauxoo/typ/-/merge_requests/905


luisg123v:
Hi @desdelinux, @imanie383 & @hugho-ad,

Could you review, please?

Regards,

imanie383:
You can create better styles for the totals( tables )
image

and especially invoice totals:
image

moylop260:
Are not there modules in the OCA with this feature?

luisg123v:
@imanie383,

- Regarding the invoice view, I agree with your comments
- Regarding the sale orders view, making it to not be displayed as a table was intentional, because it's like a summary with only two columns and there's no risk of confusion. I'll check with @desdelinux what way we should take here

@moylop260,

I think no, I couldn't find any module that satisfied this requirement, maybe because it's kind of specific.

moylop260:
👍

luisg123v:
@imanie383,

Done, now both cases are displayed as tables, and both uses the same style as Odoo does. This was accorded with @desdelinux 

@desdelinux,

As we discussed by chat, I'm now explicitly excluding draft and cancelled documents. When documents are validated and then cancelled, they still appear in the portal, because the customer is still a follower.

I also added a note in the README pointing that out.

Could you bot review again, please?

Regards,

desdelinux:
LGTM @luisg123v 

@imanie383 With the new style, I think it's enough

imanie383:
@luisg123v 

image

You are duplicating the pager

luisg123v:
@imanie383 you're right! I forgot to remove it.

Done, fixed.

CC @CarmenMiranda

imanie383:
LGTM 👌

imanie383:
@luisg123v 

LGTM 👍

PR #1520, opened 2022-01-25
[MIG] stock_by_warehouse_mrp_bom: Migration to 15.0 T#52780

agarzaarvizu:
### This PR refers to task[#52780](https://www.vauxoo.com/web#id=52780&action=1928&model=project.task&view_type=form&cids=1&menu_id=1490)
- Migration of the module stock_by_warehouse_mrp_bom to 15.0


agarzaarvizu:
@CarmenMiranda @luisg123v please review.

luisg123v (changes requested):
There's only one commit

agarzaarvizu:
@luisg123v Done.

stock_by_warehouse_mrp_bom/README.rst
```diff
@@ -0,0 +1,42 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+    :alt: License: LGPL-3
+
+Functional Rationale:
+---------------------
+
+Provide a nice widget in the product on the bill of material's line to see the quantity of product
+by warehouse in order to help you to review such information as fast as possible.
+
+.. image:: ./static/description/widget_by_warehouse.png
```
CarmenMiranda: The image in here should be the one in the BOM line and not the one in the product

stock_by_warehouse_mrp_bom/README.rst
```diff
@@ -0,0 +1,42 @@
...
+    :alt: Example on Form View
+    :width: 600px
+
+Also, provide another widget in the product on the bill of material's line to see the quantity in stock by locations
```
CarmenMiranda: I think this can be omitted @luisg123v What do you think?

This function is from the `stock_by_warehouse` module

CarmenMiranda (changes requested):

stock_by_warehouse_mrp_bom/README.rst
luisg123v: @CarmenMiranda x2

I though that was requested by you.

agarzaarvizu:
@luisg123v @CarmenMiranda done.

CarmenMiranda:
@agarzaarvizu In the image, it seems that there is no `warehouse_id` in the line, is that Ok?

Maybe if there is no warehouse there can be a change in the message so it doesn't say "The stock in: that is immediately..."
@luisg123v What do you think?

agarzaarvizu:
@CarmenMiranda @luisg123v  Done. I had to added a view cause the product in the BoM line was editable, so I inherited the view to change that attribute. Now is working correctly.

CarmenMiranda:
> @CarmenMiranda @luisg123v Done. I had to added a view cause the product in the BoM line was editable, so I inherited the view to change that attribute. Now is working correctly.

Ohh ok, it was only editable with the tree view, is Ok for me but I don't know if that is Ok functionally @luisg123v what do you think?

Just update the translations, remember that we change the view, and in that change, the span was added so the term changed

stock_by_warehouse_mrp_bom/views/mrp_bom_inherit.xml
```diff
@@ -0,0 +1,12 @@
+<odoo>
+    <record id="mrp_bom_view_form_inherit" model="ir.ui.view">
+        <field name="name">mrp.bom.stock.warehouse</field>
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='bom_line_ids']/tree" position="attributes">
+                <attribute name="editable"/>
```
luisg123v: sorry, but I don't agree with this change. We should try to avoid UX changes unless necessary.

BTW, even if you needed this view:
- File name should be `mrp_bom_views.xml`
- External ID should be the same as the original one

For more info, you could check [the guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#15file-naming)

stock_by_warehouse_mrp_bom/views/mrp_bom_view.xml
```diff
@@ -0,0 +1,28 @@
...
+<odoo>
+
+    <record id="mrp_bom_line_view_form_inherit" model="ir.ui.view">
+        <field name="name">mrp.bom.line.stock.warehouse</field>
+        <field name="model">mrp.bom.line</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_line_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//form/sheet/group" position="before">
+                <div class="text-center alert alert-info" role="alert">
+                    <p>The stock <span attrs="{'invisible':[('warehouse_id','=',False)]}">in: <field name="warehouse_id"/></span> that is immediately available is:
```
luisg123v: How could be the warehouse empty?

It's a required field and it's filled by default.

stock_by_warehouse_mrp_bom/views/mrp_bom_view.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="mrp_bom_line_view_form_inherit" model="ir.ui.view">
+        <field name="name">mrp.bom.line.stock.warehouse</field>
+        <field name="model">mrp.bom.line</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_line_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//form/sheet/group" position="before">
+                <div class="text-center alert alert-info" role="alert">
```
luisg123v: Thanks for adding the role attribute.

luisg123v (changes requested):

stock_by_warehouse_mrp_bom/views/mrp_bom_view.xml
CarmenMiranda: The field `warehouse_id` is related to the one in the [`picking_type_id` of the bom](https://github.com/odoo/odoo/blob/a33e04df6/addons/mrp/models/mrp_bom.py#L57) and that field is not required, even in the [view](https://github.com/odoo/odoo/blob/38e5d002ac/addons/mrp/views/mrp_bom_views.xml#L138) it says that can be invisible depending on the `type`, so if there is no other field in the BOM where the warehouse can be obtained it means that the warehouse can be empty

luisg123v: Thanks a lot @CarmenMiranda, I completely forgot we were talking about BoMs,m, not sale orders. Resolving.

stock_by_warehouse_mrp_bom/views/mrp_bom_inherit.xml
CarmenMiranda: In the [original view](https://github.com/odoo/odoo/blob/38e5d002/addons/mrp/views/mrp_bom_views.xml#L82) the field is only editable in the tree view, the way this widget is presented right now needs the form view to show it

If this change is not done a refactor should be done to show the widget in the line, something like the one I did in ABSA (asynchronous)

luisg123v: @CarmenMiranda but if the tree view is set to be inline editable, shouldn't the form view be displayed anyway when not editing the record?

CarmenMiranda: True, I don't remember right now if it can be triggered in edit mode, @agarzaarvizu Could you review if the button can be triggered please

stock_by_warehouse_mrp_bom/views/mrp_bom_view.xml
CarmenMiranda: Hahaha, yes is a BoM, personally, I haven't worked with this module, so I really don't know much about it hahaha, I just used the code to understand it, and I didn't found another way to get the warehouse 🙈

stock_by_warehouse_mrp_bom/views/mrp_bom_inherit.xml
agarzaarvizu: @CarmenMiranda If I'm in the BoM and click the line, the BoM changes to the edit mode automatically and the only button in the line is the one that shows you the product_product view to editing the view. I'm not sure if that's what you meant.

agarzaarvizu:
@CarmenMiranda @luisg123v Name of the view changed and new translation file added.

stock_by_warehouse_mrp_bom/views/mrp_bom_inherit.xml
luisg123v: OK, then I agree with this change. Besides, it's the same [done for sales in v14](https://github.com/Vauxoo/addons-vauxoo/blob/08280b001af01ca202d944369ace2f20e8e70efd/stock_by_warehouse_sale/views/sale_view.xml#L26)

luisg123v (approved):

PR #1523, opened 2022-02-19
[FIX] stock_by_warehouse: add some sudos lost during migration to 13.0 i#17886

luisg123v:
When migration of this module was performed from 12.0 to 13.0 (#1460),
some sudo's were removed because we considered them no longer required.
However, they actually are, in case there are record rules defined for
the warehouse model. Even if the current user doesn't have enough access
to see some warehouses, they need to know if there's available stock
elsewhere.

### Dummy MR:

- [vauxoo/typ!936](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/936)

### DIFF
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -73,8 +73,8 @@ class ProductProduct(models.Model):
         # itself, we enable this context management
         warehouse_id = self._context.get('warehouse_id')
 
-        for warehouse in self.env['stock.warehouse'].search([]):
-            product = self_origin.with_company(warehouse.company_id).with_context(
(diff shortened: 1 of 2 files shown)


luisg123v:
@desdelinux, @keylor2906, @hugho-ad,

Could you review, please?

CC @deivislaya

desdelinux:
LGTM :)

There is a visual bug that we must attack in another mr.

When there are many warehouses: there is no scroll that allows you to scroll down to see them all, so you only see a certain amount.

cc @keylor2906 @deivislaya @hugho-ad @luisg123v

hugho-ad (approved):

PR #1524, opened 2022-02-20
[FIX] stock_by_warehouse: add some sudos lost during migration to 13.0

luisg123v:
When migration of this module was performed from 12.0 to 13.0 (#1460),
some sudo's were removed because we considered them no longer required.
However, they actually are, in case there are record rules defined for
the warehouse model. Even if the current user doesn't have enough access
to see some warehouses, they need to know if there's available stock
elsewhere.

Forward-port of #1523


luisg123v:
This is just a forward-port by cherry-picking, so merging.

PR #1525, opened 2022-02-22
[FIX] default_warehouse_from_sale_team: access error when chained rule i#17897

luisg123v:
When a stock movement triggers an inventory rule that involves another
warehouse on which the current user has no access, an access error is
triggered. That's because, thanks to this module, users can only see
warehouses allowed in their sales teams.

To solve the above, access is granted to all warehouses when computing
inventory rules to apply.

### Dummy MR:

- [vauxoo/typ!936](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/936)

### DIFF
--- a/default_warehouse_from_sale_team/models/procurement_group.py
+++ b/default_warehouse_from_sale_team/models/procurement_group.py
@@ -1,0 +1,12 @@
+from odoo import api, models
+
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def _search_rule(self, route_ids, product_id, warehouse_id, domain):
+        """Grant access to provided warehouse if it's not allowed to the current user"""
+        if warehouse_id._access_unallowed_current_user_salesteams():
+            warehouse_id = warehouse_id.sudo()
+        return super()._search_rule(route_ids, product_id, warehouse_id, domain)
--- a/default_warehouse_from_sale_team/models/__init__.py
+++ b/default_warehouse_from_sale_team/models/__init__.py
@@ -4,6 +4,7 @@ from . import crm_team
 from . import default_warehouse_mixin
 from . import default_picking_type_mixin
 from . import ir_sequence
+from . import procurement_group
 from . import purchase_order
 from . import purchase_requisition
 from . import res_users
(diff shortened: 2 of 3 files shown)


luisg123v:
@desdelinux and (@keylor2906 or @hugho-ad),

Could you review, please?

FYI reusing a dummy again, but this change is applied there.

Regards,

hugho-ad:
why in v11 this error was not happen?

luisg123v:
@hugho-ad,

Because in v11 there was [this](https://github.com/Vauxoo/addons-vauxoo/blob/2719ebed2246896c17228cebc16fce9808459f46/default_warehouse_from_sale_team/models/sales_team.py#L88).

desdelinux (approved):
LGTM

It also solves the issue of manual transfers between warehouses

@hugho-ad @luisg123v please merged

default_warehouse_from_sale_team/models/procurement_group.py
```diff
@@ -0,0 +1,23 @@
...
+        flow that shouldn't be restricted by permissions.
+        """
+        rule_warehouse = self.env.ref("default_warehouse_from_sale_team.rule_default_warehouse_wh")
+        if (
+            warehouse_id
+            and not self.env.su
+            and warehouse_id.check_access_rights("read", raise_exception=False)
+            and self.env["ir.rule"]._get_failing(warehouse_id, mode="read") == rule_warehouse
+        ):
+            warehouse_id = warehouse_id.sudo()
```
hugho-ad, line 11: why not by default always use sudo as the old approach?

desdelinux: @luisg123v @hugho-ad I think that using sudo for everything we could also solve this case: https://www.vauxoo.com/web#id=17904&action=1918&model=helpdesk.ticket&view_type=form&cids=1&menu_id=1431

luisg123v: @hugho-ad I would like to avoid that approach, as we would be forcin sudo in all situations, independently of whether is required.

FYI the other issue commented by @desdelinux is not solved by adding sudo, because it's an access error on a different model (picking).

hugho-ad (approved):

PR #1528, opened 2022-02-24
[IMP] default_warehouse_from_sale_team: Default journal for vendor bills i#17892

luisg123v:
Currently, when creating a customer invoice, a default journal is set
if it's configured in the sales team. However, that feature only works
for customer invoices but not for vendor bills.

This change implements the default journal feature also for vendor
bills.

In addition, now company and currency are also considered when
computing default journal, for more accurate results.

### Dummy MR:

- [vauxoo/typ!950](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/950)

### DIFF
--- a/default_warehouse_from_sale_team/models/account_move.py
+++ b/default_warehouse_from_sale_team/models/account_move.py
@@ -7,20 +7,18 @@ class AccountMove(models.Model):
     @api.model
     def _search_default_journal(self, journal_types):
         """If a team is provided and it has a sales journal set, take it as 1st alternative"""
-        team = self.env.context.get("salesteam")
-        journal_on_team = None
-        if team and journal_types == ["sale"]:
-            journal_on_team = team._get_default_journal_sale()
+        team = self.env.context.get("salesteam") or self.team_id
+        journal_on_team = team._get_default_journal(journal_types)
         return journal_on_team or super()._search_default_journal(journal_types)
 
     @api.onchange('team_id')
     def _onchange_team_id(self):
-        if (
-            self.state != "draft"
-            or not self.team_id
-            or self.move_type not in self.get_sale_types(include_receipts=True)
-        ):
+        if self.state != "draft" or not self.is_invoice(include_receipts=True) or not self.team_id.journal_team_ids:
             return {}
-        default_journal = self.team_id._get_default_journal_sale()
-        if default_journal:
-            self.journal_id = default_journal
+        self = self.with_company(self.company_id)
+        default_journal_ctx = {
+            "default_move_type": self.move_type,
+            "default_currency_id": self.currency_id.id,
+        }
+        default_journal = self.with_context(**default_journal_ctx)._get_default_journal()
+        self.journal_id = default_journal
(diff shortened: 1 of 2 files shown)


luisg123v:
Hi @desdelinux & @keylor2906,

Could you review, please?

CC @hugho-ad

default_warehouse_from_sale_team/models/account_move.py
```diff
@@ -7,20 +7,18 @@ class AccountMove(models.Model):
     @api.model
     def _search_default_journal(self, journal_types):
         """If a team is provided and it has a sales journal set, take it as 1st alternative"""
-        team = self.env.context.get("salesteam")
-        journal_on_team = None
-        if team and journal_types == ["sale"]:
-            journal_on_team = team._get_default_journal_sale()
+        team = self.env.context.get("salesteam", self.team_id)
```
hugho-ad:
```suggestion
        team = self.env.context.get("salesteam") or self.team_id
```

default_warehouse_from_sale_team/models/account_move.py
```diff
@@ -7,20 +7,18 @@ class AccountMove(models.Model):
...
+        journal_on_team = team._get_default_journal(journal_types)
```
hugho-ad, line 11: you could use 

is_sale_document/ is_purchase_document

what happen with moves created manually 

and credit notes?

luisg123v: - Credit notes should also be considered
- In case of journal entries, it won't be triggered by the onchange because the onchange validates it's an invoice `is_invoice` which is `is_sale_document` + `is_purchase_document`, so the only time this could be called is when created, when `self.team_id` is not set. I could validate it, though.

default_warehouse_from_sale_team/models/account_move.py
luisg123v: Done.

default_warehouse_from_sale_team/models/account_move.py
```diff
@@ -7,20 +7,18 @@ class AccountMove(models.Model):
...
-        default_journal = self.team_id._get_default_journal_sale()
-        if default_journal:
-            self.journal_id = default_journal
+        self = self.with_company(self.company_id)
+        default_journal_ctx = {
+            "default_move_type": self.move_type,
+            "default_currency_id": self.currency_id.id,
+        }
+        default_journal = self.with_context(**default_journal_ctx)._get_default_journal()
+        self.journal_id = default_journal
```
CarmenMiranda, line 24: Hi, @luisg123v In here the `if default_journal ` is not needed anymore?

luisg123v: @CarmenMiranda no, because I'm calling the native method to retrieve the default journal, which tries to find a journal based on the salesteam as first choice, but fallbacks on Odoo's native behavior. So, we know in advance there will be a journal, unless there's no journal at all.

CarmenMiranda: Ohh, ok, I get it

CarmenMiranda (approved):
Technically LGTM 👍🏼

default_warehouse_from_sale_team/models/account_move.py
luisg123v: FYI I double-checked this doesn't affect journal entries and I can confirm it does not, so resolving.

desdelinux (approved):
LGTM ready for merged @luisg123v @hugho-ad

PR #1530, opened 2022-03-03
[FIX] default_warehouse_from_sale_team: typo in README t#53868

luisg123v:
Which was causing list of contributors not being correctly rendered.


rolandojduartem (approved):
LGTM!

PR #1531, opened 2022-03-09
[FIX] default_warehouse_from_sale_team: Sale Team Constraint Message

CarmenMiranda:
### [T#53868](https://www.vauxoo.com/web#id=53868&view_type=form&model=project.task)

Improving the message of the `sale_team_id` constraint, also, changing the `string` and `help` of the `sale_team_ids` field because is not readonly and the allowed teams should be set that field.

### DIFF
--- a/default_warehouse_from_sale_team/i18n/es.po
+++ b/default_warehouse_from_sale_team/i18n/es.po
@@ -52,6 +52,11 @@ msgstr ""
 " equipos de ventas"
 
 #. module: default_warehouse_from_sale_team
+#: model:ir.model.fields,field_description:default_warehouse_from_sale_team.field_res_users__sale_team_ids
+msgid "Allowed sales teams"
+msgstr "Equipos de ventas permitidos"
+
+#. module: default_warehouse_from_sale_team
 #: model:ir.model,name:default_warehouse_from_sale_team.model_default_picking_type_mixin
 msgid "Default Operation Type"
 msgstr "Tipo de Operación por defecto"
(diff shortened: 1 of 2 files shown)


default_warehouse_from_sale_team/i18n/es.po
```diff
@@ -51,6 +51,11 @@ msgstr ""
 "Permite al usuario poder ver todos los tippos de operación, sin importar sus"
 " equipos de ventas"
```
luisg123v, line 52: Shouldn't be "Equipos de venta"?

default_warehouse_from_sale_team/i18n/es.po
```diff
@@ -219,23 +230,16 @@ msgid "Stock Move"
...
-#: model:ir.model.fields,help:default_warehouse_from_sale_team.field_res_users__sale_team_ids
-msgid ""
-"This is only an informative field. Go to Sales > Sales > Sales Teams to edit"
-msgstr ""
-"Esto es sólo un campo informativo. Vaya a Ventas> Ventas> Equipos de ventas "
-"para editar"
+#: code:addons/default_warehouse_from_sale_team/models/res_users.py:0
+#, python-format
+msgid "The chosen team (%s) is not in the allowed sales teams for this user"
+msgstr "El equipo elegido (%s) no está entre los equipos de ventas permitidos para este usuario"
```
luisg123v:
```suggestion
msgstr "El equipo de venta seleccionado (%s) no está entre los equipos permitidos para este usuario"
```

Based on [this](https://github.com/odoo/odoo/blob/210913ad8475d543bf04a3fce9a9616ab879cce1/odoo/addons/base/i18n/es.po#L28899)

luisg123v (changes requested):

default_warehouse_from_sale_team/i18n/es.po
CarmenMiranda: This change is not mine 🙈

default_warehouse_from_sale_team/i18n/es.po
CarmenMiranda: Change done 😄

default_warehouse_from_sale_team/i18n/es.po
luisg123v: Sorry, I meant your change, you also used "equipos de ventas". But I just notices this is also the native translation, so resolving.

luisg123v (approved):

PR #1535, opened 2022-03-11
[FIX] default_warehouse_from_sale_team: access error on journal name i#17898

luisg123v:
Avoid access errors when user has access to a journal item but not to its
journal, e.g. when reconciling bank statements.

### Dummy MR:
 - [vauxoo/typ!950](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/950)


luisg123v:
@CarmenMiranda and @desdelinux,

Could you review, please?

CarmenMiranda (approved):
Technically LGTM 👍🏼

PR #1536, opened 2022-03-11
[FIX] stock_by_warehouse: Fix unittest error caused by action_confirm

rolandojduartem:
The issue was that the action_confirm takes 2 units of product immediately before action_assign.
On v14.0 the products were taken from picking after action_assign. Using action_confirm and
action_assign consecutively mantains the original behaviour from the test. This explained behaviour
is due to stock.move changes where this model handles action_assign when action_confirm is
executed. Functionality is preserved, but the change in the unittest is necessary to follow new
behaviour.

**References:**
https://github.com/odoo/odoo/commit/a838f2af683e03cd03ba02532f002ccd5c2642f5,
https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L1169
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1239

**Before:**

stock_by_warehouse_fix_unitest_error

**After:**

stock_by_warehouse_fix_unitest_solution

### DIFF
--- a/stock_by_warehouse/tests/test_stock_available_unreserved.py
+++ b/stock_by_warehouse/tests/test_stock_available_unreserved.py
@@ -9,14 +9,15 @@ class TestStockLogisticsWarehouse(TransactionCase):
         self.product_obj = self.env['product.product'].with_context(tracking_disable=True)
         self.template_obj = self.env['product.template']
         self.supplier_location = self.env.ref('stock.stock_location_suppliers')
-        self.stock_location = self.env.ref(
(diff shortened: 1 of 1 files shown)


rolandojduartem:
Hi, @luisg123v, could you review this fix, please?

rolandojduartem:
Hi, @luisg123v, I added some references and a brief explanation about the change in the unittest, could you review it, please?

luisg123v (approved):

PR #1537, opened 2022-03-17
[IMP] default_warehouse_from_sale_team: make ineritable bypassed rules i#17955

luisg123v:
We allow to bypass record rules related to sales teams when confirming
a  sale order, so deliveries may be created even if the involved routes
belong to other warehouses the current user has no access to.

This commit allows to extend those rules.

### Dummy MR:

- [vauxoo/typ!970](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/970)

### DIFF
--- a/default_warehouse_from_sale_team/models/stock_warehouse.py
+++ b/default_warehouse_from_sale_team/models/stock_warehouse.py
@@ -1,4 +1,4 @@
-from odoo import models, fields
+from odoo import api, models, fields
 
 
 class StockWarehouse(models.Model):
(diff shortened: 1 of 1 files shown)


luisg123v:
@CarmenMiranda or @keylor2906 or @hugho-ad,

Could you review, please?

CC @desdelinux @rolandojduartem

luisg123v:
@keylor2906 thanks for merging.

I'd recommend using the fast-forward method when merging, to avoid an extra merge commit to be included in the branch history.

PR #1539, opened 2022-05-11
[IMP] account_chatter: add tracking to allowed_journal_ids field in account.account model

edy1192:
- Add tracking to `allowed_journal_ids` field in `account.account` model.

- Avoid traceback adding `tag_ids` or `tax_ids` fields.

```
  File "/home/odoo/instance/odoo/odoo/http.py", line 639, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/odoo/instance/odoo/odoo/http.py", line 315, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: create_tracking_values() missing 1 required positional argument: 'model_name'
```

### DIFF
--- a/account_chatter/models/account_account.py
+++ b/account_chatter/models/account_account.py
@@ -18,16 +18,17 @@ class AccountAccount(models.Model):
     tax_ids = fields.Many2many(tracking=True)
     tag_ids = fields.Many2many(tracking=True)
     group_id = fields.Many2one(tracking=True)
+    allowed_journal_ids = fields.Many2many(tracking=True)
 
     def _message_track(self, tracked_fields, initial):
-        """Perform a field tracking over tax_ids and tag_ids
+        """Perform a field tracking over tax_ids, tag_ids and allowed_journal_ids
 
         This is performed manually because field tracking over many2many fields is not
         natively supported.
         """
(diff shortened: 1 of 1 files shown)


account_chatter/models/account_account.py
```diff
@@ -18,16 +18,17 @@ class AccountAccount(models.Model):
     tax_ids = fields.Many2many(tracking=True)
     tag_ids = fields.Many2many(tracking=True)
     group_id = fields.Many2one(tracking=True)
+    allowed_journal_ids = fields.Many2many(tracking=True)
 
     def _message_track(self, tracked_fields, initial):
-        """Perform a field tracking over tax_ids and tag_ids
+        """Perform a field tracking over tax_ids and tag_ids and allowed_journal_ids
```
luisg123v:
```suggestion
        """Perform a field tracking over tax_ids, tag_ids and allowed_journal_ids
```

luisg123v (changes requested):

account_chatter/models/account_account.py
edy1192: Done!

luisg123v (approved):

PR #1540, opened 2022-05-19
[IMP] account_chatter: add tracking to allowed_journal_ids field

edy1192:
- Add tracking to allowed_journal_ids field in account.account model.


luisg123v (changes requested):
Missing adding the new field to the list of manually-tracked fields, as well as docstring

edy1192:
Done @luisg123v

luisg123v (approved):

PR #1541, opened 2022-05-19
[REF] account_chatter: remove inherit and tracking in some fields

edy1192:
- This change is because in `15.0` the `account.account` model has tracking to some fields, so is not necessary the `mail.thread` inherit and overwrite the fields to add tracking.


edy1192:
Hi @luisg123v could you check this please?

luisg123v (approved):

PR #1543, opened 2022-05-19
[IMP] account_chatter: Add admin and root users as defaults on groups

ajqn9094:
- Add admin and root users as default on the group `group_show_account_chatter_notifications`.


ajqn9094:
@luisg123v Can you review please

luisg123v (approved):
FYI @randall-vx @edy1192

PR #1547, opened 2022-05-20
[MIG] account_chatter: finish up migration to v15 t#56103

luisg123v:
Even though an initial migration was already performed, module was
still not working.

Code was re-written to be cleaner and less intrusive to the tracking
mechanism.

### DIFF
--- a/account_chatter/models/mail_tracking_value.py
+++ b/account_chatter/models/mail_tracking_value.py
@@ -1,0 +1,31 @@
+from odoo import api, models
+
+
+class MailTrackingValue(models.Model):
+    _inherit = "mail.tracking.value"
+
+    @api.model
+    def create_tracking_values(self, initial_value, new_value, col_name, col_info, tracking_sequence, model_name):
+        """Perform a field tracking over many2many fields of model ``account.account``
+
+        This is performed manually because field tracking over many2many fields is not
+        natively supported.
+        """
+        values = super().create_tracking_values(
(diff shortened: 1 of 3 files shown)


moylop260:
👍

randall-vx:
LGTM @luisg123v

account_chatter/models/mail_tracking_value.py
```diff
@@ -0,0 +1,29 @@
...
+
+        This is performed manually because field tracking over many2many fields is not
+        natively supported.
+        """
+        values = super().create_tracking_values(
+            initial_value, new_value, col_name, col_info, tracking_sequence, model_name,
+        )
+        if values or model_name != "account.account" or col_info["type"] != "many2many":
+            return values
+        field = self.env["ir.model.fields"]._get(model_name, col_name)
```
CarmenMiranda, line 19: In the [original method](https://github.com/odoo/odoo/blob/14d4622af854c835c7553f3ad2c1f05957edcc6b/addons/mail/models/mail_tracking_value.py#L52-L53) is added a return early if the field doesn't exist, wouldn't be good adding it??

luisg123v: I don't consider it as necessary, as that should be (at least in my understanding) an impossible case, because if we reached that point, it means we're on a m2m field marked with `tracking=True`, so the field clearly exists.

I've applied your suggestion anyway, though.

CarmenMiranda (approved):
Technically LGTM 👍🏼

PR #1551, opened 2022-07-14
[ADD] search_action_domain: Add a filter option using server actions T#59176

rolandojduartem:
This module was moved to addons-vauxoo. This module lets the user to use server
actions as a filters in the "Favorites" section, in other words, the user
might use complex queries to search the records.

Now, "Create A Filter" is not in action section anymore, it is a button
in the server action's form view to create the filter in the selected
model.

**Server Action:**

search_action_domain_action

**Filter:**

search_action_domain_filter

**Result:**

search_action_domain_result


rolandojduartem:
Hi, @luisg123v, @nhomar, @tomeyro and @frahikLV, as we talked, this module was moved here. Please, may you review this PR?

cc @hugho-ad @desdelinux

tomeyro:
LGTM

search_action_domain/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Search Action Domain",
+    "summary": "Send a list of server actions via context to join the domains on the returned actions to a search.",
+    "author": "Vauxoo",
+    "website": "https://www.vauxoo.com",
+    "license": "LGPL-3",
+    "category": "Base/Base",
```
frahikLV: "Base" as main category isn't on  https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/data/ir_module_category_data.xml 
It could be "Technical" category, isn't?

 @luisg123v what do you think?

search_action_domain/tests/test_models.py
```diff
@@ -0,0 +1,49 @@
+import ast
```
frahikLV: Consider using
`from odoo.tools.safe_eval import safe_eval`

frahikLV (changes requested):
@rolandojduartem 

Here are my suggestions

search_action_domain/data/ir_actions_server_data.xml
```diff
@@ -0,0 +1,37 @@
...
+        'model_id': record.model_id.model,
+        'domain': '[]',
+        'context': "{'search_action_domain': {'%s': [%s]}}" % (record.model_id.model, record.id),
+        'active': True,
+        'user_id': False,
+        'is_default': False,
+        'action_id': False,
+        'sort': '[]',
+    })
+    env['ir.model.data'].sudo().create({
```
hugho-ad: could you check if it's posible to use _update_xmlids?

I thinks it's not possible because it's private

luisg123v: Yes it is possible.

search_action_domain/README.rst
```diff
@@ -0,0 +1,74 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+    :alt: License: LGPL-3
+
+Search Action Domain
+====================
+
+This module lets the user to use server actions as a filters in the "Favorites"
+section, in other words, the user might use complex queries to search the records.
```
luisg123v:
```suggestion
section. In other words, the user might use complex queries to search the records.  
```

or:

```suggestion
section, or in other words, the user might use complex queries to search the records.  
```

search_action_domain/README.rst
```diff
@@ -0,0 +1,74 @@
...
+This module lets the user to use server actions as a filters in the "Favorites"
+section, in other words, the user might use complex queries to search the records.  
+
+To use this kind of filter, go to:
+
+1. Settings > Technical > Actions > Server Actions > Create
+
+2. Select model and set the state as a Python code
+
+3. Write a valid Python code to search in the selected model
```
luisg123v: And the return value?

search_action_domain/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Server Action:
+
+    .. image:: ../search_action_domain/static/description/search_action_domain_action.png
+       :alt: search_action_domain: Create As Filter
+       :width: 400px
+
+Filter:
+
+    .. image:: ../search_action_domain/static/description/search_action_domain_filter.png
```
luisg123v, line 31: Are you sure about ".." in the path?

Check if that works functionally (opening module description).

search_action_domain/README.rst
```diff
@@ -0,0 +1,74 @@
...
+   :target: https://vauxoo.com
+
+This module is maintained by Vauxoo.
+
+A latinamerican company that provides training, coaching,
+development and implementation of enterprise management
+systems and bases its entire operation strategy in the use
+of Open Source Software and its main product is odoo.
+
+To contribute to this module, please visit http://www.vauxoo.com.
```
luisg123v: https

search_action_domain/README.rst
```diff
@@ -0,0 +1,74 @@
...
+addons-vauxoo/issues/new?body=module:%20
+search_action_domain%0Aversion:%20
+15.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Contributors
+------------
+
+* Tomás Álvarez <tomas@vauxoo.com>
+* Francisco Luna <fluna@vauxoo.com>
+* Andy Quijada <andy@vauxoo.com>
```
luisg123v: Not really a contributor in this case

search_action_domain/__manifest__.py
luisg123v: I'd use "Extra Tools", as most modules in server-ux

search_action_domain/data/ir_actions_server_data.xml
```diff
@@ -0,0 +1,37 @@
...
+<odoo>
+
+    <record id="create_as_ir_filter" model="ir.actions.server">
+        <field name="name">Create As Filter</field>
+        <field name="model_id" ref="base.model_ir_actions_server"/>
+        <field name="binding_model_id" ref="base.model_ir_actions_server"/>
+        <field name="binding_type">action</field>
+        <field name="state">code</field>
+        <field name="code">
+for record in records:
```
luisg123v: - Move to Python code + a button
- Use `_update_xmlid`
- Use the module name in the external ID, instead of `__export__` (making sure to use it noupdate). This will need a migration script

search_action_domain/tests/test_models.py
```diff
@@ -0,0 +1,49 @@
...
+
+from odoo.tests import Form, TransactionCase, tagged
+
+
+@tagged("models")
+class TestModels(TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.partner_obj = self.env["res.partner"]
```
luisg123v: Remove, it's better to don't have models this way to avoid them to get attached to a specific user and environment. If you use `self.env["res.partner"]` instead, it will take current env and user

search_action_domain/tests/test_models.py
```diff
@@ -0,0 +1,49 @@
...
+        search_action.model_id = self.partner_model
+        search_action.state = "code"
+        search_action.code = code
+        search_action = search_action.save()
+        return search_action
+
+    def test_01_search_action_domain(self):
+        # Create the server action
+        search_action = self.create_server_action(
+            code=(
```
luisg123v: Use multiline string, as a functional action would do

luisg123v (changes requested):

search_action_domain/README.rst
rolandojduartem: Yes, it is working, actually, I copied it from account_chatter module

search_action_domain/__manifest__.py
rolandojduartem: This was changed to "Extra Tools", as Luis suggested below.

search_action_domain/tests/test_models.py
rolandojduartem: Done!

search_action_domain/data/ir_actions_server_data.xml
rolandojduartem: Yes, it was possible, thanks for your recommendation.

search_action_domain/README.rst
rolandojduartem: Done!

search_action_domain/README.rst
rolandojduartem: Done!

search_action_domain/README.rst
rolandojduartem: Done!

search_action_domain/README.rst
rolandojduartem: Done!

search_action_domain/__manifest__.py
rolandojduartem: Done!

search_action_domain/data/ir_actions_server_data.xml
rolandojduartem: Done!

search_action_domain/tests/test_models.py
rolandojduartem: Done!

search_action_domain/tests/test_models.py
rolandojduartem: Done!

rolandojduartem:
Hi, @luisg123v, @hugho-ad and @frahikLV, your changes were applied. Please, may you review them?

Also, the "Create as Filter" option was moved from action section to the button in Server Actions

cc @nhomar @tomeyro @desdelinux

search_action_domain/data/ir_actions_server_data.xml
luisg123v: > Move to Python code + a button

Done, thanks.


> Use the module name in the external ID, instead of `__export__` (making sure to use it noupdate). This will need a migration script

- Here you used the model, it should be the module. Remember external IDs are in the form `module.record_name`. The reason to use this module as module is to know exactly what records were created by the feature and for the system to being able to remove them if the module gets uninstalled.
- Migration script is pending

> Use `_update_xmlid`

Sorry, I just noticed I wrote the wrong method, I meant `_load_records` ([here](https://git.vauxoo.com/vauxoo/sbdgroup/-/blob/db74d5e1527d57b48e94e694e83610b65867ec2e/sbd_importer/models/sbd_importer.py#L266) an example). That method:
- can create a record with its external ID
- Can receive several records at once
- You don't need to validate if it already exists. If it doesn't, it creates it. If it does, record is updated.

It's the same method internally called when loading data.

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,40 @@
+from odoo import models
+
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    def create_as_ir_filter(self):
+        model_data = self.env["ir.model.data"].sudo()
+        for record in self:
+            if not record.state == "code":
```
luisg123v: If you add it as a filtered in the for, you don't need this extra condition

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,40 @@
...
+                continue
+            xid = "%s.ir_filters_server_action_%s" % (
```
luisg123v:
```suggestion
            xid = "search_action_domain.ir_filters_server_action_%s" % (
```

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,40 @@
...
+                record.model_id.model.replace(".", "_"),
+                record.id,
+            )
+            if self.env.ref(xid, raise_if_not_found=False):
```
luisg123v: When using `_load_records` this won't be required

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,40 @@
...
+                continue
+            created_filter = self.env["ir.filters"].create(
+                {
+                    "name": record.name,
+                    "model_id": record.model_id.model,
```
luisg123v: Did you mean this?

```suggestion
                    "model_id": record.model_id.id,
```

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,40 @@
...
+                    "domain": "[]",
+                    "context": "{'search_action_domain': {'%s': [%s]}}"
+                    % (record.model_id.model, record.id),
+                    "active": True,
+                    "user_id": False,
+                    "is_default": False,
```
luisg123v: I think this is already the default. If so, remove this and other defaults, so if those are changed and button is re-pressed, those values are not lost

search_action_domain/tests/test_models.py
```diff
@@ -0,0 +1,48 @@
...
+            "res_partner.ir_filters_server_action_%s" % search_action.id
+        )
+        self.assertTrue(created_filter)
+        self.assertEqual(created_filter.name, search_action.name)
+        # Apply the filter succesfully
+        context = safe_eval(created_filter.context)
+        found_partner_action = (
+            self.env["res.partner"]
+            .with_context(
+                search_action_domain=context["search_action_domain"],
```
luisg123v: What is this for?

luisg123v (changes requested):

search_action_domain/README.rst
frahikLV: Hahaha poor andy

search_action_domain/tests/test_models.py
rolandojduartem: The idea of this module is applying the filter to search, adding this context, to append the domain from the action, why?

luisg123v: Maybe you should pass the whole context

```suggestion
                **context
```

BTW, `with_contex(**ctx)` should be in the same line I think

search_action_domain/models/models.py
```diff
@@ -0,0 +1,61 @@
...
+            return []
+        return self._get_cached_action_domain_to_include(
+            self._name,
+            ",".join(map(str, action_ids)),
+            id(http.request and http.request.httprequest),
+        )
+
+    @api.model
+    def _search(self, args, *oargs, **kwargs):
+        if isinstance(args, list):
```
luisg123v: This is adding an extra isinstance to literally every search on every model.

Same below in `read_group`

luisg123v (changes requested):
I'm a little bit concerned about inheriting too basic methods and not being careful enough about performance

search_action_domain/models/ir_actions_server.py
rolandojduartem: Hi, @luisg123v, probably this is an inconsistency, model_id is a selection field, so  record.model_id.model is the right value in this case. reference: https://github.com/odoo/odoo/blob/884a99357c1c44b2201e7116d7f6189626695f92/odoo/addons/base/models/ir_filters.py#L21

search_action_domain/models/ir_actions_server.py
rolandojduartem: no, is_default does not have a default value, but for this case, null and false are equivalent, but I set is_default to False. reference: https://github.com/odoo/odoo/blob/884a99357c1c44b2201e7116d7f6189626695f92/odoo/addons/base/models/ir_filters.py#L22

luisg123v: That's what I meant. If it doesn't have a default value, then `False` is the default values.

I think we should include only those values that differ from their defaults

rolandojduartem:
Hi, @luisg123v, I applied your suggestions. Please, may you review them?

search_action_domain/migrations/15.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+    convert_export_reference_to_search_action_domain_for_filter(cr)
+
+
+def convert_export_reference_to_search_action_domain_for_filter(cr):
+    cr.execute(
+        """
+        UPDATE
+            ir_model_data
+        SET
+            module = 'search_action_domain'
```
luisg123v: You need to also set them as noupdate, otherwise they will be deleted on next module update

search_action_domain/migrations/15.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+        WHERE
+            name like 'ir_filters_server_action_%'
+            AND module = '__export__'
+        """
+    )
+    _logger.info(
+        "%s records from ir_model_data were updated from __export__ to search_action_domain"
+        % cr.rowcount
```
luisg123v: Instead of:

`_logger.info(message % args)`

Use:

`_logger.info(message, args)`

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,25 @@
...
+                [
+                    {
+                        "xml_id": "search_action_domain.ir_filters_server_action_%s"
+                        % record.id,
+                        "noupdate": True,
+                        "values": {
+                            "name": record.name,
+                            "model_id": record.model_id.model,
+                            "context": "{'search_action_domain': {'%s': [%s]}}"
+                            % (record.model_id.model, record.id),
```
luisg123v: For readability, I would recommend:

```
"key": "val" % (
    arg1, arg2,
),
```

search_action_domain/models/ir_actions_server.py
```diff
@@ -0,0 +1,25 @@
+from odoo import models
+
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    def create_as_ir_filter(self):
+        for record in self.filtered(lambda r: r.state == "code"):
+            self.env["ir.filters"]._load_records(
```
luisg123v: This can receive a list, you can pass all at once.

Maybe a list comprehension

```
self.env["ir.filters"]._load_records(
    [
        ...
    ]
    for record in self.filtered(lambda r: r.state == "code")
)
```

search_action_domain/tests/test_models.py
```diff
@@ -0,0 +1,42 @@
...
+action = {'domain': [('id', 'in', elements.ids)]}
+            """
+        )
+        # Execute "Create As Filter" method from the created server action
+        search_action.create_as_ir_filter()
+        # Check if the filter was created successfully
+        created_filter = self.env.ref(
+            "search_action_domain.ir_filters_server_action_%s" % search_action.id
+        )
+        self.assertTrue(created_filter)
```
luisg123v, line 35: use `raise_if_not_found=False` in previous line, otherwise this one makes no sense, as it won't be reached even if filter is not found

luisg123v (changes requested):
- Migration commit should be MIG
- You're mentioning this was moved to addons-vauxoo but, where is being moved from? I'd left that info only in PR description, because it's not as related for history where was it stored (IMHO)

search_action_domain/models/ir_actions_server.py
luisg123v: Got it, resolving.

search_action_domain/models/models.py
```diff
@@ -0,0 +1,59 @@
...
+            res = action.with_context(active_model=model).run()
+            res_domain = res and isinstance(res, dict) and res.get("domain")
+            domain += res_domain if res_domain and isinstance(res_domain, list) else []
+        return domain
+
+    @api.model
+    def _get_action_domain_to_include(self):
+        ctx_dict = self.env.context.get("search_action_domain") or {}
+        if not isinstance(ctx_dict, dict):
+            return []
```
luisg123v: Remove condition, efficiency

search_action_domain/models/models.py
```diff
@@ -0,0 +1,59 @@
...
+        action_ids = ctx_dict.get(self._name) or []
+        if (
+            not action_ids
+            or not isinstance(action_ids, list)
```
luisg123v: Remove. There shouldn't be wrong types. And if they are, I consider it's better to fail than slowing down every correct operation.

Same for action IDs

luisg123v (changes requested):

moylop260:
LGTM

search_action_domain/models/ir_actions_server.py
rolandojduartem: Black is used here.

rolandojduartem:
@luisg123v last changes were applied, please, may you review them?

search_action_domain/migrations/15.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    convert_export_reference_to_search_action_domain_for_filter(cr)
+
+
+def convert_export_reference_to_search_action_domain_for_filter(cr):
+    cr.execute(
+        """
+        UPDATE
+            ir_model_data
+        SET
+            module = 'search_action_domain'
```
luisg123v: Missing comma, syntax error:

```suggestion
            module = 'search_action_domain',
```

luisg123v (changes requested):

search_action_domain/models/ir_actions_server.py
luisg123v: OK, resolving

search_action_domain/models/models.py
rolandojduartem: Done!

search_action_domain/models/models.py
rolandojduartem: Done!

search_action_domain/migrations/15.0.1.0.0/post-migration.py
rolandojduartem: Done!

rolandojduartem:
Hi, @luisg123v, Change is done, please, may you review it?

luisg123v (approved):

PR #1554, opened 2022-07-21
[FIX] search_action_domain: Make installable with base_automation T#59176

rolandojduartem:
At the moment to install search_action_domain with base_automation appears
 an error. Error indicates 'create_as_ir_filter is not a valid action on
base.automation'. This fix avoid the validation for the buttons if
base automation is in process.

### DIFF
--- a/search_action_domain/models/ir_ui_view.py
+++ b/search_action_domain/models/ir_ui_view.py
@@ -1,0 +1,18 @@
+from odoo import models
+
+
+class IrUiView(models.Model):
+    _inherit = "ir.ui.view"
+
+    def _validate_tag_button(self, node, name_manager, node_info):
+        """At the moment to install search_action_domain with base_automation, an error
+        appears to indicate the created button is not a valid action for base.automation.
+        This fix avoid the validation for the created button if base_automation
+        installation is in process.
+        """
+        if (
+            name_manager.model._name == "base.automation"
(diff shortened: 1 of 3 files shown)


rolandojduartem:
Hi, @luisg123v, could you check this fix, I tested it and this module installed correctly.

search_action_domain/models/ir_ui_view.py
```diff
@@ -0,0 +1,10 @@
+from odoo import models
+
+
+class IrUiView(models.Model):
+    _inherit = "ir.ui.view"
+
+    def _validate_tag_button(self, node, name_manager, node_info):
+        if name_manager.model._name == "base.automation":
```
luisg123v: Add more conditions to affect only the desired button.

Current.

This condition is OK to be the 1st one, because this will discard faster all validations of all other models

luisg123v (changes requested):

luisg123v (changes requested):
- Please also add a small docstring to explaing that's to don't fail when the base automation module is installed.
- We're hiding the error, but we need also to hide the button when we're on an automated action. That could be achieved by making invisible when `usage = base_automation`

rolandojduartem:
@luisg123v changes are ready

luisg123v (approved):

PR #1562, opened 2022-08-26
[ADD] ci: enable tests using GitHub actions t#54296

luisg123v:
The following tests are added:
- pre-commit-vauxoo
- No dependency files
- Odoo

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,72 @@
+name: tests
+
+on:
+  pull_request:
+    branches:
+      - "15.0*"
+  push:
+    branches:
+      - "15.0"
+
+env:
+  TRAVIS_BUILD_DIR: ${{ github.workspace }}
+  LINT_CHECK: "0"
+  TESTS: "0"
+
+jobs:
+  pre-commit-vauxoo:
+    runs-on: ubuntu-latest
+    name: pre-commit-vauxoo
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache pre-commit and pip packages
+        id: cache-pre-commit-pip
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pre-commit
+            ~/.cache/pip
+          key: cache-pre-commit-pip
+      - name: Install pre-commit-vauxoo
+        run: pip install --upgrade pre-commit-vauxoo
+      - name: Run pre-commit-vauxoo
+        run: pre-commit-vauxoo
+  no-dependency-files:
+    runs-on: ubuntu-latest
+    name: No dependency files
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


luisg123v:
@moylop260 

Could you review, please?

I know it's red, but that's why I wanted to enable it. Next step is to fix those reds.

Regards,

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,33 @@
...
+    name: ${{ matrix.name }}
+    env:
+      PRECOMMIT_HOOKS_TYPE: ${{ matrix.name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [mandatory, optional]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Upgrade pre-commit-vauxoo to te latest version
```
moylop260:
```suggestion
      - name: Upgrade pre-commit-vauxoo to the latest version
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: pip3 install --upgrade pre-commit-vauxoo
+      - name: Run pre-commit-vauxoo
+        run: pre-commit-vauxoo
```
moylop260: How are you set optional or mandatory to the command?

moylop260 (commented):
The OPTIONAL pipeline is showing green status

Shouldn't be yellow one?

.github/workflows/pre-commit.yml
luisg123v: Setting variable from job name:

```
env:
  PRECOMMIT_HOOKS_TYPE: ${{ matrix.name }}
```

moylop260:
Could you merge the build pre-commit-optional and pre-commit-mandatory in order to review all the logs only in one build even if it is green, please?

Since github doesn't have yellow status or allow_failure as gitlab so the optional will be green all the time but we have 2 pre-commit builds so the developer will need to check the log of optional but they could get confused about what pre-commit build check

IMHO it is better if you have the full logs only in one build for this case in order to check only one output instead of 2

luisg123v:
@moylop260 I agree, I wasn't aware GitHub doesn't support yellow when I implemented this

moylop260:
excellent!

Also, you even don't need the container to run the tests

I think it could be even faster running over the original GH ubuntu but with cache

I will make a PR to your PR...

moylop260:
Please, check the following changes:
 - https://github.com/Vauxoo/addons-vauxoo/pull/1564

It is running the pre-commit build spending 23s

moylop260:
Also, could you add a error if the dev adds `oca_dependencies.txt` file in the CI, please?

moylop260:
👍 
(After removing the configuration files)

PR #1567, opened 2022-08-29
[FIX] *: Fix lints

edy1192:
**account_analytic_business_unit module:**
- C7902(missing-readme).
- C8103(manifest-deprecated-key).
- C8106(manifest-version-format).
- F401 'odoo._' imported but unused.
- E303 too many blank lines.
- E251 unexpected spaces around keyword / parameter equals.

**account_chatter module:**
- W7902(duplicate-xml-record-id).

### DIFF
--- a/account_chatter/views/account_account_views.xml
+++ b/account_chatter/views/account_account_views.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
 
-    <record id="view_account_form" model="ir.ui.view">
+    <record id="view_account_form_show_chatter" model="ir.ui.view">
         <field name="name">account.account.form.inherit.account.chatter</field>
         <field name="model">account.account</field>
         <field name="inherit_id" ref="account.view_account_form"/>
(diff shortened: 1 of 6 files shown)


edy1192:
Hi @luisg123v could you check this please?

Do I need to do anything extra regarding the changed xml id in `account_chatter` and manifest version in `account_analytic_business_unit`?

account_chatter/views/account_account_views.xml
```diff
@@ -17,7 +17,7 @@
             </xpath>
         </field>
     </record>
-    <record id="view_account_form" model="ir.ui.view">
+    <record id="view_account_form_inherit" model="ir.ui.view">
```
luisg123v: Since this is the standard view inheritance (no groups), it could maintain the original ID, and the above could be named `view_account_form_show_chatter`

luisg123v (changes requested):

account_chatter/views/account_account_views.xml
edy1192: Done!

luisg123v (approved):

PR #1575, opened 2022-10-19
[MIG] account_multicompany_code: Migrate to v15

JesusValdez96:
- Fix black, isort and prettier lints
- Fix pylints in res_company model (use Python 3 style super() without arguments)


JesusValdez96:
@hugho-ad Please review this.

luisg123v (changes requested):
We accorded by mail when this module was migrated to v14 that we will depend on the module `res_company_code`, because its constraint is more efficient (it's an SQL constraint instead of a Python one) and the code is considered also on the company switcher. That module is [here](https://github.com/OCA/multi-company/tree/14.0/res_company_code)

hugho-ad:
@luisg123v 

why not to use only the oca one?

the mig to v14 was performed by you
https://github.com/Vauxoo/addons-vauxoo/pull/1458

hugho-ad:
ya lei los hilos completos sobre este modulo 
 creo que @nhomar no espera esta dependencia del modulo de la oca.

En mi opinión deberiamos manterner nuestro modulo sin depender  de la oca

luisg123v:
No, what @nhomar said was to don't replace this module with the OCA's one, because this module also adds other feature.

This does two things:
1. Add a code to the company and ensure it's unique
2. Use that code in accounts

The 1st point is covered by the OCA module on a more complete and efficient manner, so I think it should be used, and only the 2nd point should be covered from our side

hugho-ad (approved):

hugho-ad:
Strategic merge

We will NOT depend on OCA module, this will remain as it's!

account_multicompany_code/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "author": "Vauxoo",
+    "category": "Accounting/Accounting",
+    "website": "https://www.vauxoo.com/",
+    "license": "LGPL-3",
+    "depends": [
+        "account",
+    ],
+    "demo": [],
+    "data": [
+        "views/company_view.xml",
```
luisg123v, line 13: `res_company_views.xml`

FYI (for next migration) black-related changes goes into a separated commit, to be able to distinguish formatting changes from actual migration changes

account_multicompany_code/models/account.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def name_get(self):
+        result = dict(super().name_get())
+        for record in self:
+            if record.company_id.code:
+                result[record.id] += " (%s)" % record.company_id.code
+        return list(result.items())
+
+
+class AccountJournal(models.Model):
```
luisg123v, line 15: Move to its own file, each model should be placed on a separated file

account_multicompany_code/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo.exceptions import ValidationError
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    code = fields.Char(help="Internal code name of the company")
+
+    @api.constrains("code")
+    def unique_code(self):
```
luisg123v, line 11: Move to SQL constraint, for efficiency, as the module `res_company_code` does

account_multicompany_code/tests/test_multicompany_code.py
```diff
@@ -0,0 +1,32 @@
+from odoo.tests import TransactionCase
+
+
+class TestMulticompanyCode(TransactionCase):
+    def setUp(self):
```
luisg123v, line 5: `setUpClass`

account_multicompany_code/views/company_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="view_company_form" model="ir.ui.view">
+        <field name="name">qf_common.company.form</field>
+        <field name="model">res.company</field>
+        <field name="inherit_id" ref="base.view_company_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']" position="before">
+                <field name="code" required="1" />
```
luisg123v, line 10: Missing changes on other views, check module `res_company_code` for more context

luisg123v (commented):
Better late than never

luisg123v:
@LuisMzz could you take care of the review I left post migration, please?

luisg123v:
@hugho-ad prefer merging using the rebase method, to don't create a merge commit.

PR #1579, opened 2022-11-11
[FIX] account_multicompany_code: Missing Square Backet

CarmenMiranda:


CarmenMiranda:
@hugho-ad @luisg123v Sorry, my bad, I fixed something before the push and I miss the bracket 🙈

PR #1580, opened 2022-11-17
[FIX] default_warehouse_from_sale_team: xpath dependency on flaky element

carralc:
This commit fixes an xpath expression that relies on a `page` element that can later be removed by the enterprise module `website_crm_score`. When trying to add the previous module as a dependency for another module, it would prevent this module from loading with an "xpath not found" error.

### DIFF
--- a/default_warehouse_from_sale_team/views/crm_team_views.xml
+++ b/default_warehouse_from_sale_team/views/crm_team_views.xml
@@ -5,6 +5,7 @@
         <field name="name">crm.team.form.inherit.default.warehouse</field>
         <field name="model">crm.team</field>
         <field name="inherit_id" ref="sales_team.crm_team_view_form"/>
+        <field name="priority">14</field>
         <field name="arch" type="xml">
             <xpath expr="//page[@name='members']" position="after">
                 <page string="Warehouse settings for the sales team">


hugho-ad (approved):

luisg123v (changes requested):
title: REF ->  FIX, as you're not refactoring the module but fixing an issue

BTW, inside is not exactly the same as after, as inside will put the page at the end, while after would put it just after members. In this case is OK because there aren't other modules adding pages, but it's a good practice to support that.

The proper fix would be:
- Make a PR to enterprise to fix that replace, to make it invisible instead of replaced (ideal solution but time consuming)
- Decrease priority of this view below 14 (which is the default one), so this one is applied before the enterprise one

carralc:
> title: REF -> FIX, as you're not refactoring the module but fixing an issue
> 
> BTW, inside is not exactly the same as after, as inside will put the page at the end, while after would put it just after members. In this case is OK because there aren't other modules adding pages, but it's a good practice to support that.
> 
> The proper fix would be:
> 
> * Make a PR to enterprise to fix that replace, to make it invisible instead of replaced (ideal solution but time consuming)
> * Decrease priority of this view below 14 (which is the default one), so this one is applied before the enterprise one

I still think changing the xpath expression is the correct way to go. If you have a clean instance with `website_crm_score` installed and you try to install `default_warehouse_from_sale_team`, installation will fail with the same missing xpath error. This fix also makes the view independent on further changes in both odoo and enterprise codebases.

luisg123v:
> If you have a clean instance with `website_crm_score` installed and you try to install `default_warehouse_from_sale_team`, installation will fail with the same missing xpath error.

Are you sure? Since the priority is less than the  default one, this view should be applied before that one.

But even if it fails, it's not this module's fault, is the enterprise one that shouldn't replace elements, specially with a so low priority.


BTW, we even have a lint for that. Replacing elements is a bad practice, it's much more preferred to hide them.

You wouldn't even need another module for that to fail. If you for instance rename the view, the old and new one will be present for a few seconds (while Odoo figures out the old one was removed), and it will also fail because the element it's trying to replace doesn't exist anymore.

luisg123v (approved):
LGTM :+1:

PR #1586, opened 2023-01-10
[IMP] product_cost_usd: Use product.template in the price calculation T#53875

ajqn9094:
Modification to avoid errors coming from the store when listing the products, due to the fact that the the ID of the record used may belong to the product.template model, therefore change the iteration object to the variable `products_qty_partner` to more directly take the value of the product used in the process


ajqn9094:
@luisg123v Could you review please?

Thanks in advance.

luisg123v (approved):

PR #1605, opened 2023-03-07
[FIX] .github: error when running tests from a fork T#66175

luisg123v:
For security reasons, when tests are run from a fork, secrets are not availabe to the jobs [1], not even if the same secrets are configured in the fork. Unfortunately, We rely in such secrets in order for vxci to work properly, so it is able to connect to Quay.io, Orchest, etc.

To solve the avove, we are making tests to run directly in the target repo, but checking out the code submitted in the source repo. That way:
- Security is preserved because actions will always be taken from the base repo, so if for instance someone submits a PR changing action's code, it will never be run
- But code to test will be taken from source repo

[1] https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,12 +1,17 @@
 name: tests
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
-      - "15.0*"
+      - "15.0"
   push:
     branches:
       - "15.0"
+
+env:
+  CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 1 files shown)


luisg123v:
Hi @moylop260 or @ruiztulio,

Could you review, please?

Note that CI won't be run until this is merged, but it was already tested on #1603

CC @rolandojduartem @Hernannc @LuisMzz 

Regards,

moylop260:
👍

moylop260 (approved):

PR #1606, opened 2023-03-08
[FIX] .github: error when running tests from a fork

luisg123v:
For security reasons, when tests are run from a fork, secrets are not availabe to the jobs [1], not even if the same secrets are configured in the fork. Unfortunately, We rely in such secrets in order for vxci to work properly, so it is able to connect to Quay.io, Orchest, etc.

To solve the avove, we are making tests to run directly in the target repo, but checking out the code submitted in the source repo. That way:
- Security is preserved because actions will always be taken from the base repo, so if for instance someone submits a PR changing action's code, it will never be run
- But code to test will be taken from source repo

[1] https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,12 +1,17 @@
 name: tests
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
-      - "16.0*"
+      - "16.0"
   push:
     branches:
       - "16.0"
+
+env:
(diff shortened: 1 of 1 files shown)


luisg123v:
This is just a forward-port of #1606, so merging.

PR #1613, opened 2023-05-08
[FIX] product_cost_usd: Show standard_price_usd field as USD T#71674

rolandojduartem:
When the currency of the company is different to USD, the field is shown with the currency of the company instead of USD, the field is set to be shown as USD.


rolandojduartem:
Hi, @luisg123v, could you review this PR? this PR is related to this [issue](https://www.vauxoo.com/web#active_id=22248&cids=1&id=22248&model=helpdesk.ticket&menu_id=)

cc @desdelinux

luisg123v (approved):

PR #1617, opened 2023-06-01
[FIX] default_warehouse_from_sale_team: Avoid return the same record at create

R4Alex:
The self in the method create inheritance could be not empty, for example, using a recordset.copy(), in that case the self is not empty.

Using the self.browse(), we will ensure that the self is empty, and the split create calls are made as expected.


R4Alex:
Hi, could you review this PR? please, thank you @luisg123v @rolandojduartem

luisg123v (approved):

PR #1621, opened 2023-07-11
[FIX] account_chatter: Account move narration field is html field no text I#22876

hugho-ad:
This module reassigns the narration field type, from html to text, that
was wrong because when the field being refereced from a related field
following error appears

	TypeError: Type of related field bank.rec.widget.st_line_narration is
	inconsistent with account.move.narration

So we remove the field inheritance, because the only reason this field
was inherited was to make it trackeable, which is not supported for
HTML fields.

### DIFF
--- a/account_chatter/models/account_move.py
+++ b/account_chatter/models/account_move.py
@@ -6,5 +6,4 @@ class AccountMove(models.Model):
 
     date = fields.Date(tracking=True)
     journal_id = fields.Many2one(tracking=True)
-    narration = fields.Text(tracking=True)
     reversed_entry_id = fields.Many2one(tracking=True)


carralc:
LGTM +1

luisg123v (changes requested):
Code LGTM, just fix typos in commit description:
- "it's" -> is
- reasing -> reas**s**ign**s**
- "TypeError: ..." Indent and leave on a separate paragraph. When you do so, it's rendered as code
- Remove ticket ID from commit description

account_chatter/models/account_move.py
```diff
@@ -6,5 +6,5 @@ class AccountMove(models.Model):
 
     date = fields.Date(tracking=True)
     journal_id = fields.Many2one(tracking=True)
-    narration = fields.Text(tracking=True)
+    narration = fields.Html(tracking=True)
```
luisg123v: Fields tracking is [not supported](https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_tracking_value.py#L57) for HTML field, so just remove this inheritance.

luisg123v (changes requested):

account_chatter/models/account_move.py
hugho-ad: done

luisg123v (approved):
LGTM :+1:

PR #1625, opened 2023-09-13
[FIX] stock_manual_transfer: remove .unlink() method override

carralc:
Due to upstream updates adding support for @api.ondelete and pre-commit-vauxoo detecting the override of .unlink() method as a code smell, this transforms the old logic into more idiomatic odoo code.


hugho-ad (approved):

luisg123v (commented):
@carralc you forgot to remove the call to super

PR #1631, opened 2023-11-02
[IMP] internal_transfer_with_agreed_amount: modify write vals in one method i#24002

CLaurelB:
Modify in a single method the values that will be written in the account
move lines so that the method can be inherited in case it is necessary
to modify more values in the lines.

In addition, two methods are added to pre and post process payments, to
be used if modifications are required in an inheritance.

Finally, obtaining currencies related to the internal transfer have been
moved to a separate method to facilitate the inheritance of the method
that validates the number of currencies.

Changelog:
- Set debit and credit values to write in _prepare_values_from_lines
  method when the exchange currency is the same as the company currency.
- Add methods to be inherited when payments also needs modifications.
- Get the currencies in a new method named _get_currencies.

### DIFF
--- a/internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
+++ b/internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
@@ -21,14 +21,18 @@ class InternalTransferMulticurrency(models.TransientModel):
         payment = self.env["account.payment"].browse(active_ids)
         return payment.destination_journal_id.currency_id or payment.company_currency_id
 
-    def _validate_currencies_in_internal_transfer(self, payment):
+    def _get_currencies(self, payment):
         payment_currency = payment.currency_id
         company_currency = payment.company_currency_id
         send_currency = payment.journal_id.currency_id or company_currency
         receive_currency = self.currency_id
         currencies = payment_currency | company_currency | send_currency | receive_currency
+        return currencies
+
+    def _validate_currencies_in_internal_transfer(self, payment):
+        currencies = self._get_currencies(payment)
         if len(currencies) == 2:
-            return currencies - payment_currency
+            return currencies - payment.currency_id
         raise ValidationError(
             _(
                 "Agreed amount is available when there are two considered currencies. "
@@ -37,12 +41,22 @@ class InternalTransferMulticurrency(models.TransientModel):
             )
         )
 
-    def _prepare_values_from_lines(self, lines, currency):
+    def _prepare_values_from_lines(self, lines, currency=None, is_same_company_currency=False):
         line = lines[0]
+        if is_same_company_currency:
+            return {"debit": self.agreed_amount} if line.debit > 0 else {"credit": self.agreed_amount}
         vals = {"debit": line.debit} if line.amount_currency > 0 else {"credit": line.credit}
         sign = 1 if line.amount_currency > 0 else -1
         vals.update({"amount_currency": sign * self.agreed_amount, "currency_id": currency.id})
         return vals
+
+    def _preprocess_payments_hooks(self, payment):
+        """This method is to be inherited in case pre-processing is required on the payments."""
+        return payment
+
+    def _postprocess_payments_hooks(self, payment):
+        """This method is to be inherited in case post-processing is required on the payments."""
+        return payment
 
     def apply(self):
         active_ids = self._context.get("active_ids")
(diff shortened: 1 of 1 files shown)


CLaurelB:
@rolandojduartem could you review, please?

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
```diff
@@ -37,13 +37,19 @@ def _validate_currencies_in_internal_transfer(self, payment):
...
+    def _prepare_values_from_lines(self, lines, currency=None, is_same_company_currency=False):
         line = lines[0]
+        if is_same_company_currency:
+            return {"debit": self.agreed_amount} if line.debit > 0 else {"credit": self.agreed_amount}
         vals = {"debit": line.debit} if line.amount_currency > 0 else {"credit": line.credit}
         sign = 1 if line.amount_currency > 0 else -1
         vals.update({"amount_currency": sign * self.agreed_amount, "currency_id": currency.id})
         return vals
 
+    def _postprocessing_payment_with_aggred_amount(self, payment):
```
rolandojduartem:
```suggestion
    def _postprocessing_payment_with_agreed_amount(self, payment):
```

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
```diff
@@ -64,8 +70,10 @@ def apply(self):
...
             credit = amls.filtered(lambda x: x.credit > 0)
-            debit.with_context(check_move_validity=False).write({"debit": self.agreed_amount})
-            credit.with_context(check_move_validity=False).write({"credit": self.agreed_amount})
+            vals = self._prepare_values_from_lines(debit, is_same_company_currency=True)
+            debit.with_context(check_move_validity=False).write(vals)
+            vals = self._prepare_values_from_lines(credit, is_same_company_currency=True)
+            credit.with_context(check_move_validity=False).write(vals)
             amls.move_id.action_post()
             to_reconcile.reconcile()
             return
```
rolandojduartem, line 90: Sorry, you have to call the method before the return too  self._postprocessing_payment_with_agreed_amount(payment)

CLaurelB: Added.

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
CLaurelB: Fixed.

CLaurelB:
@rolandojduartem could you review again, please?

rolandojduartem (approved):
LGTM :+1:  please, @luisg123v, could you review it?

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
```diff
@@ -37,13 +37,23 @@ def _validate_currencies_in_internal_transfer(self, payment):
...
         vals = {"debit": line.debit} if line.amount_currency > 0 else {"credit": line.credit}
         sign = 1 if line.amount_currency > 0 else -1
         vals.update({"amount_currency": sign * self.agreed_amount, "currency_id": currency.id})
         return vals
 
+    def _preprocessing_payment_with_agreed_amount(self, payment):
+        """This method is to be inherited in case pre-processing is required on the payments."""
+        return
+
+    def _postprocessing_payment_with_agreed_amount(self, payment):
```
luisg123v: Since this is a hook and everything here is for internal transfers, I think it should be named something like:

```suggestion
    def _postprocess_payments_hooks(self, payment):
```

Same below.

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
```diff
@@ -37,13 +37,23 @@ def _validate_currencies_in_internal_transfer(self, payment):
...
+        """This method is to be inherited in case post-processing is required on the payments."""
+        return
```
luisg123v: Not returning anything, so not even required.

But maybe, for convenience, we could return `True` or `payment`.

Same below.

luisg123v (changes requested):
By reading commit message, it's not clear why the change, there's only a changelog.

luisg123v:
Missing task ID on MR title

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
CLaurelB: Commit message and method names updated.

internal_transfer_with_agreed_amount/wizards/internal_transfer_multicurrency.py
CLaurelB: Returning payment.

CLaurelB:
> Missing task ID on MR title

Since it's related to a ticket, I added the ticket ID on the title.

CLaurelB:
@luisg123v could you review again, please?

luisg123v (approved):
LGTM :+1:

PR #1632, opened 2023-11-08
[IMP] internal_transfer_with_agreed_amount: update tests to use company currency i#24002

CLaurelB:
Change journal and currency in USD to use the company currency in order to be able to run the tests when the company currency is different from USD.

Chagelog:
- Change USD to the company currency.
- Remove USD from variables and replace them with local in some cases.
- Allow reconciliation to the company transfer account.
- Remove unused parameter from create_multicurrency_transfer method.


CLaurelB:

luisg123v (approved):

Issue #1633, opened 2024-01-17
Inaccuracies in Stock Information Displayed by 'stock_by_warehouse' Product Widget


allanpa88:
The 'stock_by_warehouse' product widget is not accurately reflecting the stock levels for products across different warehouses. This issue persists regardless of the complexity of the stock distribution (e.g., multiple units across various warehouses or a single unit in one warehouse).

Multi-Warehouse Inconsistency example:
A product with 10 units, distributed equally in 2 different warehouses.
![image](https://github.com/Vauxoo/addons-vauxoo/assets/3640142/b015a699-1060-4548-8fc3-9f814e4bec9c)

The widget incorrectly displays stock information. This inaccuracy is also evident in warehouses where the product is not stocked.
![ezgif-7-40ef44299f](https://github.com/Vauxoo/addons-vauxoo/assets/3640142/be35e382-faeb-42c7-952b-486e25d87e67)

Single Warehouse Inconsistency example:
A simpler case where a product has only 1 unit in a single warehouse.
The widget still computes and displays an incorrect stock value.
![image](https://github.com/Vauxoo/addons-vauxoo/assets/3640142/6568fe6d-8ac5-4b55-aa48-aa97cb70fadf)

Impact of Issue:
This inaccurate display of stock information can lead to mismanagement of inventory.

PR #1634, opened 2024-01-17
[FIX] stock_by_warehouse: Improve stock information accuracy

allanpa88:
Issue reported on #1633

With the re-compute information the data it's shown as expected:
image

### DIFF
--- a/stock_by_warehouse/models/product_template.py
+++ b/stock_by_warehouse/models/product_template.py
@@ -54,6 +54,7 @@ class ProductTemplate(models.Model):
                 .with_company(warehouse.company_id)
                 .with_context(warehouse=warehouse.id, location=False)
             )
+            tmpl.invalidate_recordset()
             if warehouse_id and warehouse_id.id == warehouse.id:
                 info["warehouse"] = tmpl.qty_available_not_res
             info["content"].append(


CarmenMiranda:
Hi @allanpa88 Just a little improvement to the commit 🙈 What about something like:

[FIX] stock_by_warehouse: Accurancy of Stock information

There was a problem with the contexts that show wrong information for several locations, by removing the cache we are forcing the correct compute of the quantity fields on the product.


And add to the description the keyword `Closes #1634 ` so the Issue is closed after the merge 👍🏼

rolandojduartem:
> Hi @allanpa88 Just a little improvement to the commit 🙈 What about something like:
> 
> [FIX] stock_by_warehouse: Accurancy of Stock information
> 
> There was a problem with the contexts that show wrong information for several locations, by removing the cache we are forcing the correct compute of the quantity fields on the product.
> 
> And add to the description the keyword `Closes #1634 ` so the Issue is closed after the merge 👍🏼

This. An explanation how invalidate_recordset works here would be valuable.

allanpa88:
> Issue reported on #1633
> 
> With the re-compute information the data it's shown as expected: image

Done! Thank you both for your valuable feedback."

@luisg123v 
Could you review and merge please?

cc: @CarmenMiranda @rolandojduartem

CarmenMiranda:
Technically LGTM 👍🏼

luisg123v (approved):
LGTM :+1

luisg123v:
@allanpa88 what's the task for this?

allanpa88:
Hi @luisg123v , the issue report was made on this task. It's not related but the issue was found testing the new module stock_by_warehouse_detailed with has stock_by_warehouse as dependency.

https://www.vauxoo.com/web#id=78019&menu_id=1490&cids=1&action=464&active_id=254&model=project.task&view_type=form

PR #1644, opened 2024-04-11
[REM] sale_margin_percentage: remove margin_percentage field in favor of native one I#25764

xmglord:
The field margin_percentage is no longer needed since native module
sale_margin have the field margin_percent that serve the same propose.
Also the field margin_threshold must be updated since it was used as a
percentage but the native field margin_percent is a ration so we must
follow the same pattern.


xmglord:
@isaako34 and @rolandojduartem please review.

sale_margin_percentage/models/res_company.py
```diff
@@ -4,4 +4,4 @@
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    margin_threshold = fields.Float(digits=(16, 2), help="Margin Threshold for product in order lines")
+    margin_threshold = fields.Float(digits=(18, 4), help="Margin Threshold for product in order lines")
```
rolandojduartem: Why this one?

rolandojduartem (changes requested):

sale_margin_percentage/models/res_company.py
xmglord: I thought it was necessary because instead of being a percentage, 50% for example, now it must be used as a ration, 0.5 in the same example. But checking [the native field](https://github.com/odoo/odoo/blob/15.0/addons/sale_margin/models/sale_order.py#L12) I think is better to remove it.

sale_margin_percentage/models/sale_order.py
```diff
@@ -1,61 +0,0 @@
-from odoo import api, fields, models
-
-
-class SaleOrder(models.Model):
-    _inherit = "sale.order"
-
-    margin_percentage = fields.Float(
```
rolandojduartem, line 7: I think we need a migration script here, it is a stored field.

rolandojduartem:
There errors in the CI, is that expected?

sale_margin_percentage/models/sale_order.py
xmglord: done

xmglord:
@luisg123v please review.

sale_margin_percentage/migrations/15.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,47 @@
...
+    order_lines._compute_margin()
+    _logger.info("The margin percentage of %s records have been recomputed.", len(order_lines))
+
+
+def remove_margin_percentage_field(cr):
+    cr.execute(
+        """
+        ALTER TABLE
+            sale_order
+        DROP COLUMN IF EXISTS
```
luisg123v: This is not necessary, this will be done automatically by Odoo when the field is removed.

sale_margin_percentage/tests/test_sale_order.py
```diff
@@ -37,37 +37,37 @@ def test_01_margin_cost_0(self):
...
         self.assertEqual(sale.order_line.margin, 0.0)
-        self.assertEqual(sale.order_line.margin_percentage, 0.0)
+        self.assertEqual(sale.order_line.margin_percent, 0.0)
 
     def test_04_margin_positive(self):
         """Margin and Margin Percentage when purchase price < sale cost"""
         sale = self.create_so(product=self.product_cost_10, quantity=4, price=12)
         self.assertEqual(sale.order_line.margin, 8.0)
-        self.assertAlmostEqual(sale.order_line.margin_percentage, 16.66, places=1)
+        self.assertAlmostEqual(sale.order_line.margin_percent, 0.1666, places=1)
```
luisg123v: Don't you need to update also places?

Same below.

sale_margin_percentage/tests/test_sale_order.py
```diff
@@ -80,8 +80,9 @@ def test_07_margin_total(self):
         self.create_so_line(sale, product=self.product_cost_5, quantity=4, price=13)
         self.create_so_line(sale, product=self.product_cost_10, quantity=5, price=12)
         self.assertEqual(sale.order_line.mapped("margin"), [48.0, -60.0, -12.0, 8.0, 0.0, 0.0, 32.0, 10.0])
-        self.assertEqual(
-            sale.order_line.mapped("margin_percentage"), [100.0, -100.0, -25.0, 16.67, 0.0, 0.0, 61.54, 16.67]
-        )
+        calculated_margins = sale.order_line.mapped("margin_percent")
+        expected_margins = [1.0, -1.0, -0.25, 0.1667, 0.0, 0.0, 0.6154, 0.1667]
+        for expected, calculated in zip(expected_margins, calculated_margins):
```
luisg123v: Why?

We should avoid loops within tests

sale_margin_percentage/models/sale_order_line.py
```diff
@@ -0,0 +1,30 @@
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
```
luisg123v, line 4: Could you split by model on a separate commit, please?

luisg123v (changes requested):

sale_margin_percentage/migrations/15.0.1.0.1/post-migration.py
xmglord: Magnificent, done.

sale_margin_percentage/tests/test_sale_order.py
xmglord: done

sale_margin_percentage/tests/test_sale_order.py
xmglord: Because if I use `assertEqual` I got error since they are not computational equal and the `assertAlmostEqual` don't support list, the `numpy.testing` have a `assertAlmostEqual` that support list but I don't think use numpy is better than the loop, what do you think?

sale_margin_percentage/models/sale_order_line.py
xmglord: Done

sale_margin_percentage/tests/test_sale_order.py
luisg123v: I think you could round the computed value

Or even better, use `assertRecordValues` that will take care of rounding.

sale_margin_percentage/models/sale_order_line.py
luisg123v: Commit message: it's -> its

xmglord: done

sale_margin_percentage/tests/test_sale_order.py
xmglord: done, the `assertRecordValues`didn't take care of rounding so I rounded the computed value.

luisg123v (approved):
LGTM :+1:

PR #1645, opened 2024-04-15
[FIX] default_warehouse_from_sale_team: Fix case of searching a stock…

keylor2906:
… rule when there is no warehouse parameter, like when is done by stock rules when the move destination location is for a different company https://github.com/Vauxoo/odoo/blob/8af8846c88bd04bcf3fa0bcbe4a720b415f994a4/addons/stock/models/stock_move.py#L830

### DIFF
--- a/default_warehouse_from_sale_team/__manifest__.py
+++ b/default_warehouse_from_sale_team/__manifest__.py
@@ -8,7 +8,7 @@
     "website": "http://www.vauxoo.com",
     "license": "LGPL-3",
     "category": "Inventory/Inventory",
-    "version": "15.0.1.0.0",
+    "version": "15.0.1.0.1",
     "depends": [
         "sale_stock",
         "purchase_requisition",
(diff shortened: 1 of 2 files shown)


suniagajose:
@keylor2906 

technically +1

but, we must waiting for a fucntional review.

i let you know right here when functional review is ready!

thanks

suniagajose (approved):
@keylor2906 

technically +1

but, we must waiting for a fucntional review.

i let you know right here when functional review is ready!

thanks

cc @hugho-ad 
cc @luisg123v 
cc @desdelinux

suniagajose:
Thanks @keylor2906 

Also, can you help us to fw  this up to 16.0 and 17.0.???

Regards,

cc @desdelinux 
cc @hugho-ad 
cc @luisg123v

hugho-ad:
fwp 16.0 https://github.com/Vauxoo/addons-vauxoo/pull/1646

PR #1650, opened 2024-04-25
[FIX] default_warehouse_from_sale_team: Fix case of searching a stock…

CLaurelB:
… rule when there is no warehouse parameter, like when is done by stock rules when the move destination location is for a different company https://github.com/Vauxoo/odoo/blob/8af8846c88bd04bcf3fa0bcbe4a720b415f994a4/addons/stock/models/stock_move.py#L830


CLaurelB:

luisg123v (approved):

PR #1651, opened 2024-05-02
[FIX] stock_by_warehouse: display stock by warehouse in widget T#79454

xmglord:
There was an error when trying to load the stock by warehouse widget caused by the incorrectly use of `onWillUpdateProps` that was change in [1], this commit fixes that error and also adapt the stock by warehouse field to be computed. Finally, some styles were adapted to have a better display and to use recent version of bootstrap.

[1]: https://github.com/odoo/odoo/commit/d682871a

### DIFF
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -16,23 +16,21 @@ class ProductProduct(models.Model):
         compute="_compute_qty_available_not_reserved",
     )
 
-    warehouses_stock = fields.Text(
-        store=False,
-        readonly=True,
-    )
+    warehouses_stock = fields.Json(compute="_compute_warehouses_stock")
 
-    warehouses_stock_location = fields.Text(store=False, readonly=True)
+    warehouses_stock_location = fields.Json(compute="_compute_warehouses_stock")
 
     warehouses_stock_recompute = fields.Boolean(store=False)
 
-    @api.onchange("warehouses_stock_recompute")
-    def _warehouses_stock_recompute_onchange(self):
-        if not self.warehouses_stock_recompute:
-            self.warehouses_stock_recompute = True
-            return
-        self.warehouses_stock = self._compute_get_quantity_warehouses_json()
-        self.warehouses_stock_location = self._compute_get_stock_location()
-        self.warehouses_stock_recompute = True
+    @api.depends("warehouses_stock_recompute")
+    def _compute_warehouses_stock(self):
+        for record in self:
+            record.warehouses_stock = (
+                record._compute_get_quantity_warehouses_json() if record.warehouses_stock_recompute else False
+            )
+            record.warehouses_stock_location = (
+                record._compute_get_stock_location() if record.warehouses_stock_recompute else False
+            )
 
     def _product_available_not_res_hook(self, quants):
         """Hook used to introduce possible variations"""
--- a/stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
+++ b/stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
@@ -1,13 +1,13 @@
 /** @odoo-module **/
 
 import {registry} from "@web/core/registry";
-import {usePopover} from "@web/core/popover/popover_hook";
+import {useService} from "@web/core/utils/hooks";
 import {localization} from "@web/core/l10n/localization";
 import {formatFloat} from "@web/views/fields/formatters";
-import {TextField} from "@web/views/fields/text/text_field";
+import {textField, TextField} from "@web/views/fields/text/text_field";
 import {archParseBoolean} from "@web/views/utils";
-
-const {Component, onWillUpdateProps} = owl;
+import {Component} from "@odoo/owl";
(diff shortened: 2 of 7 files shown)


xmglord:
@rolandojduartem @luisg123v please review.

stock_by_warehouse/tests/test_stock_available_unreserved.py
```diff
@@ -124,12 +124,12 @@ def test01_stock_levels(self):
         self.compare_qty_available_not_res(product_a, 0)
         self.compare_qty_available_not_res(self.template_ab, 0)
 
-        self.assertFalse(product_a.warehouses_stock)
+        self.assertEqual(product_a.warehouses_stock, "{}")
```
xmglord: Even with the Json field the assertFalse didn't worked @luisg123v

luisg123v (approved):
LGTM :+1: 

I'll wait for @rolandojduartem's approval to merge.

stock_by_warehouse/tests/test_stock_available_unreserved.py
luisg123v: I only want to ask: Why doesn't this happen in `stock_by_warehouse_sale`?

xmglord: because it doesn't have test

stock_by_warehouse/models/product_product.py
```diff
@@ -16,23 +16,21 @@ class ProductProduct(models.Model):
...
-            self.warehouses_stock_recompute = True
-            return
-        self.warehouses_stock = self._compute_get_quantity_warehouses_json()
-        self.warehouses_stock_location = self._compute_get_stock_location()
-        self.warehouses_stock_recompute = True
+    @api.depends("warehouses_stock_recompute")
+    def _compute_warehouses_stock(self):
+        for record in self:
+            record.warehouses_stock = (
+                record._compute_get_quantity_warehouses_json() if record.warehouses_stock_recompute else "{}"
```
rolandojduartem: is the else "{}" required after the change to json field?

stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
```diff
@@ -21,25 +21,28 @@ StockAvailabilityPopOver.template = "stock_by_warehouse.StockAvailabilityPopOver
...
-        onWillUpdateProps((nextProps) => this.formatData(nextProps));
+        useSpecialData((orm, props) => {
+            this.formatData(props);
+            this.render();
+        });
     }
 
     initializeVariables() {
-        this.info = JSON.parse(this.props.value);
+        this.info = JSON.parse(this.props.record.data[this.props.name]);
```
rolandojduartem, line 33: is json.parse required after the change to json field?

rolandojduartem:
Screenshot from 2024-05-03 10-28-30

Screenshot from 2024-05-03 10-28-22

Functionally it does not work :(

stock_by_warehouse/models/product_product.py
xmglord: No with False seems to be enough. I've update it

stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
xmglord: in that line no but because is initializing but in the other it is necessary.

xmglord:
> Screenshot from 2024-05-03 10-28-30
> 
> Screenshot from 2024-05-03 10-28-22
> 
> Functionally it does not work :(

For me it is working:
image
 :c

stock_by_warehouse/models/product_product.py
xmglord: Actually removing that and changing it to False solve the change in tests @luisg123v

stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
rolandojduartem: if it is required, it would be required in the both cases or not required in the both cases, that is weird

xmglord: Is weird because if I remove it from that line everything works fine but if removed from formatData then it fails.

rolandojduartem:
Functionally it is OK. @xmglord it did not worked because you had renamed the js file, but you did not changed the manifest, so it was loading nothing. It is Ok functionally

stock_by_warehouse/static/src/components/warehouse/warehouse_field.esm.js
rolandojduartem: OK, NVM, @xmglord, it is required in the both lines, js gets the value as string, the Json.parse converts that to a json object

rolandojduartem:
Just one comment more, the spanish translation was not updated during the migration process. @xmglord 

cc @luisg123v 

PD. After that, @luisg123v could merge this PR whenever he wants, I think it LGTM

xmglord:
@luisg123v already update the translation file so this is ready to merge.

cc: @rolandojduartem

PR #1649, opened 2024-04-25
[MIG] stock_manual_transfer: Migration to 17.0 T#79454

xmglord:

PR #1652, opened 2024-05-06
[MIG] stock_manual_transfer: Migration to 16.0 T#81066

allanpa88:
- Changed name for deprecated modules like stock.location.route to
  stock.route and picking.moves to move.line.ids.
- Added user to group uom to fix unit tests and avoid error for not
  null violation in product uom.
- Improved preferred Route domain to show only stock routes related
  to the supplied warehouse.
- Added return early pattern in action_validate() method to avoid
  duplicated transfers after first validation and state change.


stock_manual_transfer/models/stock_manual_transfer.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    procurement_group_id = fields.Many2one("procurement.group", copy=False)
+    picking_ids = fields.One2many(
+        "stock.picking",
+        string="Transfers",
+        compute="_compute_picking_ids",
+    )
+
+    def action_validate(self):
+        if self.state == "valid":
```
hugho-ad:
```suggestion
        if self.state != "draft":
```

hugho-ad (commented):
Just one comment, then LGTM :+1:

stock_manual_transfer/models/stock_manual_transfer.py
allanpa88: Done!

hugho-ad (approved):

stock_manual_transfer/demo/res_groups_demo.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
```
luisg123v: Demo data shouldn't be marked as noupdate.

stock_manual_transfer/demo/res_groups_demo.xml
```diff
@@ -0,0 +1,7 @@
...
+    <record model="res.groups" id="uom.group_uom">
```
luisg123v: - ID goes before model
- Modify the user, not the group
- I don't like this, this is a technical group, which mean it should be enabled or disabled for all users, but you're enabling it for a single user.

stock_manual_transfer/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import procurement_group
+from . import stock_location
```
luisg123v: `stock_route`

stock_manual_transfer/models/stock_location.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class StockLocationRoute(models.Model):
```
luisg123v: Rename also class.

stock_manual_transfer/tests/test_stock_manual_transfer.py
```diff
@@ -0,0 +1,89 @@
...
+    def create_manual_transfer(self):
+        transfer = Form(self.env["stock.manual_transfer"])
+        transfer.route_id = self.warehouse_route_reception
+        transfer.warehouse_id = self.warehouse
+        transfer = transfer.save()
+        self.create_manual_transfer_line(transfer)
+        return transfer
+
+    def create_manual_transfer_line(self, transfer):
+        self.env.ref("uom.group_uom").write({"users": [(4, self.env.user.id)]})
```
luisg123v: - Not correct, this is not part of a manual transfer creation. If really necessary, this should be in the SetUpClass method.
- Use the command syntax

stock_manual_transfer/tests/test_stock_manual_transfer.py
```diff
@@ -26,9 +26,9 @@ def create_manual_transfer(self):
         return transfer
 
     def create_manual_transfer_line(self, transfer):
+        self.env.ref("uom.group_uom").write({"users": [(4, self.env.user.id)]})
         with Form(transfer) as tr, tr.transfer_line_ids.new() as line:
             line.product_id = self.product
-            line.product_uom_qty = 1
```
luisg123v: Why?

stock_manual_transfer/demo/res_groups_demo.xml
```diff
@@ -0,0 +1,7 @@
...
+        <field name="users" eval="[(4, ref('base.user_admin'))]" />
```
luisg123v: Use the Command syntax

luisg123v (changes requested):
Document performed changes in commit description.

luisg123v:
@allanpa88 Please check the migration done by @xmglord on #1649

FYI @xmglord we will base the migration to 17.0 on top of this one.

stock_manual_transfer/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "category": "Inventory/Inventory",
+    "version": "16.0.1.0.0",
+    "depends": [
+        "stock",
+    ],
+    "data": [
+        "security/res_groups.xml",
+        "security/ir.model.access.csv",
+        "data/ir_sequence_data.xml",
+        "views/stock_location_route_views.xml",
```
luisg123v: Rename file.

luisg123v (changes requested):

stock_manual_transfer/demo/res_groups_demo.xml
allanpa88: This is not required, so I will delete the file. However, manual access will be needed for testing purposes.

The original idea was to create a plug-and-play environment where you could start testing functionality without the need for prior configurations.

stock_manual_transfer/tests/test_stock_manual_transfer.py
allanpa88: It is not needed at this point in the test, as the value is already set to 1. However, we can keep the line even though it is unnecessary.

stock_manual_transfer/tests/test_stock_manual_transfer.py
allanpa88: It's required to avoid a not null Violation error with the product uom relation in the transfer lines.

stock_manual_transfer/models/stock_manual_transfer.py
```diff
@@ -63,6 +63,8 @@ class StockManualTransfer(models.Model):
     )
 
     def action_validate(self):
+        if self.state != "draft":
+            return
```
luisg123v, line 70: How is this possible?

stock_manual_transfer/models/stock_manual_transfer.py
```diff
@@ -28,9 +28,9 @@ class StockManualTransfer(models.Model):
         states={"draft": [("readonly", False)]},
     )
     route_id = fields.Many2one(
-        "stock.location.route",
+        "stock.route",
         string="Preferred Route",
-        domain=[("manual_transfer_selectable", "=", True)],
+        domain="[('manual_transfer_selectable', '=', True), ('supplied_wh_id', '=', warehouse_id)]",
```
luisg123v: Are you sure? Take into account warehouse fields are not mandatory in routes.

luisg123v (changes requested):
Commit message:
- "Changed name for deprecated modules like stock.location.route": not a module but a model
- "picking.moves to move.line.ids": these models don't exist

stock_manual_transfer/models/stock_manual_transfer.py
allanpa88: "I don't have an idea how to reproduce the error. But is an issue reported.
 Please remember that this module is already in use but was installed from the Client Road main app (https://git.vauxoo.com/vauxoo/7llantas)
 
 The PM's group has declared it as an issue that cannot be replicated, but a couple of times, the transfers originated from this module has been found duplicated.

For more context, please refer to the comments by Julio Zerna and Zuly Lira in helpdesk ticket 81066. They are the authors of the return early suggestion as an answer of this issue

https://www.vauxoo.com/web#id=81066&cids=1&menu_id=1431&model=project.task&view_type=form

stock_manual_transfer/models/stock_manual_transfer.py
allanpa88: Yes, I'm experiencing the same issue. 
This feature was requested by Zuly Lira and is already in the production environment. 
When I asked her if this feature was required for all customers, she confirmed that it was a better way to work with the module. This is the origin for that change:
https://git.vauxoo.com/vauxoo/7llantas/-/commit/94a57eca77c58724ef2f3412a0fd921e39d0a420

Here is a snippet of our conversation and her request: 
Zuly Lira, [9/5/24 11:19 A. M.]
Hola @Allan Porras [Vauxoo] recuerdo que hicimos una mejora para que en rutas solo mostrar las relacionadas al almacén seleccionado https://imgur.com/Vq7Mufb y haciendo pruebas ahora en el staging, aparecen todas las rutas https://imgur.com/45Zd5vx
Me ayudas a revisar por favor.
Gracias.

allanpa88, [9/5/24 11:24 A. M.]
Revisando
Igual ese módulo como te había mencionado se va a pasar a addons-vauxoo
o sea, ya no es sólo de Road

allanpa88, [9/5/24 11:25 A. M.]
consideras que es una característica que se deba mantener en el módulo general?

Zuly Lira, [9/5/24 11:25 A. M.]
siiii
es mucho mas funcional así...

And it's true, if you select an incorrect route of the entire list, you could get some functional errors difficult to resolve for a non technical or Odoo expert user.

luisg123v (approved):
LGTM :+1:

stock_manual_transfer/tests/test_stock_manual_transfer.py
```diff
@@ -8,6 +8,7 @@ class TestStock(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env.user.groups_id += cls.env.ref("uom.group_uom")
```
xmglord, line 11: @luisg123v this shouldn't be done, there is a group restriction in the stock manual transfer view over the unit of measure in [here](https://github.com/Vauxoo/addons-vauxoo/blob/16.0/stock_manual_transfer/views/stock_manual_transfer_views.xml#L47) and in v15 it worked fine, the field didn't show and a user without the multi unit of measure group and they can make the manual transfers but now in the v16 you cannot preform a manual transfer if you don't belong to the multi unit group.

image

cc: @allanpa88

PR #1653, opened 2024-05-08
[FIX] internal_transfer_with_agreed_amount: Fixing AssertionError in internal_transfer_with_agreed_amount test

allanpa88:
The test was failing due to an attempt to write to the invisible field 'destination_journal_id'. This occurred because the 'destination_journal_id' field was set to be invisible when the 'is_internal_transfer' field is not True. Consequently, when the 'journal_id' field changed in the form, the 'is_internal_transfer' value changed to False, triggering the error.

This commit resolves the issue and prevents the AssertionError.


rolandojduartem:
@luisg123v this LGTM, the compute of that field is triggered by https://github.com/odoo/odoo/blob/021f2ebd784f35b1872dc065c081b547920e46ee/addons/account/models/account_payment.py#L422, so it makes sense for me.

hugho-ad (approved):

luisg123v (approved):

PR #1655, opened 2024-05-16
[FIX] stock_manual_transfer: Avoid Unit of Measure validation error without group

allanpa88:
A validation error is raised when a user without the "Manage Multiple Units of Measure" access tries to save or validate a manual transfer.

This fix allows any user to create manual transfers, but only users with "Units of Measure" access can modify this attribute on the lines.

This is the answer to the request suggested by German here:
https://github.com/Vauxoo/addons-vauxoo/pull/1652#discussion_r1602424134

I think this is a great solution, better than giving the user access.

Here you can see evidence of the function test:
1. User without group
User without group

2. Without the fix we cannot save or validate the Manual Transfer.
1  Error trying to save without uom permission

3. With the fix any user could create and modify the lines:
validated without permision

### DIFF
--- a/stock_manual_transfer/__manifest__.py
+++ b/stock_manual_transfer/__manifest__.py
@@ -9,7 +9,7 @@
     "website": "https://www.vauxoo.com",
     "license": "LGPL-3",
     "category": "Inventory/Inventory",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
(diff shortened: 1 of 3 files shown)


allanpa88:
Hi @luisg123v could you review this please.

cc: @xmglord

luisg123v (approved):

PR #1658, opened 2024-05-30
[FIX] portal_multicurrency_totals: fix portal access and password setup in demo data t#79454

CLaurelB:
Changelog:
- Updated the field of the record 'grant_portal_access' that is required to grant access to the partner user.
- Updated the function parameters to execute the method 'action_grant_access' in the 'portal.wizard.user' model.
- Updated the reference to the IDs of the users associated with the partner when creating a 'change.password.user' record.

### DIFF
--- a/portal_multicurrency_totals/demo/res_partner_demo.xml
+++ b/portal_multicurrency_totals/demo/res_partner_demo.xml
@@ -2,22 +2,12 @@
 <odoo>
 
     <!-- Grant portal access to the customer -->
-    <record id="grant_portal_access" model="portal.wizard.user">
-        <field
-            name="user_id"
-            model="res.partner"
-            eval="[
-                Command.create(
-                    {
-                        'partner_id': ref('base.res_partner_3'),
-                        'email': obj(ref('base.res_partner_3')).email,
-                        'is_portal': True,
-                    }
-                ),
-            ]"
-        />
+    <record id="grant_portal_access" model="portal.wizard">
+        <field name="partner_ids" eval="[Command.set([ref('base.res_partner_3')])]" />
     </record>
-    <function model="portal.wizard.user" name="action_grant_access" eval="[ref('grant_portal_access')]" />
+    <function model="portal.wizard.user" name="action_grant_access">
+        <value model="portal.wizard" eval="obj(ref('grant_portal_access')).user_ids.ids" />
+    </function>
 
     <!-- Set a password for the portal user so it may log-in -->
     <record id="set_password_portal_user" model="change.password.wizard">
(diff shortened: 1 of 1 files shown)


CLaurelB:

portal_multicurrency_totals/demo/res_partner_demo.xml
```diff
@@ -27,7 +17,7 @@
             eval="[
                 Command.create(
                     {
-                        'user_id': obj(ref('base.res_partner_3')).user_ids.id,
+                        'user_id': obj(ref('base.res_partner_3')).user_ids.ids[0],
```
luisg123v: Why this?

The partner should have only one user, shouldn't it.?

portal_multicurrency_totals/demo/res_partner_demo.xml
```diff
@@ -2,22 +2,12 @@
...
-                    }
-                ),
-            ]"
-        />
+    <record id="grant_portal_access" model="portal.wizard">
+        <field name="partner_ids" eval="[Command.set([ref('base.res_partner_3')])]" />
     </record>
-    <function model="portal.wizard.user" name="action_grant_access" eval="[ref('grant_portal_access')]" />
+    <function model="portal.wizard.user" name="action_grant_access">
+        <value model="portal.wizard" eval="obj(ref('grant_portal_access')).user_ids.ids" />
```
luisg123v, line 9: Since it's only one param, the eval could be included into the `<function>` tag

luisg123v (changes requested):
Commit title is too long

portal_multicurrency_totals/demo/res_partner_demo.xml
CLaurelB: The partner could be linked to multiple users. However, now .id doesn’t return the first ID as it did before. That’s why .ids with an index has to be used.

portal_multicurrency_totals/demo/res_partner_demo.xml
CLaurelB: Even if the object is from the portal.wizard model and the function is from the portal.wizard.user model?

luisg123v: Good point, I forgot that.

portal_multicurrency_totals/demo/res_partner_demo.xml
luisg123v: It's the same behavior:
- `.id` returns an integer
- `.ids` returns a list

I think it's not necessary. But if you prefer to assure that, I suggest `user_ids[0].id` instead

Even though it's technically possible, a user should only be linked to a single partner. The field is a o2m because the inverse field is a m2o.

CLaurelB: Suggestion applied.

luisg123v (approved):
LGTM :+1:

PR #1662, opened 2024-06-05
[IMP] default_warehouse_from_sale_team: make abstract model more general T#79454

xmglord:
In v14 when a record is created, the fields with default values were not in the vals passed to the create method but since v15 this behaviour changed so this commit makes the validation of the abstract model more robust considering also the default value.

### DIFF
--- a/default_warehouse_from_sale_team/models/default_warehouse_mixin.py
+++ b/default_warehouse_from_sale_team/models/default_warehouse_mixin.py
@@ -47,7 +47,9 @@ class DefaultWarehouseMixin(models.AbstractModel):
 
     def _get_salesteam_from_vals(self, vals):
         """Determine sales team from creation values"""
-        if "name" in vals:
+        name_field = self._fields["name"]
+        default_name = name_field.default(self) if callable(name_field.default) else name_field.default
+        if vals.get("name", default_name) != default_name:
             # Already has a name, so salesteam won't be used anyway
             return self.env["crm.team"]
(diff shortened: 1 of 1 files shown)


xmglord:
@luisg123v please review, this is related to the stock manual transfer MR in [here](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/1139)

default_warehouse_from_sale_team/models/default_warehouse_mixin.py
```diff
@@ -47,7 +47,9 @@ def create(self, vals):
 
     def _get_salesteam_from_vals(self, vals):
         """Determine sales team from creation values"""
-        if "name" in vals:
+        name_field = self._fields["name"]
+        default_name = name_field.default(self) if callable(name_field.default) else name_field.default
+        if vals.get("name", default_name) != default_name:
```
luisg123v, line 52: This doesn't seem correct to me. Why is this necessary?

luisg123v (changes requested):
CC @teposteaj

default_warehouse_from_sale_team/models/default_warehouse_mixin.py
xmglord: As discussed earlier I'll close this PR since [this new PR](https://github.com/Vauxoo/addons-vauxoo/pull/1664) solve the issue of why this was created.

PR #1663, opened 2024-06-05
[IMP]  stock_manual_transfer: Add user for default in Manual Transfers group T#79454

ajteposte:
>>>
By default, it is recommended that the admin and root users have all permissions due to their roles and to facilitate functional testing.
>>> 

Added the root user (`base.user_root`) and the admin user (`base.user_admin`) to the `group_stock_manual_transfer` group. Since during functional testing in the TyP module, it was detected that they were not added by default.

### DIFF
--- a/stock_manual_transfer/security/res_groups.xml
+++ b/stock_manual_transfer/security/res_groups.xml
@@ -3,6 +3,7 @@
 
     <record id="group_stock_manual_transfer" model="res.groups">
         <field name="name">Can make manual transfers</field>
+        <field name="users" eval="[Command.link(ref('base.user_root')), Command.link(ref('base.user_admin'))]" />
     </record>
 
 </odoo>


stock_manual_transfer/security/res_groups.xml
```diff
@@ -3,6 +3,7 @@
 
     <record id="group_stock_manual_transfer" model="res.groups">
         <field name="name">Can make manual transfers</field>
+        <field name="users" eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]" />
```
luisg123v: Use the Command syntax, please

luisg123v (changes requested):
Commit title should reference the module, not the repo.

stock_manual_transfer/security/res_groups.xml
ajteposte: @luisg123v I make the change, is it correct how I did it?

luisg123v: Not applied correctly, I meant:

`<field name="users" eval="[Command.link(ref('base.user_root')), Command.link(ref('base.user_admin'))]" />`

For more info, read [this](https://github.com/orgs/OCA/discussions/157#discussioncomment-8753910)

luisg123v (changes requested):
Regarding commit presentation:
- Module name is in plural, but the module is in singular (transfer instead of transfers)
- In author, name seems to be inverted

luisg123v (approved):
LGTM :+1:

luisg123v:
@TeposteAJ please submit PRs also for 15.0 and 16.0

PR #1664, opened 2024-06-05
[FIX] stock_manual_transfer: name shouldn't be saved in a new record T#79454

xmglord:
The name shouldn't be present in the vals list when creating a record and this commit fixes that by making it readonly instead of force_save which lead to the same functional behaviour but without adding the field to the vals list.


xmglord:
@luisg123v please review.

cc: @TeposteAJ

luisg123v (approved):
LGTM :+1:

PR #1670, opened 2024-06-13
[MIG] internal_transfer_with_agreed_amount: Migration to 17.0 t#71297

CLaurelB:
Changelog:
- Replace `attrs` with an expression for the `invisible` attribute in the view.
- Add payment amount and currency update after the move is set to draft. This is because the undoing of the reconciliation triggers an update on the payment, and the amount gets updated to the previous agreed amount along with its corresponding currency. This happens when the agreed amount is edited.
- Enable the euro currency in the test since it's required for multicurrency transfers.
- Fix ambiguous variable name.


CLaurelB:
@randall-vx @luisg123v could you review, please?

internal_transfer_with_agreed_amount/views/account_payment_views.xml
```diff
@@ -12,27 +12,15 @@
...
                     string="Confirm with an agreed amount"
                     groups="internal_transfer_with_agreed_amount.allow_confirm_internal_transfer_with_agreed_amount"
-                    attrs="{
-                        'invisible': [
-                            '|',
-                            ('is_internal_transfer', '=', False),
-                            ('state', '!=', 'draft')
-                        ]
-                    }"
+                    invisible="is_internal_transfer == False or state != 'draft'"
```
luisg123v: `not is_internal_transfer`

Same below.

internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
```diff
@@ -88,7 +88,7 @@ def test_02_transfer_eur_local(self):
         )
 
         # Check journal item in EUR
-        eur_aml = aml.full_reconcile_id.reconciled_line_ids.filtered(lambda l: l != aml)
+        eur_aml = aml.full_reconcile_id.reconciled_line_ids.filtered(lambda line: line != aml)
```
luisg123v, line 91: This commit goes in the previous version.

internal_transfer_with_agreed_amount/i18n/es.po
```diff
@@ -0,0 +1,129 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* internal_transfer_with_agreed_amount
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0+e\n"
```
luisg123v: Missing updating translations. At the OCA this step is not required because they handle translations on a separate place (Weblate), but we don't.

luisg123v (changes requested):

internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
CLaurelB: Moved to the previous version in the following PR: https://github.com/Vauxoo/addons-vauxoo/pull/1671

luisg123v (approved):
LGTM :+1:

PR #1672, opened 2024-06-20
[IMP] stock_manual_transfer: rename group's external ID T#79454

luisg123v:
Rename group's external ID, from:
`group_stock_manual_transferp` → `group_user`
to follow usual naming conventions.

The reason for the old name is first implementation of this was not as a standalone module.

In addition, IDs and names of model access rights are also modified to
follow conventions.


stock_manual_transfer/security/ir.model.access.csv
```diff
@@ -1,3 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_stock_can_stock_manual_transfer,stock_can_stock_manual_transfer,model_stock_manual_transfer,group_stock_manual_transfer,1,1,1,1
```
CLaurelB, line 2: Does changing the ID and name of the access right require updating the records in the ir_model_data and ir_model_access tables?

Additionally, it seems this change is missing from the commit description.

luisg123v: @CLaurelB 

> Does changing the ID and name of the access right require updating the records in the ir_model_data and ir_model_access tables?

No, because these are updatable records (`noupdate=0`). It would be similar to update views.

> Additionally, it seems this change is missing from the commit description.

I didn't include it because I didn't consider it as necessary, but done, "better explicit than implicit".

CLaurelB:
LGTM 👍

CLaurelB (approved):

TeposteAJ:
@luisg123v  LGTM 👍

ajteposte (approved):
@luisg123v LGTM 👍

PR #1673, opened 2024-07-06
[IMP] default_warehouse_from_sale_team: hook to fill allowed sales teams T#79454

luisg123v:
Implement a hook to fill allowed sales teams in users that are already
a member of any team.

Since this module implements a feature to restrict which sales teams a
user may be a member of, users that already belong to any team are
configured to be allowed for those teams, to avoid
inconsistencies between allowed and already-configured memberships. In
other words, if a user already belongs to a team, it most likely means
they should be allowed to belong to it, so allowance is granted.

### DIFF
--- a/default_warehouse_from_sale_team/hooks.py
+++ b/default_warehouse_from_sale_team/hooks.py
@@ -1,0 +1,25 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(env):
+    fill_user_allowed_salesteams(env)
+
+
+def fill_user_allowed_salesteams(env):
+    """Fill allowed sales teams in users that are already members of any team.
+
+    Since this module implements a feature to restrict which sales teams a user may be a member of, users that already
+    belong to any team are configured to be allowed for those teams, to avoid inconsistencies between allowed and
+    already-configured memberships. In other words, if a user already belongs to a team, it most likely means they
+    should be allowed to belong to it, so allowance is granted.
+    """
+    teams_per_user = env["crm.team.member"]._read_group(
+        domain=[],
+        groupby=["user_id"],
+        aggregates=["crm_team_id:recordset"],
+    )
+    for user, teams in teams_per_user:
+        user.sale_team_ids |= teams
+    _logger.info("Field 'Allowed sales Teams' has been set to %d users.", len(teams_per_user))
--- a/default_warehouse_from_sale_team/__init__.py
+++ b/default_warehouse_from_sale_team/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from .hooks import post_init_hook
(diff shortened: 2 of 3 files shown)


default_warehouse_from_sale_team/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(env):
+    fill_user_allowed_salesteams(env)
+
+
+def fill_user_allowed_salesteams(env):
+    """Fill allowed salesteams in users that are already members of any team
```
xmglord: missing dot at the end of line

default_warehouse_from_sale_team/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Since this module implements a feature to restrict what sales teams a user may be member of, users that already
```
xmglord: what sales teams -> **which** sales teams
a user may be member of -> a user may be **a** member of
inconcistencies -> incon**s**istencies

xmglord (changes requested):

xmglord:
commit description:
f.ill -> fill
missing final dot at the first paragraph
same comments as in docstring

default_warehouse_from_sale_team/hooks.py
luisg123v: TL;DR
My understanding was 1st line of a multi-line docstring shouldn't include a trailing dot, similar as commit titles. However, by reading [the docstring conventions for multi-line comments](https://peps.python.org/pep-0257/#multi-line-docstrings), I see, while not explicitly mentioned, the provided example includes it. For this, I've changed it.

Short answer: applied, thanks!

luisg123v:
@xmglord Fixes to commit messages applied.

default_warehouse_from_sale_team/hooks.py
xmglord: Thanks to you for the docstring conventions link.

xmglord:
> @xmglord Fixes to commit messages applied.

Not applied yet

default_warehouse_from_sale_team/hooks.py
luisg123v: Sorry, I didn't see this comment, applied. I modified the whole message for clarity.

BTW, I don't think replacing "what" by "which" is necessary on this case, but applied anyway.

xmglord (approved):
LGTM :+1:

default_warehouse_from_sale_team/hooks.py
xmglord: Since is not necessary lets say it was a nitpick

TeposteAJ:
LGTM :+1:

ajteposte (approved):
LGTM 👍

desdelinux (approved):

PR #1674, opened 2024-07-07
[IMP] sale_margin_percentage: optimize migration script T#78545

luisg123v:
There's a migration script to recompute sales marging on sale order lines, which was raising memory errors with databases that contain many lines.

The following optimizations are applied:
- Don't call compute method directly but mark the field to be computed. This avoids one write per record and makes the ORM to take care of batches.
- Read currency directly from the line, instead of from the pricelist which in turn requires reading the order.
- Don't read currency if quantity is zero as it won't be used anyway


CLaurelB (approved):

CLaurelB:
LGTM 👍

PR #1675, opened 2024-07-08
[FIX] product_cost_usd: order in recompute purchase price I#26616

xmglord:
The `sale_stock_margin` module recomputes the purchase price in sale
order lines and calls `super()` only over the lines without a stock
move related to it. This causes this module to not recompute the lines
with stock moves related, leading to incorrect computation of the
purchase price.

This commit adds the `sale_stock_margin` module as a dependency in
order to first run this module compute.


xmglord:
@JulioSerna please review.

cc: @desdelinux

JulioSerna (approved):

PR #1676, opened 2024-07-15
[IMP] sale_margin_percentage: make purchase price editable I#26616

xmglord:
This module was originally created in v11 for a specific customer, and
the purchase price needed to be non-editable for some reason. Now the
purchase price does not need to be non-editable. If it is needed, it
should be managed in the view because the field is natively editable.

### DIFF
--- a/sale_margin_percentage/models/sale_order_line.py
+++ b/sale_margin_percentage/models/sale_order_line.py
@@ -7,7 +7,6 @@ class SaleOrderLine(models.Model):
     margin_threshold = fields.Float(
         default=lambda self: self.env.user.company_id.margin_threshold, help="Limit margin set in sales configuration"
     )
-    purchase_price = fields.Float(readonly=True, help="Price purchase of product")
 
     @api.depends("price_subtotal", "product_uom_qty", "purchase_price")
     def _compute_margin(self):


luisg123v:
> Since this customer is now in v15, the purchase price does not need to be non-editable.

I don't see the cause-effect relation here.

xmglord:
> > Since this customer is now in v15, the purchase price does not need to be non-editable.
> 
> I don't see the cause-effect relation here.

It was not ready :c 
Now please review @luisg123v

sale_margin_percentage/models/sale_order_line.py
```diff
@@ -7,7 +7,7 @@ class SaleOrderLine(models.Model):
     margin_threshold = fields.Float(
         default=lambda self: self.env.user.company_id.margin_threshold, help="Limit margin set in sales configuration"
     )
-    purchase_price = fields.Float(readonly=True, help="Price purchase of product")
+    purchase_price = fields.Float(help="Price purchase of product")
```
luisg123v: Just remove inheritance

luisg123v (changes requested):
Update PR's description to match commit's description

sale_margin_percentage/models/sale_order_line.py
xmglord: done

luisg123v (approved):

PR #1677, opened 2024-07-26
[FIX] default_warehouse_from_sale_team: remove incorrect record rule T#79454

luisg123v:
There was a rule to grant read-only access to pickings. However, the correct way to grant read-only access to a model is to uncheck the "Read" field on the desired rule, which would cause that rule to not be considered when verifying accesses.


ajteposte (approved):
Looks good to me. c:
 :+1:

isaako34:
LGTM 👍

isaako34 (approved):

luisg123v:
Thanks a lot @TeposteAJ @isaako34.

PR #1678, opened 2024-07-27
[IMP] default_warehouse_from_sale_team: make ineritable bypassed rules

luisg123v:
We allow to bypass record rules related to sales teams when confirming a  sale order, so deliveries may be created even if the involved routes belong to other warehouses the current user has no access to.

This commit allows to extend those rules.

Forward-port of: #1537


luisg123v:
Merging since this is a clean forward-port of a change missing from migration (#1537).

PR #1680, opened 2024-08-27
[FIX] product_cost_usd: correctly set purchase price in multicompany I#27116

xmglord:
Since each company has its own purchase price, we must explicitly set the company we are working with (the one in the sale order line) to correctly set the purchase price.


xmglord:
@luisg123v please review. I'm not sure why CI is red it seams a problem with the docker configuration.

luisg123v (approved):
LGTM :+1:

PR #1682, opened 2024-08-29
[FIX] typ: Missing label in routes view

ajteposte:
The fields for warehouses and manual_transfers_selectable were placed on a single line. There was no clarity on what they represented, as they were just two checkboxes without the appropriate labels. Therefore, the latter was moved to a separate field and made invisible where it was previously displayed.

Related https://git.vauxoo.com/vauxoo/typ/-/issues/484

#### Form Routes View.
* Before development. Without the `warehouses` checkboxes checked. It doesn't show anything.           
image   
* After development. Without the `warehouses` & `manual_transfers` checkboxes checked. There is a single checkbox per line, and each field has its corresponding label, which provides greater clarity.          
image   

#### Form Routes View.    
* Before development. With the checkboxes checked, it shows a selection field.            
image     
* After development. With the checkboxes checked, it shows a selection field for warehouses.                
image

### DIFF
--- a/stock_manual_transfer/views/stock_route_views.xml
+++ b/stock_manual_transfer/views/stock_route_views.xml
@@ -5,7 +5,7 @@
         <field name="model">stock.route</field>
         <field name="inherit_id" ref="stock.stock_location_route_form_view" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='warehouse_selectable']" position="after">
+            <xpath expr="//div[field[@name='warehouse_ids']]" position="after">
                 <field name="manual_transfer_selectable" string="Manual transfers" />
             </xpath>
         </field>


ajteposte:
@luisg123v  Could you review, please? :smiling_face_with_tear:

stock_manual_transfer/views/stock_route_views.xml
```diff
@@ -5,7 +5,7 @@
         <field name="model">stock.route</field>
         <field name="inherit_id" ref="stock.stock_location_route_form_view" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='warehouse_selectable']" position="after">
+            <xpath expr="//div[field[@name='warehouse_ids']]" position="after">
```
luisg123v, line 8: For readability, I think it would be better to reference the selectable field:

```suggestion
            <xpath expr="//div[field[@name='warehouse_selectable']]" position="after">
```

(I had to check original code to understand this)

luisg123v (changes requested):
Regarding commit message:
- Title: `typ` → `stock_manual_transfer`
- Description doesn't match, because no field is being made invisible
- Since this is a community project, referencing private issues wouldn't make sense because they won't be accessible anyway.

luisg123v:
Superseded by #1683, where the correct source repo was used.

PR #1683, opened 2024-08-29
[FIX] stock_manual_transfer: Missing label in routes view

ajteposte:
The fields for `warehouses_selectable` and `manual_transfers_selectable` were placed on a single line. There was no clarity on what they represented, as they were just two checkboxes without the appropriate labels.

Therefore, `manual_transfers_selectable` was placed on its own line.

#### Form Routes View.
* Before development. Without the `warehouses` checkboxes checked. It doesn't show anything.           
image   
* After development. Without the `warehouses` & `manual_transfers` checkboxes checked. There is a single checkbox per line, and each field has its corresponding label, which provides greater clarity.          
image   

#### Form Routes View.    
* Before development. With the checkboxes checked, it shows a selection field.            
image     
* After development. With the checkboxes checked, it shows a selection field for warehouses.                
image

### DIFF
--- a/stock_manual_transfer/views/stock_route_views.xml
+++ b/stock_manual_transfer/views/stock_route_views.xml
@@ -5,7 +5,7 @@
         <field name="model">stock.route</field>
         <field name="inherit_id" ref="stock.stock_location_route_form_view" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='warehouse_selectable']" position="after">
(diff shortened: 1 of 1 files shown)


luisg123v (changes requested):
In commit description:
- "Therefore, manual_transfers_selectable was moved to a separate field." → to its own line

luisg123v (approved):
LGTM :+1:

PR #1684, opened 2024-09-05
[FIX] default_warehouse_from_sale_team: usability for readonly users T#79454

luisg123v:
Limited users are able to see some pickings in readonly mode, but there were some usability issues that are fixed in this commit:
- It was possible to mark as TODo and check availability, because those actions don't require write access to the pickings. This is fixed by hiding involved buttons.
- Even though validating a picking does require access, it's also hidden because there's no point on showing a button that can't be used anyway.
- Similarly, the Print button is also hidden when the picking is not done yet, because it requires access to mark the picking as printed.
- The Return button is also hidden. Even though it doesn't require write access to the current picking, it will require creation access on the new picking.
- The Cancel button is not hidden because it requires being inventory administrator, which is not compatible with being a limited user.
### Dummy MR:
- [vauxoo/typ!1254](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/1254)

### DIFF
--- a/default_warehouse_from_sale_team/__manifest__.py
+++ b/default_warehouse_from_sale_team/__manifest__.py
@@ -19,6 +19,7 @@
         "views/ir_sequence_views.xml",
         "views/res_users_views.xml",
         "views/stock_picking_type_views.xml",
+        "views/stock_picking_views.xml",
         "security/res_groups_security.xml",
         "security/ir_rule_security.xml",
     ],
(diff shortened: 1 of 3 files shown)


ajteposte (approved):
Hi @luisg123v , I agree with the proposal. The code looks good to me  :relaxed:

rolandojduartem:
LGTM :+1:

rolandojduartem (approved):

PR #1688, opened 2025-01-09
[FIX] default_warehouse_from_sale_team: validate cache access when no record exists i#28375

CLaurelB:
Add a validation to avoid errors when accessing the cache without a record. The change ensures that the `team_id` value is only retrieved from the cache when a valid `account.move` record exists.


CLaurelB:

cc. @desdelinux

luisg123v (approved):
LGTM :+1: 

Could you create a dummy MR in TyP to ensure it's now green, please?

CC @andreagidaltig

PR #1691, opened 2025-02-12
[IMP] stock_custom_exchange_rate_date: add the module price t#84027

andreagidaltig:
[Task #84027](https://www.vauxoo.com/web#id=84027&cids=1&menu_id=1530&active_id=647534&model=project.task&view_type=form)

Define the price of the module and keep only the necessary keys in the manifest to comply with Odoo's App Store guidelines.


luisg123v (changes requested):
This is a community module (public repo) under the LGPL license, I don't think it would make much of a sense to charge for it.

PR #1697, opened 2025-04-08
[ADD] base_sync_login_email: module to Restrict editing emails of users I#29116

luisg123v:
It is natively possible for an internal user to edit the e-mail of the 
contact associated to their user. This could be a security risk, because 
someone could (accidentally?) change their own e-mail to an address 
different from the one used to sign up (e.g. an e-mail outside the 
organization), which could lead to an external person to receive sensitive 
information or intended to be sent only to internal users.

To prevent the above, this new module disables editing e-mails of contacts 
associated with internal users if their e-mail matches their user login. 
Now, to edit an e-mail, you will need to edit the user login, which will 
update contact e-mails automatically, ensuring both login and e-mail are 
synced.

This adds two security layers:
- It prevents accidental or unauthorized changes to contact e-mails.
- It restricts e-mail modifications to administrators, as only they have 
  permission to edit user records.

### DIFF
--- a/base_sync_login_email/README.rst
+++ b/base_sync_login_email/README.rst
@@ -1,0 +1,72 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.png
+    :alt: License: LGPL-3
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+
+===========================
+Sync User Login With E-Mail
+===========================
+
+This module ensures the e-mail address of a contact associated with an internal
+user remains synchronized with the user's login.
+
+**Why?**
+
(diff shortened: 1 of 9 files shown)


luisg123v:
CC @desdelinux

moylop260:
👍

moylop260:
FYI Odoo has fixed from frontend in the following PR:
 - https://github.com/odoo/odoo/pull/111708

(Even if it is reproduced for other instances with custom checkouts)

UPDATE: I have created the following issue
 - https://github.com/odoo/odoo/issues/205635

moylop260 (approved):
👍

PR #1698, opened 2025-05-31
[FIX] base_sync_login_email: constraint raised unexpectedly I#29788

luisg123v:
Currently, if the module is installed and there are already partners whose
email doesn't match the user's one, the constraint is triggered under
unexpected circumstances. For instance, when enabling/disabling technical
groups, which are the ones used by config settings not to grant access but
to enable features globally (e.g. product variants, multi warehouses, etc).

This is because, when those groups are enabled or disabled, they are added
or removed from all internal users, even archived ones. This makes the
`partner_share` field to be recomputed on the related partners, which in
turn triggers the constraint.

To solve this, the `partner_share` field is removed from the constraint's
dependencies. It was initially added so that the constraint was triggered
when a non-internal user became an internal one, but that situation
shouldn't be as common, and such a change may only be made by an
administrator anyway.

### DIFF
--- a/base_sync_login_email/models/res_partner.py
+++ b/base_sync_login_email/models/res_partner.py
@@ -6,7 +6,7 @@ from odoo.tools import single_email_re
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    @api.constrains("email", "partner_share", "active")
+    @api.constrains("email", "active")
     def _check_email_internal_user(self):
         """If the partner belongs to an internal user, validate the email matches the user login"""
         for partner in self:


luisg123v:
Hi @moylop260,

Could you review, please?

CC @oscarolar @desdelinux

moylop260:
> @luisg123v 
>
> for instance when enabling/disabling technical groups...
> even if the user and partner are archived.

Let me know more info, please

What is the case of use enabling technical groups for an archived user where its `user.login` is different to `partner.email`?

Second question,
What do you think about fixing the record matching the emails? (Is it not possible?)

luisg123v:
@moylop260,

> What is the case of use enabling technical groups for an archived user where its user.login is different to partner.email?

I mean the groups that are enabled/disabled to everybody, those are used to enable/disable features instead of granting permissions (e.g. `stock.group_multy_warehouse` to enable  the feature of multiple warehouses). In other words, groups used by `res.config.settings`. Those are called technical groups.

But it's not even needed to change them, you can reproduce it by simply saving any config setting.

> What do you think about fixing the record matching the emails?

It's possible, but IMO we can't (shouldn't) rely on all data to be fixed for this module to work, I think existing data should be left alone. Besides, the constraint will be evaluated for all internal users every time a `rs.config.settings` is saved.

moylop260:
OMG I got it

Thank you for your explanation

👍 after commit message explains in a better way the `res.config.settings` issue

luisg123v:
@moylop260 I have improved commit message to explaing on a better way the issue.

Could you check if it's clear enough now, please?

Regards,

moylop260:
> @moylop260 I have improved commit message to explaing on a better way the issue.
> 
> Could you check if it's clear enough now, please?
> 


```diff
-  >[FIX]base_sync_login_email: constraint raised unexpectedly
+  [FIX] base_sync_login_email: Constraint raised unexpectedly
```

moylop260:
The message is good! ty

Only fix typo, please

luisg123v:
Thanks @moylop260.

I have added the missing space. Regarding uppercasing 1s letter, my understanding is lowercase is expected; the Odoo team has even corrected that to me in reviews before, and most commits are lowercased.

moylop260:
Could you remove the `>` in the start of the message `>[FIX]...`

luisg123v:
Done @moylop260

moylop260 (approved):
👍 

cc @fernandahf

PR #1700, opened 2025-07-01
 [16.0][MIG] controller_report_xls: Migration to 16.0. t#87779

CDanielSanSeg:
[MIG] controller_report_xls: Migration to 16.0


hugho-ad (approved):
LGTM :+1:

luisg123v:
Please fix PR title.

hugho-ad:
@luisg123v 
done

controller_report_xls/controllers/main.py
```diff
@@ -30,8 +30,8 @@
 import xlwt
 from cssutils import parseString
 from lxml import etree
-from openerp.addons.report.controllers import main
-from openerp.addons.web.http import request, route
+from odoo.addons.report.controllers import main
+from odoo.addons.web.http import request, route
```
luisg123v: This is an "issue" coming from the previous version, it therefore could be fixed there. This is not strictly part of the migration.

controller_report_xls/controllers/main.py
```diff
@@ -30,9 +30,10 @@
 import xlwt
 from cssutils import parseString
 from lxml import etree
+from werkzeug import url_decode
+
 from odoo.addons.report.controllers import main
 from odoo.addons.web.http import request, route
-from werkzeug import url_decode
```
luisg123v: This is autoformatting, it should be in a dedicated commit where `pre-commit-vauxoo` autofixes are applied. Besides, this could be applied in the previous version.

Same for other autoformatting changes.

controller_report_xls/controllers/main.py
```diff
@@ -30,9 +30,10 @@
 import xlwt
 from cssutils import parseString
 from lxml import etree
+from werkzeug import url_decode
```
luisg123v: Commit title: module name is incomplete

controller_report_xls/controllers/main.py
```diff
@@ -43,6 +44,7 @@ def get_css_style(csstext, style):
     cssutils.log.setLevel(logging.CRITICAL)
     cssnode = parseString(csstext)
     stylesheet = cssnode.cssRules
```
luisg123v: Move to the previous version.

controller_report_xls/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# To avoid warnings in the module controller_report_xls
+cssutils==2.7.1
```
luisg123v: - Does this file do anything? Requirements file should be placed at the top of the repository.
- Use `test-requirements.txt` instead, so the check to avoid requirements file doesn't complain
- Adjust title to use `ADD` and to mention the requirements file as module

controller_report_xls/i18n/es_ES.po
```diff
@@ -0,0 +1,15 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 8.0\n"
```
luisg123v: The file `es.po` should be regenerated as part of the migration. Besides, remove file `es_ES.po` and similar ones in the previous version.

controller_report_xls/controllers/main.py
```diff
@@ -0,0 +1,265 @@
+###########################################################################
+#    Module Writen to OpenERP, Open Source Management Solution
```
luisg123v: Remove all headers, in the previous version. This is a very old syntax.

controller_report_xls/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "XLS Report Controller",
+    "version": "16.0.1.0.0",
+    "author": "Vauxoo",
+    "category": "Tools",
+    "website": "http://www.vauxoo.com/",
```
luisg123v: https

controller_report_xls/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "license": "AGPL-3",
```
luisg123v: LGPL-3

controller_report_xls/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    ],
+    "external_dependencies": {"python": ["cssutils"]},
+    "demo": [],
+    "data": [],
+    "test": [],
+    "js": [],
+    "css": [],
+    "qweb": [],
+    "installable": True,
+    "auto_install": False,
```
luisg123v: Remove all empty keys, see:

https://git.vauxoo.com/vauxoo/project-template/-/merge_requests/129

luisg123v (changes requested):

luisg123v:
Missing task ID on PR title

CDanielSanSeg:
@luisg123v can you check this MR?

controller_report_xls/requirements.txt
luisg123v: Missing the following part:

> mention the requirements file as module

I mean, the touched module in this case is the requirements file itself, not the module.

BTW, in both the requirements and the commit message you say it's to avoid warnings. I think that is not accurate because, AFAIK, if the requirement is not meet, the module fails to install. In other words, it's not a warning but an actual error.

I'd say instead:

```
[ADD] test-requirements.txt: add cssutils as requirement

Because it's required by the module `controller_report_xls`.
```

And in the requirements:

`# Required by controller_report_xls`

controller_report_xls/i18n/es_ES.po
luisg123v: File was not regenerated in this version, still pointing to 15.0.

hugho-ad:
@luisg123v 

comments Done

luisg123v (approved):
LGTM :+1:

PR #1707, opened 2025-09-19
[IMP] internal_transfer_with_agreed_amount: avoid reliance on existing configuration in tests i#30486

CLaurelB:
Update the test setup to be deterministic and independent of the current environment configuration. A new company is created with USD as its main currency, ensuring consistency across different environments. EUR is set as the foreign currency, explicitly guaranteeing that the local and foreign currencies always differ.

Additionally, the required accounts and bank journals are created for the company, so that the tests have all the necessary records available without relying on pre-existing data. This makes the tests fully self-contained.

### DIFF
--- a/internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
+++ b/internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
@@ -6,20 +6,45 @@ class TestInternalTransferWithAgreedAmount(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        main_company = cls.env.ref("base.main_company")
-        main_company.transfer_account_id.write({"reconcile": True})
-        cls.currency = main_company.currency_id
-        cls.currency_eur = cls.env.ref("base.EUR")
-        bank = cls.env["account.journal"].search([("type", "=", "bank")], limit=1)
-        cls.bank_journal = bank.copy()
-        cls.bank_journal.write({"name": "Bank " + cls.currency.name, "code": "BNK68"})
-        cls.bank_journal_eur = bank.copy({"currency_id": cls.currency_eur.id})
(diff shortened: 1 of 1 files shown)


edy1192:
LGTM 👍🏼 

@hugho-ad or @hbto could you check this, please?

internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
```diff
@@ -7,19 +8,63 @@ class TestInternalTransferWithAgreedAmount(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
         main_company = cls.env.ref("base.main_company")
-        main_company.transfer_account_id.write({"reconcile": True})
+        if not main_company.transfer_account_id:
+            transfer_account = cls.env["account.journal"].search(
+                [("type", "=", "asset_current"), ("company_id", "=", main_company.id)], limit=1
+            )
+            if not transfer_account:
```
luisg123v: Tests should be deterministic. Tat means, their flow should not vary depending on the current configuration, because you can't assure they would work on environments different from the one used in the CI.

luisg123v (changes requested):

internal_transfer_with_agreed_amount/tests/test_internal_transfer_with_agreed_amount.py
CLaurelB: Should I create the records instead of using the existing ones for the journals, for example, in order not to rely on the current configuration of journals and accounts?

luisg123v: Yes. In fact, you should probably create a new company, to be sure about the currency. That's what localization modules do.

luisg123v (approved):
LGTM :+1:

PR #1708, opened 2025-09-24
[IMP] default_warehouse_from_sale_team: clarify AccessError for journal restrictions

ajteposte:
Improve UX by returning a specific AccessError when access is denied due to journal restrictions tied to sales teams (on account.move). The message now lists the affected journal(s) and suggests selecting one that matches the user’s branch/team. Security behavior is unchanged; this only refines the error copy.

Previously, users only saw a generic access denied error and assumed they lacked general billing permissions; this makes the cause explicit.

Added the necessary translations for the new error message and tests for the new flow.

Related to: typ!1371

### DIFF
--- a/default_warehouse_from_sale_team/models/ir_rule.py
+++ b/default_warehouse_from_sale_team/models/ir_rule.py
@@ -1,0 +1,25 @@
+from odoo import _, models
+from odoo.exceptions import AccessError
+
+
+class IrRule(models.Model):
+    _inherit = "ir.rule"
+
+    def _make_access_error(self, operation, records):
+        """Ensure that the selected journal is accessible for the user's team/branch.
+        If no team are defined for the journal, access is open to all users.
+        """
+        if records and records._name == "account.move":
+            journals = records.mapped("journal_id").sudo()
+            if journals:
+                journal_names = ", ".join(journals.mapped("display_name"))
+                records.invalidate_recordset()
+                return AccessError(
+                    _(
+                        "You do not have access to the selected journal(s) for this operation.\n"
+                        "Journal(s): %s \n\n"
+                        "Please select a journal that corresponds to your branch.",
+                        journal_names,
+                    )
+                )
+        return super()._make_access_error(operation, records)
--- a/default_warehouse_from_sale_team/tests/test_default_warehouse.py
+++ b/default_warehouse_from_sale_team/tests/test_default_warehouse.py
@@ -1,4 +1,7 @@
-from odoo import Command
+import re
+
+from odoo import Command, fields
+from odoo.exceptions import AccessError
 from odoo.tests import Form, TransactionCase, tagged
 
 
@@ -123,3 +126,57 @@ class TestSalesTeamDefaultWarehouse(TransactionCase):
         pick.action_assign()
         pick.move_ids.write({"quantity": 1, "picked": True})
         pick.button_validate()
+
+    def test_05_validate_saleteam_journal_access(self):
+        """Enforces record rule: move must use a public journal or one from the user's sales team;
+        otherwise AccessError with custom message.
+        """
+        today = fields.Date.context_today(self.env.user)
+        public_journal = self.env["account.journal"].create(
+            {
+                "name": "Public Sales",
+                "type": "sale",
+                "code": "PUB",
+            }
+        )
+        invoice = self.env["account.move"].create(
+            {
+                "move_type": "out_invoice",
+                "partner_id": self.env.ref("base.res_partner_2").id,
+                "company_id": self.company.id,
+                "date": today,
+                "invoice_date": today,
+                "invoice_date_due": today,
+                "is_payment_only": True,
+                "journal_id": public_journal.id,
+            }
+        )
+        # Case 1: Journal assigned to demo user's team → should succeed
+        my_team_journal = self.env["account.journal"].create(
+            {
+                "name": "My team Journal",
+                "type": "sale",
+                "code": "My-J",
+                "section_id": self.sales_team.id,
+            }
+        )
+        invoice.with_user(self.demo_user).write({"journal_id": my_team_journal.id})
+        self.assertEqual(invoice.journal_id, my_team_journal)
+
+        # Case 2: Public Journal assigned to another team → should raise UserError
+        other_team = self.env["crm.team"].create({"name": "Other Team"})
+        other_team_journal = self.env["account.journal"].create(
+            {
+                "name": "Other team Journal",
+                "type": "sale",
+                "code": "OTH-J",
+                "section_id": other_team.id,
+            }
+        )
+        error_msg = (
+            "You do not have access to the selected journal(s) for this operation.\n"
+            f"Journal(s): {other_team_journal.display_name} \n\n"
+            "Please select a journal that corresponds to your branch."
+        )
+        with self.assertRaisesRegex(AccessError, re.escape(error_msg)):
+            invoice.with_user(self.demo_user).write({"journal_id": other_team_journal.id})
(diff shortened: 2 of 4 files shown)


default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
+from odoo import _, models
+from odoo.exceptions import AccessError
+
+
+class IrRule(models.Model):
+    _inherit = "ir.rule"
+
+    def _make_access_error(self, operation, records):
+        """Ensure that the selected journal is accessible for the user's team/branch.
```
hugho-ad, line 9: parece que esto podria ser un patch para tu proyecto interno, no lo veo dentro del scope de este modulo

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+        If no team are defined for the journal, access is open to all users.
+        """
+        if records and records._name == "account.move":
+            journals = records.mapped("journal_id")[:3].sudo()
```
hugho-ad:
```suggestion
            journals = records.mapped("journal_id").sudo()
```
why the index?

default_warehouse_from_sale_team/models/ir_rule.py
ajteposte: Hola @hugho-ad , fue recomendación de @luisg123v ... originalmente lo trate acá: https://git.vauxoo.com/vauxoo/typ/-/merge_requests/1371 

Igual creo que no entorpece ninguna funcionalidad para otros proyectos. Al contrario es más orientativo con respecto al error, desde mi perspectiva (que no descarto este sesgada a mi proyecto jaja)

default_warehouse_from_sale_team/models/ir_rule.py
ajteposte: It’s just in case there are situations where multiple records are involved, so the message doesn’t get too long. But I’m unsure whether to leave it like this or not, I just need opinions c:, tks

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class IrRule(models.Model):
+    _inherit = "ir.rule"
+
+    def _make_access_error(self, operation, records):
+        """Ensure that the selected journal is accessible for the user's team/branch.
+        If no team are defined for the journal, access is open to all users.
+        """
+        if records and records._name == "account.move":
```
luisg123v, line 12: Use an early return.

default_warehouse_from_sale_team/models/ir_rule.py
luisg123v, line 9: Docstring doesn't match.

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+            journals = records.mapped("journal_id").sudo()
+            if journals:
+                journal_names = ", ".join(journals.mapped("display_name"))
+                records.invalidate_recordset()
```
luisg123v, line 16: You should probably call the super so it takes care of logging the mssage and invalidating recordsets.

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = "ir.rule"
+
+    def _make_access_error(self, operation, records):
+        """Ensure that the selected journal is accessible for the user's team/branch.
+        If no team are defined for the journal, access is open to all users.
+        """
+        if records and records._name == "account.move":
+            journals = records.mapped("journal_id").sudo()
+            if journals:
+                journal_names = ", ".join(journals.mapped("display_name"))
```
luisg123v, line 15: Format them as a list, as done natively.

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+                records.invalidate_recordset()
+                return AccessError(
+                    _(
+                        "You do not have access to the selected journal(s) for this operation.\n"
```
luisg123v, line 19: Message is not as clear, because it doesn't mention the originating record nor the operation

"Sorry, %s has not access to %s because the related journals are outside of their selected brnaches" (or similar).

The idea is extending the native message to provide more information, not less.

default_warehouse_from_sale_team/models/ir_rule.py
```diff
@@ -0,0 +1,25 @@
...
+class IrRule(models.Model):
+    _inherit = "ir.rule"
+
+    def _make_access_error(self, operation, records):
+        """Ensure that the selected journal is accessible for the user's team/branch.
+        If no team are defined for the journal, access is open to all users.
+        """
+        if records and records._name == "account.move":
+            journals = records.mapped("journal_id").sudo()
+            if journals:
```
luisg123v, line 14: This is re-writing the access error message in al,most all cases, as long as the journal entries have a journal, independently of what caused it. You should assure it's caused only by the associated rule by using `self._get_failing(records, mode)`

default_warehouse_from_sale_team/tests/test_default_warehouse.py
```diff
@@ -123,3 +126,57 @@ def test_04_warehouse_team_sale_policy(self):
         pick.action_assign()
         pick.move_ids.write({"quantity": 1, "picked": True})
         pick.button_validate()
+
+    def test_05_validate_saleteam_journal_access(self):
```
luisg123v, line 130: typo in salesteam

default_warehouse_from_sale_team/tests/test_default_warehouse.py
```diff
@@ -123,3 +126,57 @@ def test_04_warehouse_team_sale_policy(self):
...
+                "code": "PUB",
+            }
+        )
+        invoice = self.env["account.move"].create(
+            {
+                "move_type": "out_invoice",
+                "partner_id": self.env.ref("base.res_partner_2").id,
+                "company_id": self.company.id,
+                "date": today,
+                "invoice_date": today,
```
luisg123v, line 148: Most of these fields are not required to be provided, as they already match their default values. You probably need only `move_type` and `journal`. Since you don't plan to validate it, not even `partner_id` should be required

luisg123v (changes requested):
Regarding this:

`Related to: typ!1371`

Provide it only in PR description.

default_warehouse_from_sale_team/models/ir_rule.py
luisg123v: I see the native method display the first six records, you could use also that number.

default_warehouse_from_sale_team/models/ir_rule.py
luisg123v: Dependiendo de qué tan limpia o invasiva sea la implementación, pienso que podría ir acá. Si lo que hace es mejorar el mensaje nativo para proveer más información sin ser invasivo, me parece útil.

Hay que tener en cuenta que las reglas de registro de este módulo pueden ocasionar errores algo complejos de entender, desde que incluso administradores contables se pueden encontrar errores con registros con los que no están trabajando directamente. Por ejemplo, al abrir una factura a la  que sí tienen acceso, pero conciliada con varios pagos, uno de los cuáles está en un diario al que el usuario actual no tiene acceso por temas de equipo de ventas.

luisg123v:
Superseded by [vauxoo/typ!1371](https://git.vauxoo.com/vauxoo/typ/-/merge_requests/1371).

PR #1719, opened 2026-01-22
[REF] *: enable CI for all stable branches and OrchestSH T#97360

luisg123v:
Starting on 2025-12-08, GitHub Actions started running workflows using the default branch instead of the PR base branch [1]. Since the default branch was 17.0 and the workflows were designed to run only when targeting that branch, CI was not executed for other branches.

To fix this, workflows are reworked to:
- Run on any stable branch (i.e. `*.0`).
- Live on branch 19.0, which is the latest available version and is now configured as the default branch on GitHub.

In addition, CI files for 19.0 are updated to:
- Support OrchestSH by enabling it through variables and installing it during tests.
- Upgrade GitHub Actions versions:
  * `actions/checkout`
  * `actions/cache`
  * `actions/stale`

Lastly, the version is updated in a few missing places.

[1]: https://github.blog/changelog/2025-11-07-actions-pull_request_target-and-environment-branch-protections-changes/#what-is-changing


luisg123v:
FYI @isaako34 @desdelinux

PR #1721, opened 2026-01-28
[IMP] default_warehouse_from_sale_team: replace setUp with setUpClass

isaako34:
This is for efficiency purposes, so the code runs only once before all test methods in the class, instead of before every test.


isaako34:
@luisg123v Could you review please?

luisg123v (approved):
LGTM 👍

PR #1725, opened 2026-02-22
[MIG] mig-product_cost_usd: Migration to 19.0 task#97815

emtz10:
Changes include:
- Rename `product_uom` to `product_uom_id` in `sale.order.line`
  dependencies and computations according to [1].
- Update `_compute_price_rule` method call and dictionary extraction to
  match the new API signature according to [2].
- Optimize `product.pricelist.item` search by using direct `browse`
  instead of filtering `item_ids`.
- Create test records (Partner, Product, Pricelist) dynamically as demo
  data is no longer loaded by default.
- Set `price_discount` to `-15` in the test pricelist to comply with
  Odoo's native markup calculation standard.
- Replace usage of the deprecated `price` field on product templates with
  `get_contextual_price()` according to [3].
- Replace usage of the deprecated `pricing` div on product template
  views with `list_price_uom` according to [4].
- Deprecate the usage of `invisible` in views according to [5].
- Removed deprecated inheritance in `product.pricelist` as it was causing
  silent fallbacks to `list_price`, breaking margin computations in
  sales orders.
- Injected the raw USD cost directly via the official `_compute_base_price`
  hook in `product.pricelist.item`.
- Added native support for UoM (Unit of Measure) conversions.
- Grant `product.group_product_pricelist` explicitly in tests because
  demo data is no longer loaded by default, removing the implicit
  activation that previously made pricelist_id visible in the Form.

[1]: https://github.com/odoo/odoo/commit/c8461a5
[2]: https://github.com/odoo/odoo/commit/57ced81
[3]: https://github.com/odoo/odoo/commit/9e99a9d
[4]: https://github.com/odoo/odoo/commit/65652c7
[5]: https://github.com/odoo/odoo/pull/137031


product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,45 @@ class PricelistItem(models.Model):
...
+                price = product.uom_id._compute_price(price, uom)
+
+            # 3. Convert from USD to the target pricelist currency
+            # We set round=False to allow Odoo's native engine to apply its own rounding rules at the end
+            if usd_currency and target_currency and target_currency != usd_currency:
+                price = usd_currency._convert(price, target_currency, self.env.company, date, round=False)
+            return price
+
+        # If it's not our custom base, fallback to the standard Odoo computation
+        return super()._compute_base_price(product, quantity, uom, date, target_currency)
```
CarmenMiranda: I think it will be better to return early so we can omit an indentation for the ones here

```suggestion
        if self.base != "standard_price_usd":
            # If it's not our custom base, fallback to the standard Odoo computation
            return super()._compute_base_price(product, quantity, uom, date, target_currency)

        target_currency = target_currency or self.currency_id
        usd_currency = self.env.ref("base.USD", raise_if_not_found=False)

        # 1. Retrieve the raw USD cost from the product
        price = product.standard_price_usd

        # 2. Unit of Measure conversion
        if uom and product.uom_id != uom:
            price = product.uom_id._compute_price(price, uom)

        # 3. Convert from USD to the target pricelist currency
        # We set round=False to allow Odoo's native engine to apply its own rounding rules at the end
        if usd_currency and target_currency and target_currency != usd_currency:
            price = usd_currency._convert(price, target_currency, self.env.company, date, round=False)
        return price
```

CarmenMiranda (changes requested):
Hi @emtz10, This is missing the historic of the changes, to see the changes in migration more easily

Also, in the first commit, it is missing some parts like:

> Rename `product_uom` to `product_uom_id` in `sale.order.line`
>   dependencies and computations [1].

It should be like:

> Renam**ing** `product_uom` to `product_uom_id` in `sale.order.line`
>   dependencies and computations **according to** [1].

Same in similar cases

Also, mention where the changes are, like "Replace `Form` view simulation with direct ORM.." something like at the end " in Tests", or maybe this will not be needed with the historic, since there will be fewer changes in the commit.

product_cost_usd/models/product_pricelist_item.py
emtz10: Done

emtz10:
Ready for review again @CarmenMiranda

product_cost_usd/models/product_template.py
```diff
@@ -36,15 +36,15 @@ def check_cost_and_price(self):
         prec = self.env["decimal.precision"].precision_get("Product Price")
         for product in self:
             usd_seller = product.seller_ids.filtered(lambda x: x.currency_id == usd_currency)[:1]
-            list_price = usd_seller.price
+            list_price = usd_seller.price if usd_seller else 0.0
```
CarmenMiranda: This can be:

```suggestion
            list_price = usd_seller.price or 0.0
```

product_cost_usd/views/product_template_views.xml
```diff
@@ -5,8 +5,7 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="product.product_template_form_view" />
         <field name="arch" type="xml">
-            <xpath expr="//div[@name='pricing']" position="after">
-                <field name="currency_usd_id" invisible="1" />
```
CarmenMiranda: Missing in doc

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
...
+            # If it's not our custom base, fallback to the standard Odoo computation
+            return super()._compute_base_price(product, quantity, uom, date, target_currency)
+        target_currency = target_currency or self.currency_id
+        usd_currency = self.env.ref("base.USD", raise_if_not_found=False)
+
+        # 1. Retrieve the raw USD cost from the product
+        price = product.standard_price_usd
+
+        # 2. Unit of Measure conversion
+        if uom and product.uom_id != uom:
```
CarmenMiranda: I think `uom` is always set, or can be?

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
...
+            price = product.uom_id._compute_price(price, uom)
+
+        # 3. Convert from USD to the target pricelist currency
+        # We set round=False to allow Odoo's native engine to apply its own rounding rules at the end
+        if usd_currency and target_currency and target_currency != usd_currency:
```
CarmenMiranda: `target_currency` should always be set right?

Also, I think the USD shouldn't be deleted, but that can remain the same

```suggestion
        if usd_currency and target_currency != usd_currency:
```

CarmenMiranda (changes requested):

product_cost_usd/views/product_template_views.xml
emtz10: Done

product_cost_usd/models/product_pricelist_item.py
emtz10: Done

product_cost_usd/views/product_template_views.xml
CarmenMiranda: Still missing the removal of the invisible field

emtz10: Done

CarmenMiranda (approved):
Technically LGTM 👍🏼

product_cost_usd/__manifest__.py
```diff
@@ -19,6 +19,4 @@
         "data/res_currency_data.xml",
         "views/product_template_views.xml",
     ],
-    "installable": True,
```
luisg123v: Do this in the source version, as not strictly related to the migration.

product_cost_usd/models/product_template.py
```diff
@@ -36,15 +36,15 @@ def check_cost_and_price(self):
         prec = self.env["decimal.precision"].precision_get("Product Price")
         for product in self:
             usd_seller = product.seller_ids.filtered(lambda x: x.currency_id == usd_currency)[:1]
-            list_price = usd_seller.price
+            list_price = usd_seller.price or 0.0
```
luisg123v: Is this required? If the field is a numeric one, its falsy value should be zero anyway.

product_cost_usd/models/sale_order_line.py
```diff
@@ -4,27 +4,34 @@
...
-            price_rule = pricelist._compute_price_rule([(line.product_id, 1, False)])
-            _price, rule = price_rule[line.product_id.id]
-            suitable_rule = pricelist.item_ids.filtered(lambda x: x.id == rule)
-            if not suitable_rule or suitable_rule.base != "standard_price_usd":
+            pricelist = line.order_id.pricelist_id
+            date = line.order_id.date_order
+            product = line.product_id
+            price_rule = pricelist._compute_price_rule(
+                product,
+                1.0,
```
luisg123v: Use the `key=val` syntax in all params, for readability.

product_cost_usd/models/sale_order_line.py
```diff
@@ -4,27 +4,34 @@
...
+                currency=pricelist.currency_id,
+                uom=line.product_uom_id,
+                date=date,
+            )
+            _price, rule = price_rule.get(product.id, (0.0, False))
+            suitable_rule = self.env["product.pricelist.item"].browse(rule)
```
luisg123v: Variable name doesn't match, it's an ID (not a record). I'd suggest something like `suitable_rule_id`

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -10,25 +10,75 @@ def setUp(self):
         super().setUp()
```
luisg123v: Migrate to `setUpClass`

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -10,25 +10,75 @@ def setUp(self):
...
         self.mxn = self.env.ref("base.MXN")
         self.usd = self.env.ref("base.USD")
-        self.partner = self.env.ref("base.res_partner_4")
-        self.product_uom = self.env.ref("uom.product_uom_unit")
-        self.product = self.env.ref("product.product_product_24")
-        self.pricelist_15_usd = self.env.ref("product_cost_usd.pricelist_15_usd")
+        self.partner = self.env["res.partner"].create({"name": "Test Partner"})
+        # Get or create the unit UOM
+        self.product_uom = self.env["uom.uom"].search([("name", "=", "Unit")], limit=1)
+        if not self.product_uom:
```
luisg123v: This is not deterministic, tests should be deterministic.

Use the external ID `uom.product_uom_unit`

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -10,25 +10,75 @@ def setUp(self):
...
+            self.product_uom = self.env["uom.uom"].create(
+                {
+                    "name": "Unit",
+                }
+            )
+        # Create a product dynamically instead of relying on demo data
```
luisg123v: Remove comment, it adds no value. Literally all tests must do the same.

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -10,25 +10,75 @@ def setUp(self):
...
+        )
+
+        # Create pricelist with USD base cost strategy
+        self.pricelist_15_usd = self.env["product.pricelist"].create(
+            {
+                "name": "Pricelist 15% USD",
+                "currency_id": self.usd.id,
+            }
+        )
+        self.env["product.pricelist.item"].create(
```
luisg123v: Why do you create it separately? Create the item when creating the pricelist.

product_cost_usd/i18n/es.po
```diff
@@ -0,0 +1,109 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* product_cost_usd
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 15.0+e\n"
```
luisg123v: Missing regenerating translations file.

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
     base = fields.Selection(
         selection_add=[("standard_price_usd", "Cost in USD")], ondelete={"standard_price_usd": "set default"}
     )
+
+    def _compute_base_price(self, product, quantity, uom, date, currency):
+        """Compute the base price for a pricelist item based on the USD standard price.
```
luisg123v, line 12: This is actually part of the migration, I see no reason to place it in a separate commit.

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
     base = fields.Selection(
         selection_add=[("standard_price_usd", "Cost in USD")], ondelete={"standard_price_usd": "set default"}
     )
+
+    def _compute_base_price(self, product, quantity, uom, date, currency):
```
luisg123v: Method signature doesn't match, missing `kwargs`

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
...
+        2. Adjusts the price according to the requested Unit of Measure (UoM) to ensure
+           correct pricing for different quantities (e.g., dozens vs. units).
+        3. Converts the adjusted USD price into the pricelist's target currency.
+
+        By returning the exact converted base price at this stage, we allow Odoo's
+        native pricing engine to seamlessly handle subsequent operations like discounts,
+        surcharges, and financial rounding rules.
+
+        :param product: product.product/product.template record
+        :param float quantity: requested quantity
```
luisg123v: This is an inheritance, hence you shouldn't include this section (neither params nor return value), as those should be identical to the native method.

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,44 @@ class PricelistItem(models.Model):
...
+        :param uom: uom.uom record for the requested Unit of Measure
+        :param datetime date: date to use for currency conversion
+        :param currency: res.currency record for the pricelist's currency
+        :return: Computed base price (float)
+        """
+        if self.base != "standard_price_usd":
+            # If it's not our custom base, fallback to the standard Odoo computation
```
luisg123v: This is obvious, don't you think?

luisg123v (changes requested):
Regarding commit message:
- "Change `standard_price_usd` from `Float` to `Monetary` field": why?
- "Create test records (..) dynamically instead of relying on demo data ... to prevent unpredictable fallbacks.": the rationale is not as accurate. The actual rationale is, demo data is no longer loaded by default, so we must no longer depend on data demo in tests. BTW, mentioning `env.ref(...)` is too technical and adds no value.

product_cost_usd/models/product_pricelist_item.py
emtz10: Done

product_cost_usd/models/product_pricelist_item.py
emtz10: Done

product_cost_usd/models/product_pricelist_item.py
emtz10: Done

product_cost_usd/i18n/es.po
emtz10: Done

product_cost_usd/tests/test_standard_price_usd.py
emtz10: Done

emtz10:
Ready for review again @luisg123v

luisg123v:
@emtz10 there's a button that says "Re-request review". In the future, you can press that one so that the PR enters under my radar.

product_cost_usd/__manifest__.py
luisg123v: Not applied yet, create a PR to 15.0 with a commit similar to [this](https://git.vauxoo.com/vauxoo/project-template/-/merge_requests/129).

product_cost_usd/models/sale_order_line.py
luisg123v: Not applied correctly, the variable that was actually a record was renamed to add the `_id` suffix.

I meant the `rule` variable:

`_price, suitable_rule_id = price_rule.get(product.id, (0.0, False))`

product_cost_usd/tests/test_standard_price_usd.py
luisg123v: Not fully applied, you still check if it was found in:

`if not cls.product_uom:`

That condition is unreachable because, even if it were not found (impossible because it's in data), it will fail because the param `raise_if_not_found` is not provided.

To fix it, just remove the condition.

product_cost_usd/i18n/es.po
luisg123v: Not fully applied, only headers were updated. You need to generate the file again.

For instance, the `product.template` model name (`_description`) is  now "Product" instead of "Product Template".

product_cost_usd/models/product_pricelist_item.py
luisg123v: Missing updating calls to `super()`.

luisg123v:
Done, reviewed.

product_cost_usd/__manifest__.py
emtz10: Done https://github.com/Vauxoo/addons-vauxoo/pull/1730

luisg123v: Just missing bringing that commit here as part of the history.

desdelinux:
Hi @emtz10 Can you please provide evidence that functional testing was performed during the migration?

product_cost_usd/models/product_template.py
```diff
@@ -12,9 +12,9 @@ class ProductTemplate(models.Model):
         compute="_compute_currency_usd_id",
         help="Technical field to show the price fields as USD in the products",
     )
-    standard_price_usd = fields.Float(
```
CLaurelB: The change in the field type would prevent it from being set as company-dependent in other modules. This is because Monetary doesn't have support for the company dependent attribute.

cc. @luisg123v

product_cost_usd/models/sale_order_line.py
```diff
@@ -4,27 +4,34 @@
...
-            date = line.order_id.date_order
             if not line.product_id:
                 continue
-            price_rule = pricelist._compute_price_rule([(line.product_id, 1, False)])
-            _price, rule = price_rule[line.product_id.id]
-            suitable_rule = pricelist.item_ids.filtered(lambda x: x.id == rule)
-            if not suitable_rule or suitable_rule.base != "standard_price_usd":
+            pricelist = line.order_id.pricelist_id
+            date = line.order_id.date_order
+            product = line.product_id
```
CLaurelB: Better set it before the `if not line.product_id` in order to use the variable there too.

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -1,34 +1,85 @@
...
+                    {
+                        "product_id": product.id,
+                        "product_uom_qty": 1.0,
+                    }
+                )
+            ],
+        }
+        if pricelist:
+            order_vals["pricelist_id"] = pricelist.id
+        order = self.env["sale.order"].create(order_vals)
```
CLaurelB: In a real scenario, the order is created using the Form. The computes are also executed during create. Could you explain why it is not possible to use the Form, please?

product_cost_usd/models/product_pricelist_item.py
```diff
@@ -7,3 +7,36 @@ class PricelistItem(models.Model):
...
+        price = product.standard_price_usd
+
+        # 2. Unit of Measure conversion
+        if product.uom_id != uom:
+            price = product.uom_id._compute_price(price, uom)
+
+        # 3. Convert from USD to the target pricelist currency
+        # We set round=False to allow Odoo's native engine to apply its own rounding rules at the end
+        if usd_currency and target_currency != usd_currency:
+            price = usd_currency._convert(price, target_currency, self.env.company, date, round=False)
```
CLaurelB, line 41: Why is no longer required to use `date = fields.Date.context_today(self)` for this convert? In the previous version it was used event though there was also a date in the parameters.

emtz10: In previous versions, the `convert` method had an [assert](https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/res_currency.py#L268) for the date. In this new version, that has been changed, and [now](https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/models/res_currency.py#L280) it checks that a date is provided; if not, it assigns the current date.

emtz10:
Functional tests


https://github.com/user-attachments/assets/137688a7-48ad-423c-989c-9acc9a2d5c52

cc @desdelinux

product_cost_usd/tests/test_standard_price_usd.py
emtz10: Fixed

product_cost_usd/models/sale_order_line.py
emtz10: Done

product_cost_usd/__manifest__.py
emtz10: Done

desdelinux:
@luisg123v

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -0,0 +1,148 @@
...
+        order = Form(self.env["sale.order"])
+        order.partner_id = partner
+        with order.order_line.new() as line:
+            line.product_id = product
+            line.product_uom_qty = 1
+        order = order.save()
+        if pricelist:
+            order.pricelist_id = pricelist
+            # Recalculate line prices based on the new pricelist
+            order._recompute_prices()
```
luisg123v: - Assigning the pricelist should be inside the Form
- It shouldn't be required to recalculate manually

product_cost_usd/views/product_template_views.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">product.template.form.view.inherit</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='list_price_uom']" position="after">
+                <field name="currency_usd_id" invisible="1" />
```
luisg123v: Since these attributes now expect Python expressions, Python booleans are preferred:

```suggestion
                <field name="currency_usd_id" invisible="True" />
```

luisg123v (changes requested):

product_cost_usd/views/product_template_views.xml
emtz10: Done

product_cost_usd/tests/test_standard_price_usd.py
emtz10: Done

product_cost_usd/tests/test_standard_price_usd.py
```diff
@@ -1,35 +1,79 @@
...
-        self.pricelist_15_usd = self.env.ref("product_cost_usd.pricelist_15_usd")
-        self.pricelist_15_mxn = self.pricelist_15_usd.copy({"name": "Pricelist 15% MXN", "currency_id": self.mxn.id})
-        self.pricelist = self.env["product.pricelist"].create({"name": "Pricelist Demo"})
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.mxn = cls.env.ref("base.MXN")
+        cls.usd = cls.env.ref("base.USD")
+        cls.partner = cls.env["res.partner"].create({"name": "Test Partner"})
+        cls.env.user.write({"group_ids": [Command.link(cls.env.ref("product.group_product_pricelist").id)]})
```
CLaurelB, line 15: Please include why this is now needed in the commit description.

emtz10: Done

CLaurelB (approved):

luisg123v (approved):
LGTM 👍

PR #1730, opened 2026-05-17
[IMP] product_cost_usd: remove superfluous manifest key

emtz10:
Following the same approach as Odoo [1], the `auto_install` key is removed because its value is the same as the default.

[1]: https://github.com/odoo/odoo/pull/90209


luisg123v (approved):
LGTM 👍

PR #1731, opened 2026-05-22
[FIX] stock_by_warehouse: display correct quantity when filtering by warehouse T#98915

rolandojduartem:
## Problem
When calling the JSON stock calculation methods (`_compute_get_quantity_warehouses_json`), the system would crash with an `AttributeError` if a caller passed a raw integer ID or a sequence of IDs in the context key `warehouse_id` (e.g., trying to access `.id` on an integer or list).
However, passing recordsets in context is highly discouraged and leads to SQL query compilation warnings in Odoo 19 (e.g., in `_get_domain_locations`), making it necessary to support native database types in context instead of full recordsets.
## Root Cause
The JSON stock calculation methods expected `warehouse_id` in the context to be a full Odoo recordset to check the matching warehouse ID (`warehouse_id.id == warehouse.id`). This restricted callers from passing standard database integer IDs in the context, which are required by Odoo 19's native SQL query builders.
## Fix
Modified the `_compute_get_quantity_warehouses_json` method in:
- `product.product` (`stock_by_warehouse/models/product_product.py`)
- `product.template` (`stock_by_warehouse/models/product_template.py`)
Replaced the manual check of the context `warehouse_id` variable with a robust `.browse(warehouse_id).exists()` call. This allows the method to seamlessly resolve and support native integer IDs, list of IDs, or empty values in context (anything except recordsets), ensuring standard-compliant behavior.
## Testing
Procedure to verify the fix:
1. Open the Odoo Python shell (`odoo-bin shell`).
2. Run the JSON stock calculation for a storable product, passing a raw integer ID or a list of integer IDs in the context:
   ```python
   product = env['product.product'].search([], limit=1)
   # Test passing a raw integer ID
   res_int = product.with_context(warehouse_id=1)._compute_get_quantity_warehouses_json()
   
   # Test passing a list of integer IDs
   res_list = product.with_context(warehouse_id=[1])._compute_get_quantity_warehouses_json()
   ```
3. Verify that both computations resolve successfully without `AttributeError` exceptions, returning consistent JSON responses.
- ✅ Verified all integer and list formats resolve seamlessly and return matching JSON calculations.
- Related Task: T#98915

### DIFF
--- a/stock_by_warehouse/models/product_template.py
+++ b/stock_by_warehouse/models/product_template.py
@@ -45,6 +45,7 @@ class ProductTemplate(models.Model):
         # Just in case it's asked from other place different than product
         # itself, we enable this context management
         warehouse_id = self.env.context.get("warehouse_id")
+        context_warehouse = self.env["stock.warehouse"].browse(warehouse_id).exists()
 
         # Limit search to companies the user has access to
         for warehouse in self.env["stock.warehouse"].sudo().search([("company_id", "in", self.env.companies.ids)]):
@@ -54,7 +55,7 @@ class ProductTemplate(models.Model):
                 .with_context(warehouse_id=warehouse.id, location=False)
             )
             tmpl.invalidate_recordset()
-            if warehouse_id and warehouse_id.id == warehouse.id:
+            if context_warehouse and context_warehouse == warehouse:
                 info["warehouse"] = tmpl.qty_available_not_res
             info["content"].append(
                 {
--- a/stock_by_warehouse/models/product_product.py
+++ b/stock_by_warehouse/models/product_product.py
@@ -76,6 +76,7 @@ class ProductProduct(models.Model):
         # Just in case it's asked from other place different than product
         # itself, we enable this context management
         warehouse_id = self.env.context.get("warehouse_id")
+        context_warehouse = self.env["stock.warehouse"].browse(warehouse_id).exists()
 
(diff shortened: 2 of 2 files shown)


rolandojduartem:
@luisg123v could you review it?

I made this in the context of stock_by_warehouse_sale migration.

cc @desdelinux

stock_by_warehouse/models/product_template.py
```diff
@@ -45,6 +45,7 @@ def _compute_get_quantity_warehouses_json(self):
         # Just in case it's asked from other place different than product
         # itself, we enable this context management
         warehouse_id = self.env.context.get("warehouse_id")
+        warehouse_id = self.env["stock.warehouse"].browse(warehouse_id).exists()
```
luisg123v: - Variable name doesn't match, no longer an ID.
- Will the variable never be a record?
Same in the other case.

luisg123v (changes requested):
Title is too technical. What issue are actually you fixing? The title describes the technical solution, not the actual issue.

stock_by_warehouse/models/product_template.py
rolandojduartem: for your second question, not, it should not.

rolandojduartem:
@luisg123v done

rolandojduartem:
@luisg123v please review

luisg123v:
@rolandojduartem please, whenever a PR is ready to review again, use the button "re-request review" so it enters under my radar.

luisg123v (approved):
LGTM 👍

PR #1733, opened 2026-05-28
[FIX] stock_by_warehouse_sale: prevent singleton error in warehouse stock compute T#98915

rolandojduartem:
## Problem
When processing multiple sale order lines at once (e.g., during bulk creation, updates, or imports), a singleton error is raised, preventing the user from completing the action.
## Root Cause
In `_compute_warehouse_stock` of the `sale.order.line` model, the `warehouse_id` assignment incorrectly referenced the active recordset `self` (e.g., `self.order_id.warehouse_id`) inside the `for record in self:` loop instead of using the single loop variable `record`.
## Fix
- Corrected the reference to use the loop variable `record` (i.e., `record.warehouse_id = record.order_id.warehouse_id`).
- Cleaned up redundant and deprecated keys (`demo`, `test`, `installable`, `auto_install`) from the module manifest.
- Related Task: T#98915

### DIFF
--- a/stock_by_warehouse_sale/__manifest__.py
+++ b/stock_by_warehouse_sale/__manifest__.py
@@ -16,8 +16,4 @@
     "data": [
         "views/sale_order_views.xml",
     ],
-    "demo": [],
-    "test": [],
-    "installable": True,
-    "auto_install": False,
 }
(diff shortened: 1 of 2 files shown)


rolandojduartem:
@luisg123v done

luisg123v (changes requested):
This is doing two unrelated changes, please split into two commits.

rolandojduartem:
@luisg123v done

luisg123v (approved):
LGTM 👍

PR #1737, opened 2026-07-05
[IMP] product_cost_usd: allow different USD costs in multicompany t#97815

CLaurelB:
Align standard_price_usd with the native standard_price field, which
is already company-dependent, so each company can have its own USD
cost value for the same product.

Before this change, all companies shared the same USD cost. This was
limiting in multicompany scenarios where each company negotiates
different supplier prices in USD.

A pre-migration script is included to convert existing values.

### DIFF
--- a/product_cost_usd/migrations/19.0.1.0.0/pre-migration.py
+++ b/product_cost_usd/migrations/19.0.1.0.0/pre-migration.py
@@ -1,0 +1,21 @@
+import logging
+
+from odoo.upgrade import util
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _convert_standard_price_usd_to_company_dependent(cr)
+
+
+def _convert_standard_price_usd_to_company_dependent(cr):
+    """Convert standard_price_usd to company-dependent."""
+    util.make_field_company_dependent(
+        cr,
+        model="product.template",
+        field="standard_price_usd",
+        type="float",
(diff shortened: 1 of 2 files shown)


product_cost_usd/migrations/19.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _convert_standard_price_usd_to_company_dependent(cr)
+
+
+def _convert_standard_price_usd_to_company_dependent(cr):
+    """Convert the standard_price_usd field to be company dependent,
+    so we need to convert the existing data to fit the new structure.
```
luisg123v: Either collaps to one line or use title + blank line + description, to follow PEP recommendations

luisg123v (changes requested):
Commit message says:

> from the product.template table column to ir.property records.

But the model `ir.property` no longer exists, so that is inaccurate.

On the other hand, I think we should mention the rationale, which is aligning with the native standard price field, that is company dependent.

product_cost_usd/migrations/19.0.1.0.0/pre-migration.py
CLaurelB: Suggestions applied.

luisg123v (approved):
LGTM 👍

PR #1734, opened 2026-06-12
[IMP] sale_margin_percentage: remove superfluous manifest key T#98917

JesusValdez96:
### Fix license badge and rename security file

This commit fixes the license badge in README from AGPL-3 to LGPL-3 to match the manifest, and rename the security file from sale_margin_percentage_security.xml to res_groups_security.xml.

### Remove superfluous manifest key

Following the same approach as Odoo [1](https://github.com/odoo/odoo/pull/90209), the `installable` key is removed because its value is the same as the default.

PR #1739, opened 2026-09-02
[MIG] base_sync_login_email: Migrate to 19.0

hugho-ad:
## What

Brings `base_sync_login_email` to 19.0. The module is not present on this branch.

## What the module does

It keeps the email of a contact linked to an internal user in sync with that user's login:

- `res.partner` gains an `@api.constrains` on `email`/`active` that rejects an address not matching the login of any internal user attached to the partner.
- `res.users.write` propagates a login change to the partner's email, but only when login and email were already in sync and the new login is a valid address.

Net effect: a contact that backs an internal user cannot silently drift away from that user's login, whichever code path attempts the write.

## Changes

| File | Change |
|---|---|
| `__manifest__.py` | version `16.0.1.0.0` → `19.0.1.0.0` |
| `models/res_partner.py` | `self.env._` replaces the module-level `_` import |
| `i18n/es.po` | regenerated on 19.0: drops `#, python-format`, header now `Odoo Server 19.0`; fixes `modifica` → `modificar` |
| `README.rst` | bug-tracker template `version:%2016.0` → `%2019.0` |

No logic change was required: `single_email_re` is still exposed by `odoo.tools`, and the test helpers `TransactionCase`, `new_test_user` and `tagged` are unchanged.

## Validation

Installed on a fresh 19.0 database, tests run with `--test-tags=base_sync_login_email`:

```
0 failed, 0 error(s) of 4 tests
```

The four cases cover: the constraint rejecting a mismatched email, login→email propagation, a non-email login leaving the partner editable, and case-insensitive comparison.

`pre-commit-vauxoo` reports no finding for this module.

### DIFF
--- a/base_sync_login_email/i18n/es.po
+++ b/base_sync_login_email/i18n/es.po
@@ -4,11 +4,11 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 16.0+e\n"
+"Project-Id-Version: Odoo Server 19.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-04-09 04:26+0000\n"
-"PO-Revision-Date: 2025-04-09 04:26+0000\n"
-"Last-Translator: Luis González <lgonzalez@vauxoo.com>\n"
+"POT-Creation-Date: 2026-09-11 17:34+0000\n"
+"PO-Revision-Date: 2026-09-11 17:34+0000\n"
+"Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -21,15 +21,26 @@ msgid "Contact"
 msgstr "Contacto"
 
 #. module: base_sync_login_email
+#: model:ir.model.fields,field_description:base_sync_login_email.field_res_partner__display_name
+#: model:ir.model.fields,field_description:base_sync_login_email.field_res_users__display_name
+msgid "Display Name"
+msgstr "Nombre para mostrar"
+
+#. module: base_sync_login_email
+#: model:ir.model.fields,field_description:base_sync_login_email.field_res_partner__id
+#: model:ir.model.fields,field_description:base_sync_login_email.field_res_users__id
+msgid "ID"
+msgstr ""
+
+#. module: base_sync_login_email
 #. odoo-python
 #: code:addons/base_sync_login_email/models/res_partner.py:0
-#, python-format
 msgid ""
 "It's not possible to change this contact's email because it's associated to an internal user and the new email doesn't match the user login.\n"
 "- Contact's email: '%s'\n"
 "- User login: '%s'"
 msgstr ""
-"No es posible modifica el correo electrónico de este contacto porque está asociado con un usuario interno y el nuevo correo electrónico no coincide con el inicio de sesión (login) del usuario.\n"
+"No es posible modificar el correo electrónico de este contacto porque está asociado con un usuario interno y el nuevo correo electrónico no coincide con el inicio de sesión (login) del usuario.\n"
 "- Correo electrónico del contacto: '%s'\n"
 "- Inicio de sesión del usuario: '%s'"
 
--- a/base_sync_login_email/i18n/es.po
+++ b/base_sync_login_email/i18n/es.po
@@ -1,0 +1,39 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* base_sync_login_email
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0+e\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2025-04-09 04:26+0000\n"
+"PO-Revision-Date: 2025-04-09 04:26+0000\n"
+"Last-Translator: Luis González <lgonzalez@vauxoo.com>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: base_sync_login_email
+#: model:ir.model,name:base_sync_login_email.model_res_partner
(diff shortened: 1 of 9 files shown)


base_sync_login_email/i18n/es.po
rolandojduartem, line 1: in this commit, last paragraph does not add anything about documentation, please, could you remove it

base_sync_login_email/i18n/es.po
```diff
@@ -4,7 +4,7 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 16.0+e\n"
+"Project-Id-Version: Odoo Server 19.0+e\n"
```
rolandojduartem: Please, regenerate the file completely, the syntaxis changed from 18.0

base_sync_login_email/i18n/es.po
```diff
@@ -0,0 +1,39 @@
...
+#. module: base_sync_login_email
+#. odoo-python
+#: code:addons/base_sync_login_email/models/res_partner.py:0
+#, python-format
+msgid ""
+"It's not possible to change this contact's email because it's associated to an internal user and the new email doesn't match the user login.\n"
+"- Contact's email: '%s'\n"
+"- User login: '%s'"
+msgstr ""
+"No es posible modifica el correo electrónico de este contacto porque está asociado con un usuario interno y el nuevo correo electrónico no coincide con el inicio de sesión (login) del usuario.\n"
```
rolandojduartem: this should be fixed on source version, but it says "modifica" instead of "modificar"

base_sync_login_email/models/res_partner.py
```diff
@@ -0,0 +1,28 @@
...
+        """If the partner belongs to an internal user, validate the email matches the user login"""
+        for partner in self:
+            new_email = (partner.email or "").strip().lower()
+            internal_users = partner.with_context(active_test=False).user_ids.filtered(lambda u: not u.share)
+            internal_emails = {
+                login.lower() for login in internal_users.mapped("login") if single_email_re.match(login)
+            }
+            if internal_emails - {new_email}:
+                raise ValidationError(
+                    _(
```
rolandojduartem: the problem in this branch is the CI is failing, do you know why?

Anyway, use self.env._

base_sync_login_email/models/res_users.py
```diff
@@ -0,0 +1,20 @@
...
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    def write(self, vals):
+        """If login and email matches and login is changed, change email accordingly"""
+        new_login = vals.get("login")
+        if (
+            new_login
+            and "email" not in vals
+            and len(self) == 1
```
rolandojduartem, line 14: ok, but what happen if there more users?

rolandojduartem (changes requested):

rolandojduartem:
About MR description

- Changes section: It could be re-written,migrate-code thing does not add anything about documentation
- Note for reviewers: remove it

base_sync_login_email/i18n/es.po
hugho-ad: Regenerated by exporting from a 19.0 instance with the module installed (`base.language.export`), not by hand. The `#, python-format` markers are gone — for reference, only 1 of the 385 `es.po` files in 19.0 core still carries one — and the header is now `Odoo Server 19.0`. The result matches the other modules on this branch: `field_description` entries present, zero `python-format`.

base_sync_login_email/i18n/es.po
hugho-ad: Fixed here: `modifica` → `modificar`. Scoped to this branch only for now, so 16.0 still carries it.

base_sync_login_email/i18n/es.po
hugho-ad: Removed. The commit body now covers what the module does and the validation result, nothing else.

base_sync_login_email/models/res_partner.py
hugho-ad: Switched to `self.env._`.

On the CI: it is not this branch. All three jobs die in the first step, `actions/checkout@v6`, after 2-4 seconds, before the code is ever fetched:

> Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access.

`.github/workflows/test.yml` on 19.0 triggers on `pull_request_target` and then checks out the PR head (`ref: ${{ env.CI_COMMIT_SHA }}`). A recent `actions/checkout@v6` release refuses that combination by default — the classic pwn-request exposure — and since the workflow pins the floating `@v6` tag, the change arrived without anyone touching this repo. That matches the timeline: fork PRs #1734-#1737 were green in June/July, and everything from a fork fails now.

It affects every fork PR on this repo, not just this one. The clean fix is on the base branch: move the jobs that execute PR code to `pull_request` (no secrets) and keep `pull_request_target` for jobs that need credentials without running fork code. Setting `allow-unsafe-pr-checkout: true` would turn CI green but re-opens exactly the hole the action is flagging, so I would rather not propose that. Happy to open a separate PR for the workflow if you want.

base_sync_login_email/models/res_users.py
hugho-ad: With more than one record the block is skipped and nothing is synced, which is the correct outcome here: `login` is unique at the database level (`_login_key = models.Constraint("UNIQUE (login)", ...)` in `odoo/addons/base/models/res_users.py`). A `write({'login': ...})` over a recordset of N>1 would assign the same login to several users and `super().write()` would raise on that constraint anyway — so there is no valid multi-record case being silently dropped.

The guard also prevents a crash before that point: without `len(self) == 1`, the `self.login` read on the next line would raise a singleton error on a multi-record set.

So the behaviour with several users is: the module stays out of the way and the standard uniqueness error surfaces, which is what you want. Left as is, but happy to add a comment in the code stating that rationale if you prefer it explicit.

rolandojduartem (approved):
LGTM 👍

hugho-ad:
@moylop260 

me ayudas pls con review y merge?

moylop260:
👍

moylop260 (approved):
👍

PR #1741, opened 2026-09-11
[FIX] base_sync_login_email: skip the login check while loading data files

hugho-ad:
## What
`_check_email_internal_user` returns early when the context carries `install_mode`, the flag Odoo sets while importing XML/CSV records (data and demo phases alike) and the one the core itself uses to relax validations in that phase (`res.currency`, `res.company`, `auth_signup`). Same approach as OCA/server-auth#986 for `password_security`. Behaviour outside data loading is unchanged. A new test covers both sides.

## Why
Records loaded from a module's files are not user input, so the safeguard has nothing to protect there and enforcing it only blocks module installation. Odoo's own demo data is the proven case: `l10n_mx_hr_payroll` (demo line 138) and `l10n_us_hr_payroll` (demo line 71) deliberately create internal users whose login differs from the partner email, and with this module installed both demo sets fail with the constraint message.

Worse, `load_demo()` rolls the failed demo back inside a `savepoint(flush=False)` and there is no flush between the module's data and demo phases, so pending writes of the regular data files are lost with the rollback while the ORM believes they are persisted. `l10n_mx_edi_payslip` then aborts the registry with `No rule parameter with code "l10n_mx_daily_min_wage"`. This is what broke the CI of https://git.vauxoo.com/vauxoo/instance/-/merge_requests/1953.

## How to Test
Install with demo, before and after the patch:

```
odoo-bin -d test --with-demo --test-enable --test-tags /base_sync_login_email --stop-after-init \
  -i base_sync_login_email,l10n_mx_hr_payroll,l10n_us_hr_payroll
```

Before: two `demo data failed to install` warnings, 4 tests. After: no warnings, 5 tests, both payroll demos loaded.

Task: t#33602

## Checklist
- [ ] CI passes (GitHub checks refuse fork PRs under `pull_request_target`, see #1740; validated through vauxoo/instance!1957 instead)
- [x] Tests cover the change
- [ ] Docs updated if behaviour changed (not needed)

### DIFF
--- a/base_sync_login_email/models/res_partner.py
+++ b/base_sync_login_email/models/res_partner.py
@@ -9,6 +9,15 @@ class ResPartner(models.Model):
     @api.constrains("email", "active")
     def _check_email_internal_user(self):
         """If the partner belongs to an internal user, validate the email matches the user login"""
+        # Records loaded from a module's data or demo files are not user input, so the safeguard has nothing to
+        # protect there and enforcing it only blocks module installation. Odoo's own demo data is the proven case:
+        # l10n_mx_hr_payroll and l10n_us_hr_payroll give internal users a login that differs from the partner
+        # email on purpose, the failed demo load is rolled back inside a savepoint without a previous flush, and
+        # the pending writes of the module's regular data are lost with it, breaking every module installed later.
+        # install_mode is the flag Odoo sets while importing XML/CSV records and the one the core itself uses to
+        # relax validations in that phase (res.currency, res.company, auth_signup).
+        if self.env.context.get("install_mode"):
+            return
         for partner in self:
             new_email = (partner.email or "").strip().lower()
             internal_users = partner.with_context(active_test=False).user_ids.filtered(lambda u: not u.share)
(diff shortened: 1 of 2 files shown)


luisg123v (changes requested):
I think not only demo data, this shouldn't block any module installation. IMO this is similar to https://github.com/OCA/server-auth/pull/986

hugho-ad:
Switched to `install_mode` in 9077a55d4: it is set by `convert.py` for every XML/CSV record in both the data and demo phases, and it is the flag the core uses for the same purpose in `res.currency`, `res.company` and `auth_signup`. Test adjusted accordingly. Verified locally installing `base_sync_login_email,l10n_mx_hr_payroll,l10n_us_hr_payroll` with demo: both demo sets load and the 5 module tests pass.

rolandojduartem:
that is the reason that I do not like corrupted CI :smile: but LGTM

rolandojduartem:
just a lot of comment just for that, just a opinion

